PR #34, opened 2020-05-05
 decouple schema processing from model

jaypipes:
Moves the processing logic for iterating over the OpenAPI3 schema
elements into a pkg/schema package, creating a Helper struct that wraps
nicely all the funky logic for constructing the model information for
the API itself.
    
"Resource" has been renamed to "CRD" to better represent what it is.

Also fixes type definition output to remove Exception classes.


mhausenblas:
This makes sense to me and overall LGTM, thanks @jaypipes! One thing, going forward, I believe we should start getting serious about is unit tests. I don't wanna block you with this, right now, but I'd suggest we should aim for it in one of the upcoming iterations.

PR #40, opened 2020-05-27
start mapping out reconciler runtime

jaypipes:
Introduce some reconciler implementation in the ACK runtime by wrapping
the code that instantiates Kubernetes object "prototypes" (basically
pointers to empty structs that implement the upstream runtime.Object
interface) and code that uses a k8s client to fetch resource
information from the k8s API server.

Added in the requeue/error handling from ALB IC v1beta2 code (see
pkg/runtime/error.go and pkg/runtime/reconciler.go) to wrap the handling
of errors received from the AWS resource managers to gel with the
`controller-runtime.Result` returned struct.

In order to do the above cleanly, I created a new AWSResourceFactory
interface, separate from the AWSResource interface. The
AWSResourceFactory interface is responsible for returning empty
Kubernetes object prototypes to callers and for producing AWSResource
objects, which are in turn fairly simple wrappers over the Kubernetes
runtime.Object.


pkg/runtime/error.go
```diff
@@ -0,0 +1,89 @@
...
+	return e.err.Error()
+}
+
+func (e *RequeueAfterError) Duration() time.Duration {
+	return e.duration
+}
+
+func (e *RequeueAfterError) Unwrap() error {
+	return e.err
+}
```
jaypipes, line 89: Note to reviewers... this comes from here: https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/v1beta2/pkg/runtime/errors.go

mhausenblas:
LGTM, thanks!

PR #45, opened 2020-06-09
flesh out example Bookstore API integration in resource manager

jaypipes:
This patch adds a simplistic implementation of the
AWSResourceManager.ReadOne() interface method for the Bookstore API
example book resource manager. We follow a similar strategy to the
AppMesh v1beta2 implementation, farming out most real work to a
findSDKBook() method that communicates with the AWS service API via the
aws-sdk-go.

The AWSResourceManager interface's methods are updated to pass a
context.Context struct. This allows us to standardize on the
XXXWithContext() functions of the aws-sdk-go and pass on a backgrounded
context into those SDK functions.
    
Besides the above findSDKBook() code, the remainder of this patch is
simply a move of the services/example/pkg/sdk package into
services/example/sdk. This directory/package does NOT appear in the real
service controllers since the import will be the
aws-sdk-go/services/$SERVICE package.


jaypipes:
Reviewers: please see https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/1e1984b500cee9b65a4429e518db7b25bd39e27a/pkg/mesh/resource_manager.go#L83-L96 for the inspiration for this patch's contents.

PR #47, opened 2020-06-09
AWSResourceDescriptor enhancements

jaypipes:
We want a struct that will encapsulate the service-specific Kubernetes and
AWS resource metadata and comparison logic. I'd previously placed this
in AWSResourceManagerFactory and AWSResourceFactory. This patch
consolidates this functionality into an AWSResourceDescriptor type.

* Renames AWSResourceFactory -> AWSResourceDescriptor
* Moves AWSResourceFactoryManager.GroupKind -> AWSResourceDescriptor.GroupKind
* Returns a metav1.GroupKind from AWSResourceDescriptor.GroupKind instead of string
* Adds Equal and Diff methods to AWSResourceDescriptor


pkg/runtime/reconciler.go
```diff
@@ -110,19 +112,29 @@ func (r *reconciler) sync(
...
 	} else {
-		// TODO(jaypipes): implement checks here for whether the desired is
-		// already equal to the observed
+		// Check to see if the latest observed state already matches the
+		// desired state and if so, simply return since there's nothing to do
+		if r.rd.Equal(desired, latest) {
+			return nil
+		} else {
+			diff := r.rd.Diff(desired, latest)
+			r.log.V(1).Info("meshSpec changed",
```
jaypipes, line 125: LOL, I need to s/meshSpec/resource/ here.

PR #56, opened 2020-06-16
Dockerfile and integtest inital setup

varun1524:
Issue #, if available:

Description of changes:
- Dockerfile for Bookstore service
- initial test setup


services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Build the manager binary
+FROM golang:1.14.1 as builder
+
+ARG workDirName=/github.com/aws/aws-service-operator-k8s
```
nithu0115, line 4: It is a good practice to use `lower_case` variable names. Please change this var to `work_dir`

services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# TODO: for now copy whole repository. Later once autogeneration is ready, need to determine which dependencies to take in
+COPY . $workDirName/
+# Build
+RUN  go build -a -o bin/manager $workDirName/services/bookstore/cmd/controller/main.go
+
+
+FROM amazonlinux:latest
+ARG workDirName=/github.com/aws/aws-service-operator-k8s
+WORKDIR /
+COPY --from=builder $workDirName/bin/manager bin/.
```
nithu0115, line 28: I like my application to be under `/app` instead of `/bin`. However, I don't have any strong option. Between, please use absolute path here.

services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
+# Build the manager binary
+FROM golang:1.14.1 as builder
+
+ARG workDirName=/github.com/aws/aws-service-operator-k8s
+WORKDIR $workDirName
+# For building Go Module required
+ENV GOPROXY=direct
```
vijtrip2, line 7: "Each ENV line creates a new intermediate layer" 

You can merge all ENVs in a single line to optimize the build.

services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# and so that source changes don't invalidate our downloaded layer
+RUN  go mod download
+# Copy the go source
+# TODO: for now copy whole repository. Later once autogeneration is ready, need to determine which dependencies to take in
+COPY . $workDirName/
+# Build
+RUN  go build -a -o bin/manager $workDirName/services/bookstore/cmd/controller/main.go
+
+
+FROM amazonlinux:latest
```
vijtrip2, line 25: Nit: Can we use Alpine images? Docker recommends using Alpine images because of their smaller size.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#from

services/bookstore/Dockerfile
varun1524: sure updating it.

services/bookstore/Dockerfile
varun1524: My reason behind it was generated output from build is executable hence kept it under bin.

services/bookstore/Dockerfile
varun1524: Yes, it was a design in older docker. But in latest best-practices docker it ADD, RUN and COPY creates intermediate layer. That and this way i think it is more readable readable too. Let me know if you have strong opinion about it. thanks

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

vijtrip2: Nice!

services/bookstore/Dockerfile
varun1524: kept /bin but used absolute path

test/integration/services/Common.go
```diff
@@ -0,0 +1,42 @@
+package services
```
A-Hilaly, line 1: The naming style used for filenames in this project (and most Go projects) is all lowercase with underscore separation when necessary, i think it's better to name this file `common.go`

test/integration/services/Common.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"flag"
+	"fmt"
+	"github.com/onsi/ginkgo/config"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"os"
+)
```
A-Hilaly, line 10: It's good practice to separate import paths in three different groups: Standard library imports / internal packages imports / external packages imports
https://github.com/golang/go/wiki/CodeReviewComments#imports

test/integration/services/bookstore/integration_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"flag"
+	"fmt"
+	"github.com/aws/aws-service-operator-k8s/test/integration/services"
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
```
A-Hilaly, line 12: Same comment for this part

test/integration/services/Common.go
jaypipes: Yep, @A-Hilaly is correct (and I've been following this pattern in the rest of the codebase). Depending on your editor or IDE of choice, you can typically find a module or plugin that autoformats this kind of thing for you. For instance, I use vim, and the vim-go support automatically runs `go fmt` whenever I save a Go file. Makes things like this a non-issue ;)

services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN  go mod download
+# Copy the go source
+# TODO: for now copy whole repository. Later once autogeneration is ready, need to determine which dependencies to take in
+COPY . $work_dir/
+# Build
+RUN  go build -a -o $work_dir/bin/manager $work_dir/services/bookstore/cmd/controller/main.go
```
jaypipes, line 22: Let's call the binary "controller" instead of "manager" please.

services/bookstore/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+
+FROM amazonlinux:2
```
jaypipes, line 25: This is cool for now. Eventually we'll want to use a smaller base image (especially since we're going to have dozens of service controller pods, so the smaller the better!) :)

We should eventually be able to use a very slim image (close to `scratch` with certs added into a well-known dir)

allamand: That’s why distroless exists :)

test/integration/services/Common.go
varun1524: oh ok. i just allowed GoLand to take care of it. but sure will fix it .ty

services/bookstore/Dockerfile
jaypipes: @allamand we'll definitely look into scratch or distroless in the future. we're going to stick with AL2 for the time being because its already cleared from an open source distribution perspective (internally at Amazon). Agree we should use a slimmer base image and we'll tackle that in a followup GH issue.

test/integration/services/bookstore/integration_test.go
```diff
@@ -0,0 +1,34 @@
...
+	RunSpecs(t, "Amazon Service Operator K8s BookStore")
+}
+
+var _ = BeforeSuite(func() {
+	fmt.Printf("Using KUBECONFIG=\"%s\"\n", framework.TestContext.KubeConfig)
+})
+
+var _ = Describe("[bookstore-integration]", func() {
+	It("should able to run", func() {
+		fmt.Println("Tests running successfully")
```
nithu0115, line 32: @varun1524, is it missing test case here ? may be add a //TODO comment here

test/integration/services/bookstore/integration_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+func init() {
+	services.RegisterFlags(flag.CommandLine)
+}
+
+func TestIntegration(t *testing.T) {
+	flag.Parse()
+	framework.AfterReadingAllFlags(&framework.TestContext)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Amazon Service Operator K8s BookStore")
```
nithu0115, line 23: // TODO comment to update the name here, probably ?

test/integration/services/bookstore/integration_test.go
varun1524: that is the next thing anyway we are working on so not needed and as per offline discussion resolving it.

PR #59, opened 2020-06-17
Integration Test Setup

varun1524:
I have updated integration_test setup bash script to setup controller and run integtest if BASE_GIT_TAG ENV variable is set. Otherwise it would run it on Latest commit only. Integrated helm chart usage to deploy and run service controller. Tested on CircleCI.
https://app.circleci.com/pipelines/github/varun1524/aws-service-operator-k8s/93/workflows/8a62282b-bed9-4c58-81b1-16516a7fa11e/jobs/184

Description of changes:


scripts/Makefile
```diff
@@ -0,0 +1,36 @@
...
+	CONTROLLER_GEN_TMP_DIR=$$(mktemp -d) ;\g
+	cd $$CONTROLLER_GEN_TMP_DIR ;\
+	go mod init tmp ;\
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0 ;\
+	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
+	}
+CONTROLLER_GEN=$(GOBIN)/controller-gen
+else
+CONTROLLER_GEN=$(shell which controller-gen)
+endif
```
jaypipes, line 36: so this is the kubebuilder-generated Makefile, yeah?

I think I'd prefer to do a couple things with this:

1) Convert it into a proper shell script instead of the Makefile monstrosity that it currently is. For example, you can make an `ensure_controller_gen` Bash function like this:

https://github.com/aws/aws-service-operator-k8s/blob/mvp/scripts/lib/aws.sh#L17

to replace the `controller-gen` Makefile target above.

2) Make the script accept a CLI argument of the AWS service to generate CRDs for (that way we have a single script that we can run for any AWS service controller)

scripts/run-integration-tests.sh
```diff
@@ -16,6 +16,7 @@ source "$DIR"/lib/cluster.sh
 # Variables used in /lib/aws.sh
 OS=$(go env GOOS)
 ARCH=$(go env GOARCH)
+GO111MODULE=on
 
 : "${AWS_DEFAULT_REGION:=us-west-2}"
 : "${K8S_VERSION:=1.14.6}"
```
jaypipes, line 22: This is fine. Might be good to test on a later K8s version, though. EKS currently supports 1.16.8, for instance.

scripts/run-integration-tests.sh
```diff
@@ -147,13 +148,38 @@ mkdir -p "$TEST_CONFIG_DIR"
...
+#            docker tag "ack"/"$d":baseversion "$AWS_ACCOUNT_ID".dkr.ecr."$AWS_DEFAULT_REGION".amazonaws.com/"$d":baseversion
+#            docker push "$AWS_ACCOUNT_ID".dkr.ecr."$AWS_DEFAULT_REGION".amazonaws.com/"$d"
+#          fi
+#        done
+#    fi
+#done
+
+# TODO: 2. We will run TODO:1 code block for latest version (includes pull request)
+#  generate all CRDs for it to /tmp/crd/test
+#  make install CRD_DIR="directory path"
```
jaypipes, line 180: Given the above requested changes, this would be something like:

```
./scripts/generate-crds.sh "$SERVICE_NAME"
```

scripts/Makefile
varun1524: >> so this is the kubebuilder-generated Makefile, yeah?
Yes I liked the way it was defined there so took part of it relevant and modified.

Sure makes sense to have bash script instead. Will apply those changes in next commit. thanks.

scripts/lib/generate-crds.sh
```diff
@@ -0,0 +1,17 @@
...
+    then
+      set -e
+      CONTROLLER_GEN_TMP_DIR="$(mktemp -d)"
+      cd "$CONTROLLER_GEN_TMP_DIR"
+      go mod init tmp
+      go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
+      rm -rf "$CONTROLLER_GEN_TMP_DIR"
+      CONTROLLER_GEN=$(GOBIN)/controller-gen
+    fi
+    $CONTROLLER_GEN "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config="$__crd_dir"
```
jaypipes, line 16: Remove the above line that actually calls the `controller-gen` binary.

The code that *calls* `controller-gen` should be in a `scripts/generate-crds.sh` file.

Also, the `paths="./..."` here should be a path to the `services/$SERVICE_ALIAS/apis/$API_VERSION/...` directory.

Additionally, remove "webook" and "rbac:roleName=manager-role" from the command. At this point, we are only generating the resource configuration manifests (CRDs)

scripts/lib/generate-crds.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+ensure_crd_gen() {
```
jaypipes, line 3: Go ahead and throw this function into a file called `scripts/lib/k8s.sh` where we will put all Kubernetes helper functions. Also, rename the function `ensure_controller_gen` since we want to have the function ONLY responsible for ensuring that the `controller-gen` binary is installed.

scripts/run-integration-tests.sh
```diff
@@ -12,13 +12,15 @@ DIR=$(cd "$(dirname "$0")"; pwd)
 source "$DIR"/lib/common.sh
 source "$DIR"/lib/aws.sh
 source "$DIR"/lib/cluster.sh
+source "$DIR"/lib/generate-crds.sh
```
jaypipes, line 15: make this `source "$DIR"/lib/k8s.sh`

scripts/run-integration-tests.sh
```diff
@@ -147,13 +149,38 @@ mkdir -p "$TEST_CONFIG_DIR"
...
 fi
 
+
+# TODO: 1. We will run this block of code for Base Version, it will CRDs and push Controller Images to ECR
+# Generate All the CRDs for all the service Base Commit/Tag in /tmp/crd/base
+#ensure_crd_gen "/tmp/crd/base"
+# In general fetch services, build Docker Image for each and push it to ECR if Image:BaseVersion already does not exist.#
+#for d in ./services/*; do
+#    if [ -d "$d" ]; then
+#      echo "$d"
```
jaypipes, line 165: And here, we will be calling:

```bash
./scripts/generate-crds.sh $d
```

scripts/run-integration-tests.sh
```diff
@@ -147,13 +149,38 @@ mkdir -p "$TEST_CONFIG_DIR"
...
+#            docker tag "ack"/"$d":baseversion "$AWS_ACCOUNT_ID".dkr.ecr."$AWS_DEFAULT_REGION".amazonaws.com/"$d":baseversion
+#            docker push "$AWS_ACCOUNT_ID".dkr.ecr."$AWS_DEFAULT_REGION".amazonaws.com/"$d"
+#          fi
+#        done
+#    fi
+#done
+
+# TODO: 2. We will run TODO:1 code block for latest version (includes pull request)
+#  generate all CRDs for it to /tmp/crd/test
+#ensure_crd_gen "/tmp/crd/test"
```
jaypipes, line 181: This needs to be `ensure_controller_gen` and called before we execute `./scripts/generate-crds.sh`

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+ensure_controller_gen() {
+  go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0
```
jaypipes, line 4: this is fine, since it's idempotent, but probably would be good to do:

```
if !is_installed "controller-gen"; then

fi
```
block around the `go get` in order to prevent unnecessarily calling it when controller-gen is already installed.

Note the `is_installed` Bash function is here:

https://github.com/aws/aws-service-operator-k8s/blob/mvp/scripts/lib/common.sh#L11

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+ensure_service_controller_running() {
+  echo "Ensure that service controllers are running for given version"
+  # Give executable permission
+  chmod +x ./scripts/generate-crds.sh
```
jaypipes, line 10: This definitely should not be done in here :) Just change the `scripts/generate-crds.sh` script's execute permission and check in that file to git.

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,43 @@
...
+  __image_version=$1
+  __crd_path=$2
+
+  for d in ./services/*; do
+    if [ -d "$d" ]; then
+      echo "service: $d"
+      ./scripts/generate-crds.sh "$d" "$__crd_path"
+        for f in "$d"/*; do
+          if [ "$f" = "$d"/"Dockerfile" ]; then
+            echo "$f"
```
jaypipes, line 21: probably best to remove the above echo :)

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,43 @@
...
+  echo "Ensure that service controllers are running for given version"
+  # Give executable permission
+  chmod +x ./scripts/generate-crds.sh
+
+  __image_version=$1
+  __crd_path=$2
+
+  for d in ./services/*; do
+    if [ -d "$d" ]; then
+      echo "service: $d"
```
jaypipes, line 17: best to remove the above echo. I presume it was just there for debugging purposes while you wrote this script?

scripts/lib/k8s.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+ensure_service_controller_running() {
+  echo "Ensure that service controllers are running for given version"
+  # Give executable permission
+  chmod +x ./scripts/generate-crds.sh
+
+  __image_version=$1
+  __crd_path=$2
+
```
jaypipes, line 14: I was expecting this function to have a `service` parameter, so that it only ensured a controller for a specific service (passed as a parameter) was running... perhaps you could move the `for` loop across the `services/` directory below outside this function and make this function smaller-scoped?

scripts/run-integration-tests.sh
```diff
@@ -173,98 +132,80 @@ export KUBECONFIG=$KUBECONFIG_PATH
...
+
+  __ack_source_tmpdir="/tmp/ack-src-$BASE_GIT_TAG"
+  echo "Checking out ack source code for $BASE_GIT_TAG ..."
+  git clone git@github.com:varun1524/aws-service-operator-k8s.git "$__ack_source_tmpdir" || exit 1
+
+  pushd "$__ack_source_tmpdir" || exit
+  git checkout -b "$BASE_GIT_TAG" "$BASE_GIT_TAG"
+
+  __test_crd_path="/tmp/crd/base"
+  ensure_service_controller_running "$BASE_GIT_TAG" "$__test_crd_path"
```
jaypipes, line 153: here, you should pass the name of the service as a parameter to `ensure_service_controller_running`

scripts/run-integration-tests.sh
```diff
@@ -173,98 +132,80 @@ export KUBECONFIG=$KUBECONFIG_PATH
...
-install_helm_chart base
-ensure_controller_pods
+BASE_INTEGRATION_DURATION=0
 if [[ "$TEST_PASS" -ne 0 ]]; then
   echo "NOTE: Skipping base test run because test is marked as failed"
+elif [[ -z "${BASE_GIT_TAG+x}" ]]; then
+  echo "NOTE: Skipping base test run because BASE_GIT_TAG is not given"
 else
-  echo "Running base integration test"
-  # TODO: RUN BASE TEST HERE
```
jaypipes, line 182: You should move the `for` loop over the contents of the `services/` directory here and for each of the `echo` calls below, tack on the name of the service that you are running integration tests for. Make sense?

scripts/lib/helm.sh
```diff
@@ -92,4 +59,21 @@ ensure_controller_pods() {
...
   fi
+}
+
+ensure_helm_chart_installed() {
+  if ! should_execute ensure_helm_chart_installed; then
+     return 1
+  fi
+
+  local __image_tag_suffix="$1"
+  echo "Installing helm chart '$HELM_LOCAL_CHART_NAME' with image $AWS_ECR_REPO_NAME:$__image_tag_suffix"
```
jaypipes, line 70: probably want to pass the Helm chart name (and the image name) as parameters to the function as opposed to using global Bash variables.

scripts/lib/aws.sh
```diff
@@ -26,3 +26,22 @@ ensure_aws_k8s_tester() {
         chmod +x $TESTER_PATH
     fi
 }
+
+ensure_ecr_image() {
+  local __image_tag="$1"
```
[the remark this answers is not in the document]
varun1524: i know variable name might be confusing. actually currently __image_tag is "service_name:short_commit_sha". __image_tag because we are actually tagging image 
$AWS_ACCOUNT_ID.dkr.ecr.$AWS_DEFAULT_REGION.amazonaws.com/$REPO_NAME by tag "service_name:short_commit_sha"
https://github.com/varun1524/aws-service-operator-k8s/blob/integtest/scripts/lib/k8s.sh#L21

I will update tag to ack-$service_name-$commit

IMAGE_NAME is this: 
https://github.com/varun1524/aws-service-operator-k8s/blob/integtest/scripts/run-integration-tests.sh#L91
That will be universal so I am pulling that from there. 

and version will be "ack-$service-$commit"

Please let me know if that looks good.

Image to show how it will look to avoid confusion.
Screen Shot 2020-06-25 at 11 36 53 AM

scripts/lib/helm.sh
varun1524: yes, earlier we were deploying all controllers together by substituing just version (commit) value.
I am working on changing that to deploying one controller by providing servicealias. Fixing this in next commit.

scripts/lib/aws.sh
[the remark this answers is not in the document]
jaypipes: ah, gotcha. yeah, maybe just rename the variable to make it clear it includes the service name :)

PR #60, opened 2020-06-19
Using helm to initialize service controllers for integration-test.

vijtrip2:
Issue #, if available:

Description of changes:
* Using the temporary docker hub images and helm charts to start service controllers inside the EKS cluster as proof of concept.
* Using this temp repo (https://hub.docker.com/repository/docker/errorbyte/ack) to build helm charts ( https://vijtrip2.github.io/aws-service-operator-k8s/helm/charts/ ) 
--------------
#### Not covered in this PR
* I will update the helm chart to use ECR repository from integration test
* @varun1524 will add the logic of building images for the latest commit and uploading to above mentioned ECR repository

### DIFF
--- a/scripts/lib/helm.sh
+++ b/scripts/lib/helm.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+
+add_helm_repo() {
+   if ! should_execute add_helm_repo; then
+     return 1
+   fi
+
+   # add the helm repository containing charts for starting AWS service controllers
+   if ! helm repo add "$HELM_LOCAL_REPO_NAME" "$HELM_REPO_SOURCE" > /dev/null 2>&1; then
+     echo "Unable to add local helm repo from '$HELM_REPO_SOURCE'"
+     TEST_PASS=1
+     return 1
+   fi
+
+  #list the charts in the the local repo
+  echo "Validating the presence of '$HELM_REPO_CHART_NAME' in local repo '$HELM_LOCAL_REPO_NAME' "
+  helm_repo_output_lines=$(helm search repo "$HELM_LOCAL_REPO_NAME" | grep "$HELM_REPO_CHART_NAME" | wc -l)
+  if [[ "$helm_repo_output_lines" -gt 0 ]]; then
+    echo "'$HELM_REPO_CHART_NAME' chart is present in local helm repo '$HELM_LOCAL_REPO_NAME'"
+  else
+    echo "'$HELM_REPO_CHART_NAME' chart is NOT present in local helm repo '$HELM_LOCAL_REPO_NAME'."
+    TEST_PASS=1
+  fi
+}
+
+install_helm() {
+  # install helm in /tmp directory
+  pushd /tmp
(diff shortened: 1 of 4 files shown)


scripts/lib/helm.sh
```diff
@@ -0,0 +1,99 @@
...
+  sleep 10
+}
+
+ensure_controller_pods() {
+  if ! should_execute ensure_controller_pods; then
+     return 1
+  fi
+
+  echo "Checking status of controller pods"
+  #TODO: make this function more fault tolerant with additional wait time of status is not Running.
```
A-Hilaly, line 91: Agreed. Maybe we can use a retry mechanism?

scripts/lib/helm.sh
```diff
@@ -0,0 +1,99 @@
...
+  echo "Sleeping 10 seconds for controller pods to come in Running state"
+  sleep 10
+}
+
+ensure_controller_pods() {
+  if ! should_execute ensure_controller_pods; then
+     return 1
+  fi
+
+  echo "Checking status of controller pods"
```
mhausenblas, line 90: Rather than manual polling, suggest to use the `wait` command, see for example https://medium.com/hackernoon/kubectl-tip-of-the-day-wait-like-a-boss-40a818c423ac

scripts/lib/helm.sh
```diff
@@ -0,0 +1,99 @@
...
+    echo "'$HELM_REPO_CHART_NAME' chart is NOT present in local helm repo '$HELM_LOCAL_REPO_NAME'."
+    TEST_PASS=1
+  fi
+}
+
+install_helm() {
+  # install helm in /tmp directory
+  pushd /tmp
+  # clone the source
+  git clone https://github.com/helm/helm.git
```
A-Hilaly, line 30: This is cloning the master branch, IMO it would be better if we use a stable release instead.

A-Hilaly, line 30: Maybe it's better to clone a stable release instead of the master branch?

A-Hilaly, line 30: Maybe it's safer to clone a stable release instead of the master branch?

vijtrip2: Very good point. Although they do not mention any specific branch/tag as stable.

I can pick any stable(pre-released) branch ATM, but do you have thoughts on how to keep it up to date, and avoid the risk of it being deprecated. 
I am guessing it has to be a periodical up-keeping item.

A-Hilaly: Good point 👍 , I agree we can use a pre-released version for now and it's better if can keep it up to date.
I can think of two solutions:
- We can inject a `$HELM_RELEASE` environment variable using CircleCI console and try to "manually" keep it up to date.
- clone the master branch and try to checkout the latest pre-release tag:
```bash
git fetch --tags
git checkout $(git tag -l | tail -1)
# this can also work to fetch latest remote tag:
# git ls-remote --tags --sort=committerdate | grep -o 'v.*' | sort -r | head -1
```

PR #62, opened 2020-06-22
Design proposal: cross account resource management

A-Hilaly:
Related to #3 

Description of changes: 
- Adding CARM (cross account resource management) design document


docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+
+With cross account resource management, a single kubernetes cluster with ACK installed can manage the resources of many different AWS accounts. Kubernetes administrators will not need to install and manage multiple clusters and ACK controllers, and AWS administrators will spend less effort managing permissions to access the services across accounts.
+
+## Proposed implementation
+
+The proposed solution will solve the problem of cross account management within a single kubernetes cluster. Given an AWS Owner account ID it will look up an IAM Role ARN and call STS:AssumeRole to pivot the AWS SDK calls in order to execute in a new AWS Account and role context.
```
nithu0115, line 11: @A-Hilaly, quick question - how will this work if we are planning to use Pod IRSA setup?

A-Hilaly: Thank you @nithu0115! I'm currently studying this case and will address it in my next commit.

docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+- Ability to have the service controller assume the same IAM role in a different account.
+- Tracebility and audibility of IAM Role related events
+- User has ability to specify the AWS account in CR Annotations
+- Unit and e2e tests with a specific ACK Controller (e.g. s3 bucket controller)
+
+## Out of scope
+
+- Any service controller specific code (e.g generated code)
+- CARM e2e tests for all controllers
+- Determination of owner account ID
```
jaypipes, line 34: right. an Owner Account ID will be already determined and passed to the `NewSession` function:

https://github.com/aws/aws-service-operator-k8s/blob/717d4ee368c4aec2b9ead5d11f4b6e89b3d082e6/pkg/runtime/session.go#L18

docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+## Potential other solutions
+
+- Store the IAM Role ARNs and account IDs as a kubernetes objects (new CRDs)
+- Hard coded configuration loaded at the start time (need to restart the controller to update the configuration)
+- Specify the role using an annotation (painful user experience, security problems)
+
+## In scope
+
+- Ability to have the service controller assume the same IAM role in a different account.
+- Tracebility and audibility of IAM Role related events
```
nithu0115, line 26: Does this mean, the controller will provide ability to audit/trace capabilities or  are you referring to the features of IAM here ?

docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+- User has ability to specify the AWS account in CR Annotations
```
nithu0115, line 27: I see you mentioned that annotation are painful user experience but you are also mentioning that this is in scope. Could you please clarify further why you thought it is painful experience ? May be a spec would help clear this to me.

docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+
+The proposed solution will solve the problem of cross account management within a single kubernetes cluster. Given an AWS Owner account ID it will look up an IAM Role ARN and call STS:AssumeRole to pivot the AWS SDK calls in order to execute in a new AWS Account and role context.
+
+Controllers will store AccountIDs and their associated IAM Role ARNs in a ConfigMap or Secret Kubernetes object. Cluster administrators will be responsible for administrating and updating the configmap/secret, and creating the target roles using the AWS Console/CLI. Controllers can reuse and existing IAM Role ARN if it matches the scope they need. 
+
+![carm-flowchart](./images/carm-flowchart.png)
+
+## Potential other solutions
+
+- Store the IAM Role ARNs and account IDs as a kubernetes objects (new CRDs)
```
nithu0115, line 19: IAM Role ARN format already has account ID, region information. Does user/admin required to pass Account ID(s) explicitly ? and why ?

docs/cross-account-resource-management.md
```diff
@@ -0,0 +1,56 @@
...
+
+With cross account resource management, a single kubernetes cluster with ACK installed can manage the resources of many different AWS accounts. Kubernetes administrators will not need to install and manage multiple clusters and ACK controllers, and AWS administrators will spend less effort managing permissions to access the services across accounts.
+
+## Proposed implementation
+
+The proposed solution will solve the problem of cross account management within a single kubernetes cluster. Given an AWS Owner account ID it will look up an IAM Role ARN and call STS:AssumeRole to pivot the AWS SDK calls in order to execute in a new AWS Account and role context.
+
+Controllers will store AccountIDs and their associated IAM Role ARNs in a ConfigMap or Secret Kubernetes object. Cluster administrators will be responsible for administrating and updating the configmap/secret, and creating the target roles using the AWS Console/CLI. Controllers can reuse and existing IAM Role ARN if it matches the scope they need.
```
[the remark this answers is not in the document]
A-Hilaly: Agreed 👍  . Let's use`ConfigMaps` for now, it's easier to manage for Kubernetes users

docs/cross-account-resource-management.md
A-Hilaly: > Does this mean, the controller will provide capabilities for user/Admin to audit/trace who assumed what ?
Exactly! controllers will log every IAM Account switch

docs/cross-account-resource-management.md
A-Hilaly: This is more for simplifying user/admin experience with ACK, they will only have to provide an Account ID. We can also support both methods. WDYT ?

nithu0115: Make sense! Thanks for clarifying.

docs/contents/proposals/cross-account-resource-management.md
```diff
@@ -0,0 +1,150 @@
...
+    123456789012=arn:aws:iam::123456789012:root
+    454545454545=arn:aws:iam::454545454545:role/S3Access
+    111111111111=arn:aws:iam::111111111111:instance-profile/Webserver
+```
+
+##### IAM Account pivot
+
+When a matching namespace is found, ACK service controllers will search for the IAM Role ARN associated with the AccountID in the `carm-account-ids` ConfigMap, and pivot the session to create resources in a new AWS account.
+
+The pivot is done by using the [aws-sdk-go](https://github.com/aws/aws-sdk-go) `session` package. `session.NewSessionWithOptions()` function will load the AWS related environment variables (injected by Pod IRSA) and initialize the [token provider](https://github.com/aws/aws-sdk-go/blob/master/aws/session/session.go#L200-L218) function. Later the token provider function is executed when [`AssumeRoleProvider.Retrieve`](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go#L329) is called and [return a set of credentials](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go#L357-L362) to make pivoted api calls.
```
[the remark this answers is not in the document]
A-Hilaly: Agreed it's almost the same function. Thanks for pointing this, i'll go for `session.NewSession()`

rothgar:
This is great! I love the namespace = account separation to allow for RBAC to be used for account separation.

Is there also an idea of (account, team) separation without using multiple clusters?

e.g. A developer on team-frontend has access to accounts prod, dev, stage but only for resources prefixed "team-frontend" (attribute based authentication). A dev on team-backend has access to the same accounts but uses a different IAM role to assume for "team-backend" resources.

I'm assuming in this case they would need to make 6 different namespaces for team-frontend-prod, team-frontend-dev, team-frontend-stage etc. The configMap would also need to specify account aliases instead of numbers for each RBAC role. (e.g. `123456789-frontend=arn::iam::123456789/Team-Frontend`). Is there an option to automatically create namespaces for each account specified. It would also be neat, and dangerous, to clean up namespaces (thus deleting AWS resources) when an account is removed from the configMap.

In a centralized ACK cluster I would see the need for developers to switch clusters frequently (ACK cluster for AWS resources, K8s cluster for each environment for pods).

Do we see a way ACK can make that easier at all?

docs/contents/proposals/cross-account-resource-management.md
```diff
@@ -0,0 +1,150 @@
...
+
+With cross account resource management, a single kubernetes cluster with ACK installed can manage the resources of many different AWS accounts. Kubernetes administrators will not need to install and manage multiple clusters and ACK controllers, and AWS administrators will spend less effort managing permissions to access the services across accounts.
+
+## Proposed implementation
+
+#### Overview
+
+The proposed solution will solve the problem of cross account management within a single kubernetes cluster. Given an AWS Owner account ID it will look up an IAM Role ARN and call **STS::AssumeRole** to pivot the AWS SDK calls in order to execute in a new AWS Account and role context.
+
+Controllers will store AccountIDs and their associated IAM Role ARNs in a `ConfigMap` Kubernetes object. Cluster administrators will be responsible for administrating and updating the `ConfigMap`, and creating the target roles using the AWS Console/CLI. Controllers can reuse and existing IAM Role ARN if it matches the scope they need.
```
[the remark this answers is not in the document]
A-Hilaly: Good point! I addressed your comment in my last [commit](https://github.com/A-Hilaly/aws-service-operator-k8s/blob/edd571ab697298011f0bc786cda4e07c8272b66f/docs/contents/design/proposals/cross-account-resource-management.md#region-determination).

docs/contents/design/proposals/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+##### Storing AWS Role ARNs
+
+AccountIDs and their associate AWS Role ARNs will be stored in a `ConfigMap` called "carm-account-ids". e.g:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: carm-account-ids
```
nithu0115, line 65: Can we rename this to `ack-carm-config-map` or something ? as it also has roles along with account IDs.

docs/contents/design/proposals/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+##### sequence diagram
+
+![sequence-diagram](../../images/carm-sequence-diagram.png)
+
+##### AWS Pod IRSA
+
+ACK Service controllers will be deployed using [AWS Pod IRSA (IAM Roles for service accounts)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), which will associate a ServiceAccount to the controller pods and result in limiting their permissions when calling the AWS API.
+Pod IRSA will limit ACK controllers API calls to the AWS service resources they are responsible for managing and provide audibility throught AWS CloudTrail.
+The injected IAM Role will be used to pivot the execution with a different target IAM Role using the STS::AssumeRole call.
+In addition to the injected AWS Credentials, Pod IRSA will also add `AWS_REGION` and `AWS_DEFAULT_REGION` to the controllers environment variables. When creating new resources and in case the default region isn't provided via annotations, the controllers will use these environement variables to set a default region for the AWS resources.
```
nithu0115, line 88: NIT: environement -> environment

docs/contents/design/proposals/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+We're not thinking of adding this feature, at least in early stages. As Justin said, it might be dangerous.
+
+> Is there also an idea of (account, team) separation without using multiple clusters?
+
+For the moment, no.
+
+> In a centralized ACK cluster I would see the need for developers to switch clusters frequently (ACK cluster for AWS resources, K8s cluster for each environment for pods). Do we see a way ACK can make that easier at all?
+
+It's a little hard to imagine a solution at this stage. We might tacle this issue in the future.
```
nithu0115, line 177: NIT: tacle -> tackle

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Proposed implementation
+
+#### Overview
+
+The proposed solution will solve the problem of cross account management within a single kubernetes cluster. Given an AWS Owner account ID it will look up an IAM Role ARN and call **STS::AssumeRole** to pivot the AWS SDK calls in order to execute in a new AWS Account and role context.
+
+Controllers will store AccountIDs and their associated IAM Role ARNs in a `ConfigMap` Kubernetes object. Cluster administrators will be responsible for administrating and updating the `ConfigMap`, and creating the target roles using the AWS Console/CLI. Controllers can reuse and existing IAM Role ARN if it matches the scope they need.
+
+To prevent unauthorized users from making any CRUD operations on resources in different AWS accounts, ACK will associate a unique Namespace to each AccountID, which will help administrators limit access to namespaces using K8s RBAC model, hence limit access to AccountIDs.
```
jaypipes, line 23: This paragraph gets the solution a bit backwards (though the sequence diagram below gets it correct). ACK will not associate a unique Namespace to each AccountID. Rather, the cluster administrator can set an AWS Owner Account ID as an annotation on each Namespace, and when this annotation is found by the ACK service controller, the value of the annotation will override the AWS Account ID used by the ACK service controller's IAM Role.

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Design details
+
+##### Namespace annotations
+
+To associate a namespace with a unique AWS account, ACK service controllers expect from cluster administrators to create namespaces with a specific annotation `services.k8s.aws/owner-account-id` e.g:
```
jaypipes, line 29: It is not a requirement that a cluster admin set this annotation. Rather, if the annotation is set on the Namespace, it acts as an override AWS Account ID for any ACK CRs created in that Namespace.

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: production-n1
+  annotations:
+    services.k8s.aws/default-aws-region: "eu-west-1"
+    services.k8s.aws/owner-account-id: "123456789012"
+```
+
+Each time a new AWS resource object is updated, ACK service controllers will lookup the namespaces annotations and decide in which namespace the AWS object should be created.
```
jaypipes, line 41: :+1:

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+In addition to the owner account id annotation, users can also specify a `services.k8s.aws/default-aws-region` annotation. This region will be used by default in case region annotation is missing in the CRDs specs.
```
jaypipes, line 43: :+1:

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+*//TODO(a-hilaly) address namespace/account-id conflict cases*
+
+*//TODO(a-hilaly) address namespace deletion case*
+
+##### Region determination
+
+To determine within which region the resources should be created, ACK controllers will, in order, look for a region in the following sources: 
+
+- Resource region annotation `services.k8s.aws/aws-region`. If provided it will override the namespace default region annotation.
```
jaypipes, line 53: maybe replace 'Resource' here with "CR"?

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ack-carm-config-map
+  namespace: default
+data:
+  accounts: |
+    123456789012=arn:aws:iam::123456789012:root
+    454545454545=arn:aws:iam::454545454545:role/S3Access
```
jaypipes, line 70: Recommend using the "property-type" keys in the `ConfigMap` YAML instead of the "file-type" keys:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: ack-carm-config-map
  namespace: default
data:
  "123456789012": arn:aws:iam::123456789012:root
  "454545454545": arn:aws:iam::454545454545:role/S3Access
```

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
```
jaypipes, line 72: TODO(amine-hilaly): Add more to this section. In particular, I'd suggest adding a paragraph describing, briefly, how the ACK runtime will need to:

* Query for the value of a particular key in a specific ConfigMap object via the Kubernetes API/client
* Watch for changes in the specific ConfigMap so that the controller is notified when new key entries are added or changed.

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+##### IAM Account pivot
+
+When a matching namespace is found, ACK service controllers will search for the IAM Role ARN associated with the AccountID in the `ack-carm-config-map` ConfigMap, and pivot the session to create resources in a new AWS account.
+
+The pivot is done by using the [aws-sdk-go](https://github.com/aws/aws-sdk-go) `session` package. `session.NewSession()` function will load the AWS related environment variables (injected by Pod IRSA) and initialize the [token provider](https://github.com/aws/aws-sdk-go/blob/master/aws/session/session.go#L200-L218) function. Later the token provider function is executed when [`AssumeRoleProvider.Retrieve`](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go#L329) is called and [return a set of credentials](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/stscreds/assume_role_provider.go#L357-L362) to make pivoted api calls.
```
jaypipes, line 77: Super! Feel free to fill in more here, including potentially some sample code snippets.

docs/design/proposals/carm/cross-account-resource-management.md
```diff
@@ -0,0 +1,177 @@
...
+
+ACK service controllers reconcile loops will log every session account pivot using zap logger values e.g:
+
+```golang
+func (r *reconciler) reconcile(req ctrlrt.Request) error {
+  ...
+
+  acctID := r.getOwnerAccountID(res)
+  r.log.WithValues(
+    "account_id", acctID,
```
jaypipes, line 100: what about the IAM Role ARN used in creating the aws-sdk-go client Session? Should we log that as well? :)

PR #65, opened 2020-06-24
ack generate controller Dockerfile

varun1524:
Issue #, if available:

Description of changes:
generate Dockerfile on "ack generate controller servicealias"

### DIFF
--- a/templates/Dockerfile.tpl
+++ b/templates/Dockerfile.tpl
@@ -1,0 +1,31 @@
+{{ template "boilerplate_hash" }}
+
+# Build the manager binary
+FROM golang:1.14.1 as builder
+
+ARG work_dir=/github.com/aws/aws-service-operator-k8s
+WORKDIR $work_dir
+# For building Go Module required
+ENV GOPROXY=direct
+ENV GO111MODULE=on
+ENV GOARCH=amd64
+ENV GOOS=linux
+ENV CGO_ENABLED=0
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN  go mod download
+# Copy the go source
+# TODO: for now copy whole repository. Later once autogeneration is ready, need to determine which dependencies to take in
+COPY . $work_dir/
+# Build
+RUN  go build -a -o $work_dir/bin/controller $work_dir/services/{{ .ServiceAlias }}/cmd/controller/main.go
+
+
+FROM amazonlinux:2
+ARG work_dir=/github.com/aws/aws-service-operator-k8s
+WORKDIR /
+COPY --from=builder $work_dir/bin/controller /bin/.
+ENTRYPOINT ["/bin/controller"]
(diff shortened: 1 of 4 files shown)


templates/Dockerfile.tpl
```diff
@@ -0,0 +1,31 @@
+{{ template "boilerplate_hash" }}
+
+# Build the manager binary
+FROM golang:1.14.1 as builder
+
+ARG work_dir=/github.com/aws/aws-service-operator-k8s
```
jaypipes, line 6: probably should have noticed this in the commit that added the Dockerfile, but I think the work_dir should be `go/src/github.com/aws/aws-service-operator-k8s/services/{{ .ServiceAlias }}`. The `go/src/` prefix is the default `GOPATH` for the golang builder image and we want to set the package path to the specific service controller...

templates/Dockerfile.tpl
```diff
@@ -0,0 +1,31 @@
...
+ENV GOOS=linux
+ENV CGO_ENABLED=0
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN  go mod download
+# Copy the go source
+# TODO: for now copy whole repository. Later once autogeneration is ready, need to determine which dependencies to take in
```
jaypipes, line 21: Let's change this now. This should be:

```
COPY services/{{ .ServiceAlias }}/. $work_dir
```

templates/Dockerfile.tpl
```diff
@@ -0,0 +1,31 @@
...
+COPY . $work_dir/
+# Build
+RUN  go build -a -o $work_dir/bin/controller $work_dir/services/{{ .ServiceAlias }}/cmd/controller/main.go
```
jaypipes, line 24: and the above would just be:

```
RUN  go build -a -o $work_dir/bin/controller $work_dir/cmd/controller/main.go
```

templates/Dockerfile.tpl
```diff
@@ -0,0 +1,31 @@
...
+
```
jaypipes, line 25: Note to self: we should copy the LICENSE and a future ATTRIBUTION.txt file directly into the image as well.

templates/Dockerfile.tpl
varun1524: I wasn't sure what will be consumed until complete implementation is ready, so kept it as TODO. 

As an example, runtime
https://github.com/aws/aws-service-operator-k8s/tree/mvp/pkg/runtime

Also we will have to maintain paths as following:
from source of root we have to copy:
COPY services/{{ .ServiceAlias }}/.  $work_dir/{{ .ServiceAlias }}/.
COPY pkg/runtime $work_dir/pkg/runtime
and more dependencies that we may need.


And Build will become this:
RUN  go build -a -o $work_dir/bin/controller $work_dir/{{ .ServiceAlias }}/cmd/controller/main.go

jaypipes: yeah, that makes sense... OK, for now let's keep it like this and refactor later.

Issue #64, opened 2020-06-23
Array types not handled properly in `ack-generate apis`


jaypipes:
While generating CRD type definitions for some APIs, I've noticed the following in some of the generated structs:

```go
type ImageScanFindings struct `aws:"ImageScanFindings"` {
	FindingSeverityCounts map[string]string `json:"findingSeverityCounts,omitempty" aws:"findingSeverityCounts"`
	Findings []!!! UNKNOWN ARRAY OBJECT TYPE !!! `json:"findings,omitempty" aws:"findings"`
	ImageScanCompletedAt string `json:"imageScanCompletedAt,omitempty" aws:"imageScanCompletedAt"`
	VulnerabilitySourceUpdatedAt string `json:"vulnerabilitySourceUpdatedAt,omitempty" aws:"vulnerabilitySourceUpdatedAt"`
}
```

when I take a look at the OpenAPI3 Schema for the API (in the above case, the ECR API), I find this:

```yaml
    ImageScanFindings:
      properties:
        findingSeverityCounts:
          $ref: '#/components/schemas/FindingSeverityCounts'
        findings:
          $ref: '#/components/schemas/ImageScanFindingList'
        imageScanCompletedAt:
          $ref: '#/components/schemas/ScanTimestamp'
        vulnerabilitySourceUpdatedAt:
          $ref: '#/components/schemas/VulnerabilitySourceUpdateTimestamp'
      type: object
```

You will note that the `findings` property points to an `ImageScanFindlingList` schema, which looks like this:

```yaml
    ImageScanFindingList:
      items:
        properties:
          attributes:
            $ref: '#/components/schemas/AttributeList'
          description:
            $ref: '#/components/schemas/FindingDescription'
          name:
            $ref: '#/components/schemas/FindingName'
          severity:
            $ref: '#/components/schemas/FindingSeverity'
          uri:
            $ref: '#/components/schemas/Url'
        type: object
      type: array
```

the above is an `array` of `object` with a set of properties (`attributes`, `description`, `name`, `severity`, and `uri`). Not coincidentally, that object schema is the same as `ImageScanFinding`'s schema:

```yaml
    ImageScanFinding:
      properties:
        attributes:
          $ref: '#/components/schemas/AttributeList'
        description:
          $ref: '#/components/schemas/FindingDescription'
        name:
          $ref: '#/components/schemas/FindingName'
        severity:
          $ref: '#/components/schemas/FindingSeverity'
        uri:
[...]

PR #66, opened 2020-06-25
correct lookup of referred-to object schemas

jaypipes:
When determining the Go type for schemas of type "object" or "array of
object", we were searching for the referred-to schema using the wrong
name. For the case of schemas with a type of "object", we were passing
the name of the property, camel-cased. In the case of the "array of
object" type schemas, we were using the name of the property instead of
the name of the schema being referred to.

What this meant was that, for example, the ECR API's ImageScanFindings
object's "findings" property, which looks like this:

```yaml
    ImageScanFindings:
      properties:
        findingSeverityCounts:
          $ref: '#/components/schemas/FindingSeverityCounts'
        findings:
          $ref: '#/components/schemas/ImageScanFindingList'
        imageScanCompletedAt:
          $ref: '#/components/schemas/ScanTimestamp'
        vulnerabilitySourceUpdatedAt:
          $ref: '#/components/schemas/VulnerabilitySourceUpdateTimestamp'
      type: object
    ImageScanFindingList:
      items:
        properties:
          attributes:
            $ref: '#/components/schemas/AttributeList'
          description:
            $ref: '#/components/schemas/FindingDescription'
          name:
            $ref: '#/components/schemas/FindingName'
          severity:
            $ref: '#/components/schemas/FindingSeverity'
          uri:
            $ref: '#/components/schemas/Url'
        type: object
      type: array
    ImageScanFinding:
      properties:
        attributes:
          $ref: '#/components/schemas/AttributeList'
        description:
          $ref: '#/components/schemas/FindingDescription'
        name:
          $ref: '#/components/schemas/FindingName'
        severity:
          $ref: '#/components/schemas/FindingSeverity'
        uri:
          $ref: '#/components/schemas/Url'
      type: object
```

we were looking for a schema named "Findings" instead of
"ImageScanFindings".

Fixes Issue #64

Issue #, if available:

Description of changes:

### DIFF
--- a/pkg/schema/util.go
+++ b/pkg/schema/util.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/aws/aws-service-operator-k8s/pkg/names"
 	"github.com/gertd/go-pluralize"
 	"github.com/getkin/kin-openapi/openapi3"
 )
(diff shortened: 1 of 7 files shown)


pkg/schema/util.go
```diff
@@ -69,22 +72,83 @@ func (h *Helper) getGoTypeFromSchema(
...
+		} else {
+			itemsSchemaRefName = strings.TrimPrefix(itemsSchemaRefName, compSchemasRef)
+		}
 		itemsSchema := h.getSchemaFromSchemaRef(schema.Items)
 		var itemType string
 		if itemsSchema.Type == "object" {
-			itemType = h.deduceArrayOfObjectsType(schemaName, itemsSchema)
+			itemType = h.deduceArrayOfObjectsType(itemsSchemaRefName, itemsSchema)
 			if itemType == "" {
 				itemType = "!!! UNKNOWN ARRAY OBJECT TYPE !!!"
```
alina-kim, line 141: Would it be possible to return an error along with the resulting string, and the itemType be empty instead?

pkg/schema/util_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/aws/aws-service-operator-k8s/pkg/names"
+	"github.com/aws/aws-service-operator-k8s/pkg/testutil"
+)
+
+func TestGoTypeFromSchemaRef(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	tests := []struct {
```
alina-kim, line 30: Are these all happy paths / will there be tests intentionally leading to the placeholder types or return values (the "unknown" strings) added later? Or do those parts never get executed?

jaypipes: they are indeed all the happy paths, good catch, Alina! I can add some of the unhappy paths to the tests.

pkg/schema/util.go
jaypipes: yeah, totally. I'll fix up.

jaypipes:
@alina-kim thanks again for your code review earlier and for keeping me honest :) Added non-happy-path tests and am now returning any errors all the way up the stack for Go type introspection.

pkg/schema/util.go
```diff
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/aws/aws-service-operator-k8s/pkg/names"
```
A-Hilaly, line 21: nit: this import path can be placed after external packages imports paths

nithu0115:
LGTM 👍

mhausenblas:
I'm all fine with this, just one thing (not specific to this commit but in general, going forward): I noticed that we have a lot of `fmt.Errorf()` returning where I think we could benefit from the fact that in Go [errors are values](https://blog.golang.org/errors-are-values). In other words: I think now is a good time to start defining our own error types ;)

PR #68, opened 2020-06-25
Setting Logger with configurable debug option

varun1524:
Issue #, if available:

Description of changes:

### DIFF
--- a/pkg/runtime/config.go
+++ b/pkg/runtime/config.go
@@ -15,12 +15,16 @@ package runtime
 
 import (
 	flag "github.com/spf13/pflag"
+	ctrlrt "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
 )
 
 type Config struct {
 	BindPort             int
 	MetricsAddr          string
 	EnableLeaderElection bool
+	EnableDevelopmentLogging bool
 }
 
 func (c *Config) BindFlags() {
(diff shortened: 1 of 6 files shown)


pkg/runtime/config.go
```diff
@@ -40,4 +41,10 @@ func (c *Config) BindFlags() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.",
 	)
+	flag.BoolVar(
+		&c.EnableDevModeLogging, "enable-dev-mode-logging",
+		false,
+		"Configures the logger to use a Zap development config (stacktraces on warnings, no sampling), "+
```
nithu0115, line 47: Nit: Probably, we can add more info about logLevel as well, something like this: `Development Mode, defaults(encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn, no sampling) + "Production Mode defaults(encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error), sampling")`

varun1524: sure. looks better

services/bookstore/cmd/controller/main.go
```diff
@@ -42,6 +43,10 @@ func main() {
 	var ackCfg ackrt.Config
 	ackCfg.BindFlags()
 	flag.Parse()
+	zapOptions := zap.Options{
+		Development: ackCfg.EnableDevModeLogging,
+	}
+	ctrlrt.SetLogger(zap.New(zap.UseFlagOptions(&zapOptions)))
```
jaypipes, line 49: We will want to hide all of this inside an `ackCfg.SetupLogger()` method. The benefits of doing that are:

* No need to import the zap library above
* There will only be a single call added to the main.go template and all generated controllers (easier maintenance)

pkg/runtime/config.go
```diff
@@ -40,4 +41,10 @@ func (c *Config) BindFlags() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.",
 	)
+	flag.BoolVar(
+		&c.EnableDevModeLogging, "enable-dev-mode-logging",
```
jaypipes, line 45: I'd prefer naming the option `enable-development-logging`. The shortened term "dev" is easily confused with "device" :)

varun1524: done,

services/petstore/cmd/controller/main.go
```diff
@@ -14,8 +14,6 @@
 package main
 
 import (
-	"os"
```
jaypipes, line 17: Looks like this was accidentally removed, though?

varun1524: yea fixed

PR #78, opened 2020-07-10
use aws-sdk-go private/model/api for generation

jaypipes:
aws-sdk-go has a private/model/api package that contains lots of utility
code, parsing functions (for the api-2.json files in aws-sdk-go) and,
frankly, a bunch of helpful stuff that will obviate the need for a good
chunk of the OpenAPI3 schema-wrangling that I'm doing in the current
ACK pkg/model package.

This PR replaces the use of OpenAPI3/Swagger documents for introspecting the AWS
service API model with the aws-sdk-go private/model/api package. This
aws-sdk-go private/model/api package already handles most of the munging
and wrangling that I was doing with OpenAPI3 Schema document walking, so
let's use it to reduce the amount of code we need.

Because private/model/api package wasn't ever really intended for
consumption, I had to place a few hacks into the ACK pkg/model package
to get around some bugs I stumbled across (including panic()s from
non-initialized `map` fields in the primary `API` struct). But overall,
using aws-sdk-go private/model/api has resulted in much less code for
the API typedef generation. And since the private/model/api.API object
contains all the linkage/knowledge about the aws-sdk-go client/API
generated structs and methods (like CreateXXXWithContext,
CreateXXXRequest, etc), implementing the controller's resource
manager.go files with aws-sdk-go's private/model/api package will be
much, much easier.

Because the aws-sdk-go private/model Go package is gated with a
"codegen" Go build tag, we need to pass that to any `go build` or
`go test` calls now, so I've created a new Makefile target called
`build-ack-generate` that does the needful.

Each API acts differently and the input/output shapes for individual API
calls follow different, inconsistent conventions. We need a way to
standardize many of these inconsistencies.
    
This PR adds a new SDKMapper object to the pkg/schema package. This
struct's purpose is to contain instructions to the code generator for
generating the SDK calls in the resource manager.
    
The first standardization that the SDKMapper helps with is identifying
the fields in an Operation's output shape that contain the primary
resource's ARN information. In constructing the model.CRD now, in
pkg/schema's `Helper.GetCRDs()` method, we now remove from the list of
the CRD's Status fields any field that has the same name as a field in
the corresponding Spec struct. In addition, if we find a field called
"arn" or "${RESOURCE}arn" in the output shape, we indicate to the
SDKMapper that this field stores the primary object's ARN information
and remove it from the Status fields.
    
See new tests of the SNS Topic, EC2 LaunchTemplate and ECR Repository's
CRD generation and lots of comments in the test assertions for more
information on how these normalizations occur.


jaypipes:
Please note that the majority of the line of code added in this PR are from the new CORAL model files added in `pkg/schema/testdata/models/apis`. Those were added in a separate commit, so you can review the PR commit-by-commit if you'd like

pkg/model/enum_def.go
```diff
@@ -30,29 +29,15 @@ type EnumValue struct {
...
 	Names  names.Names
-	GoType string
 	Values []EnumValue
 }
 
-func NewEnumDef(names names.Names, goType string, values []interface{}) (*EnumDef, error) {
+func NewEnumDef(names names.Names, values []string) (*EnumDef, error) {
 	enumVals := make([]EnumValue, len(values))
 	for x, item := range values {
-		if goType == "string" {
```
jaypipes, line 40: Note to reviewers: there are only string enums in the AWS service API model definitions. Thus, I removed the GoType stuff here and simplified things.

cmd/ack-generate/command/common.go
```diff
@@ -16,9 +16,14 @@ package command
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 )
 
+const (
+	sdkRepoURL = "https://github.com/aws/aws-sdk-go"
```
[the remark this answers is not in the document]
jaypipes: ya, good point. I'll add that in a followup patch.

cmd/ack-generate/command/apis.go
```diff
@@ -72,7 +65,7 @@ func generateAPIs(cmd *cobra.Command, args []string) error {
 	if len(args) != 1 {
```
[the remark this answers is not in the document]
jaypipes: true, but this is the function signature that is required for Cobra ExecE attributes :)

cmd/ack-generate/command/controller.go
```diff
@@ -72,10 +71,13 @@ func generateController(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	sh, err := getSchemaHelper()
+	ensureSDKRepo(optCacheDir)
```
A-Hilaly, line 74: i think error need to be handled here

cmd/ack-generate/command/common.go
```diff
@@ -51,3 +56,30 @@ func isDirWriteable(fp string) bool {
...
+// cloneSDKRepo git clone's the aws-sdk-go source repo into the cache and
+// returns the filepath to the clone'd repo
+func cloneSDKRepo(srcPath string) (string, error) {
+	clonePath := filepath.Join(srcPath, "aws-sdk-go")
+	if _, err := os.Stat(clonePath); os.IsNotExist(err) {
+		cmd := exec.Command("git", "clone", "--depth", "1", sdkRepoURL, clonePath)
+		return clonePath, cmd.Run()
+	}
+	return clonePath, nil
+}
```
A-Hilaly, line 85: Maybe we should also run a a `git pull` if directory already exist

templates/pkg/crd_manager.go.tpl
```diff
@@ -0,0 +1,130 @@
+{{ template "boilerplate" }}
```
A-Hilaly, line 1: nit: adding a comment - similar to what `controller-gen` does - to the boilerplate saying `// Code generated by ack-generate. DO NOT EDIT.`  might be useful

cmd/ack-generate/command/common.go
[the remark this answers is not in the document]
jaypipes: Technically, srcPath is not hard-coded. It's derived from `cacheDir` which is configurable via the `--cache-dir` CLI flag.

jaypipes: Perhaps in a followup patch, sure. Left this simple for now :)

templates/pkg/crd_manager.go.tpl
jaypipes: ah, yes great point. lemme update.

Issue #92, opened 2020-07-22
Cannot deepcopy-gen `time.Time` fields


jaypipes:
Example output when building RDS service controller after generating the API types:

```
# github.com/aws/aws-controllers-k8s/services/rds/apis/v1alpha1
services/rds/apis/v1alpha1/zz_generated.deepcopy.go:134:8: (*in).DeepCopyInto undefined (type *time.Time has no field or method DeepCopyInto)
```

Turns out we will need to replace all `*time.Time` fields with `*metav1.Time` fields.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go

`metav1.Time` is basically `time.Time` that has been manually deepcopy-gen'd

PR #93, opened 2020-07-23
set input/output Go code for member fields

jaypipes:
This patch adds support for recursively setting member field values for
SDK input shape structs and CR Status fields. For member field values
that are themselves structs, we create a variable of type equal to the
SDK shape corresponding to the member field. We then set the member
fields of *that* variable using the SDK shape's SetXXX() methods.
    
For example, suppose we have an SDK Shape called `CreateBookInput` that
has a `Book` field that is of type `BookData`, which itself has fields
that are further nested shapes representing the book's Author and the
Author's address. This patch modifies the code generation to construct a
`CreateBookInput` shape like so:
    
```go
res := &sdkapi.CreateBookInput{}
tmp0 := &sdkapi.BookData{}
tmp0f0 := &sdkapi.Author{}
tmp0f0f0 := &sdkapi.Address{}
tmp0f0f0.SetStreet(*ko.Spec.Author.Address.Street)
tmp0f0f0.SetCity(*ko.Spec.Author.Address.City)
tmp0f0f0.SetState(*ko.Spec.Author.Address.State)
tmp0f0.Address = tmp0f0f0
tmp0f0.SetName(*r.ko.Author.Name)
tmp0.Author = tmp0f0
res.Book = tmp0
```
    
Likewise, for the Go code that sets a CR's `Status` fields, we now
generate something like the following:
    
```go
ko = r.ko.DeepCopy()
tmp0 := &svcapitypes.ImageData{}
tmp0.Location = resp.ImageData.Location
tmp0.Tag = resp.ImageData.Tag
ko.Status.ImageData = tmp0
ko.Status.Name = resp.Name
```

Also in this PR:

Upstream apimachinery/metav1 package has a `Time` struct that has
`DeepCopy()` implementation already generated for it. Replace all
`time.Time` fields with `metav1.Time` fields.
    
Fixes Issue #92


jaypipes:
Note to reviewers: I know this code is not pretty. It's tough to read and I will work on making it easier to consume in a later patch. For now at least, this PR should resolve some of the service controller generation/build errors we were seeing, though there are still issues with the validity of the generated code in the `sdk.go` files. I'll tackle more of those issues tomorrow.

pkg/model/crd_test.go
```diff
@@ -188,6 +368,12 @@ func TestEC2LaunchTemplate(t *testing.T) {
 	ko.Status.LatestVersionNumber = resp.LaunchTemplate.LatestVersionNumber
 	ko.Status.LaunchTemplateID = resp.LaunchTemplate.LaunchTemplateId
 	tmp0 := []*svcapitypes.Tag{}
+	for _, elem0 := range ko.Status.Tags {
+		tmpElem0 := &svcapitypes.Tag{}
+		tmpElem0.Key = elem0
+		tmpElem0.Value = elem0
```
jaypipes, line 374: This above is wrong. Should be generating this:

```go
	tmpElem0.Key = elem0.Key
	tmpElem0.Value = elem0.Value
```

Issue #98, opened 2020-07-23
Need to convert metav1.Time to time.Time when setting SDK struct fields


jaypipes:
Generated the ElastiCache service controller and see this when calling `make test`:

```
services/elasticache/pkg/resource/replication_group/sdk.go:66:38: cannot use resp.ReplicationGroup.AuthTokenLastModifiedDate (type *time.Time) as type *"k8s.io/apimachinery/pkg/apis/meta/v1".Time in assignment
```

I have seen this in other generated service controllers as well. Basically, we need to "unpack" the wrapped `time.Time` that is within the `metav1.Time` struct:

https://github.com/kubernetes/kubernetes/blob/1ed2cf189571bf0264822adf0cfe88007a30e934/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/time.go#L34

Obviously, when assigning SDK output Shape fields that are of type `time.Time` to CRD fields that are of type `metav1.Time`, we'll need to do a similar conversion the other way, to remedy build failures like this:

```
services/elasticache/pkg/resource/cache_cluster/sdk.go:67:38: cannot use resp.CacheCluster.AuthTokenLastModifiedDate (type *time.Time) as type *"k8s.io/apimachinery/pkg/apis/meta/v1".Time in assignment
```

jaypipes:
Still seeing issues with `time.Time` fields:

```
services/apigatewayv2/pkg/resource/domain_name/sdk.go:114:35: invalid indirect of f1iter.CertificateUploadDate.Time (type time.Time)
```

Issue #101, opened 2020-07-23
Ambiguity in generated sdk.go file for controllers.


vijtrip2:
**Describe the bug**
Some of the autogenerated code throws error when building controller's main.go

**Steps to reproduce**
* Add '_ "github.com/aws/aws-controllers-k8s/services/apigatewayv2/pkg/resource/api"' to main.go file so that resourceManagerFactory is initialized.
* run go build -o ../../bin/controller main.go
* Following errors show up apart from https://github.com/aws/aws-controllers-k8s/issues/98
```
# github.com/aws/aws-controllers-k8s/services/apigatewayv2/pkg/resource/api
../../pkg/resource/api/sdk.go:61:3: undefined: tmpElem0
../../pkg/resource/api/sdk.go:61:36: elem0.ImportInfo undefined (type *string has no field or method ImportInfo)
../../pkg/resource/api/sdk.go:62:23: undefined: tmpElem0
../../pkg/resource/api/sdk.go:67:3: undefined: tmpElem1
../../pkg/resource/api/sdk.go:67:36: elem1.Warnings undefined (type *string has no field or method Warnings)
../../pkg/resource/api/sdk.go:68:23: undefined: tmpElem1
../../pkg/resource/api/sdk.go:108:3: undefined: tmpElem0
../../pkg/resource/api/sdk.go:108:36: elem0.ImportInfo undefined (type *string has no field or method ImportInfo)
../../pkg/resource/api/sdk.go:108:36: too many errors
```

* iteration should be performed on responseObject and not Ko.Status.ImportInfo
```
tmp0 := []*string{}
	for _, elem0 := range ko.Status.ImportInfo {
		tmpElem0.__listOf__string = elem0.ImportInfo
		tmp0 = append(tmp0, tmpElem0)
	}
	ko.Status.ImportInfo = tmp0
```

**Expected outcome**
Controller's main.go to build successfully

**Environment**

* Kubernetes version 1.14
* Using EKS (yes/no), if so version?  yes
* AWS service targeted (S3, RDS, etc.) apigatewayv2

vijtrip2:
The above code fixes the original issue, but following is one more case where the generated code fails to build.

This bug is surfacing in sdkFind method for apigatewayv2 api resource. 
../../pkg/resource/api/sdk.go:64:3: undefined: f7elem


```
        ko.Status.APIEndpoint = resp.ApiEndpoint
	ko.Status.APIID = resp.ApiId
[...]

PR #108, opened 2020-07-24
correct sdk.go code generation

jaypipes:
Fixes a series of mistakes in the sdk.go generation of struct-fields and
array-fields for the Go code that constructs and initializes SDK Input
shapes as well as CRD Status structs.

we construct variables containing temporary storage for sub-elements
and sub-fields that are structs. Names of fields are "f" appended by
the 0-based index of the field within the set of a struct's set of
fields. Nested structs simply append another "f" and the field index
to the variable name.

This means you can tell what field a temporary fields variable
represents by the name.

For example, the field variable name "f0f5f2", it contains the third
field of the sixth field of the first field of the input shape being
constructed.

If we have two levels of nested struct fields, we will end
up with a targetVarName of "field0f0f0" and the generated code
might look something like this:

```go
res := &sdkapi.CreateBookInput{}
f0 := &sdkapi.BookData{}
f0f0 := &sdkapi.Author{}
f0f0f0 := &sdkapi.Address{}
f0f0f0.SetStreet(*ko.Spec.Author.Address.Street)
f0f0f0.SetCity(*ko.Spec.Author.Address.City)
f0f0f0.SetState(*ko.Spec.Author.Address.State)
f0f0.Address = field0f0f0
f0f0.SetName(*r.ko.Author.Name)
f0.Author = field0f0
res.Book = field0
```

It's ugly but at least consistent and mostly readable...

For populating list fields, we need an iterator and a temporary
element variable. We name these "{fieldName}iter" and
"{fieldName}elem" respectively. For nested levels, the names will be
progressively longer.

For list fields, we want to end up with something like this:

```
res := &sdkapi.CreateCustomAvailabilityZoneInput{}
f0 := []*sdkapi.VpnGroupMembership{}
for _, f0iter := ko.Spec.VPNGroupMemberships {
    f0elem := &sdkapi.VpnGroupMembership{}
    f0elem.SetVpnId(f0elem.VPNID)
    f0 := append(f0, f0elem)
}
res.VpnMemberships = f0
```

Issue #101

Code generated to set SDK shape and CRD struct field values for
"timestamp" types was not properly unwrapping the wrapped time.Time
value from the wrapper metav1.Time struct. This meant we were running
into build errors that looked like this:

```
cannot use resp.ReplicationGroup.AuthTokenLastModifiedDate (type *time.Time) as type *"k8s.io/apimachinery/pkg/apis/meta/v1".Time in assignment
cannot use resp.CacheCluster.AuthTokenLastModifiedDate (type *time.Time) as type *"k8s.io/apimachinery/pkg/apis/meta/v1".Time in assignment
```

This patch addresses that by special-casing the code generated for
setting CRD field or SDK type definition fields for timestamp types to
either pop the wrapped time.Time or wrapping a time.Time in a
metav1.Time.

Fixes Issue #98

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -170,17 +170,65 @@ func (r *CRD) SpecFieldNames() []string {
 // GoCodeSetInput returns the Go code that sets an input shape's member fields
 // from a CRD's Spec fields.
 //
-// We loop through the Spec fields, outputting code that looks something like
-// this:
+// Assume a CRD called Repository that looks like this pseudo-schema:
 //
-//   res.SetAttributes(r.ko.Spec.Attributes)
-//   res.SetName(*r.ko.Spec.Name)
-//   tmp0 := []*svcsdk.Tag{}
-//   res.Tags = tmp0
+// .Status
+//   .Authors ([]*string)
+//   .ImageData
+//     .Location (*string)
+//     .Tag (*string)
+//   .Name (*string)
+//
+// And assume an SDK Shape CreateRepositoryInput that looks like this
+// pseudo-schema:
+//
+// .Repository
+//   .Authors ([]*string)
+//   .ImageData
+//     .Location (*string)
+//     .Tag (*string)
+//   .Name
+//
+// This function is called from a template that generates the Go code that
+// represents linkage between the Kubernetes objects (CRs) and the aws-sdk-go
+// (SDK) objects. If we call this function with the following parameters:
+//
+//  opType:			OpTypeCreate
+//  sourceVarName:	ko.Spec
+//  targetVarName:	res
+//  indentLevel:	1
+//
+// Then this function should output something like this:
+//
+//   field1 := []*string{}
+//   for _, elem0 := range r.ko.Spec.Authors {
+//       elem0 := &string{*elem0}
+//       field0 = append(field0, elem0)
+//   }
+//   res.Authors = field1
+//   field1 := &svcsdk.ImageData{}
+//   field1.SetLocation(*r.ko.Spec.ImageData.Location)
+//   field1.SetTag(*r.ko.Spec.ImageData.Tag)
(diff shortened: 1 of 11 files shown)


pkg/model/crd.go
```diff
@@ -170,17 +170,65 @@ func (r *CRD) SpecFieldNames() []string {
...
+// (SDK) objects. If we call this function with the following parameters:
+//
+//  opType:			OpTypeCreate
+//  sourceVarName:	ko.Spec
+//  targetVarName:	res
+//  indentLevel:	1
+//
+// Then this function should output something like this:
+//
+//   field1 := []*string{}
```
vijtrip2, line 203: nit: it should be "field0 := []*string{}" right?

pkg/model/crd.go
```diff
@@ -207,110 +255,110 @@ func (r *CRD) GoCodeSetInput(
...
+		//
+		// res := &sdkapi.CreateBookInput{}
+		// f0 := &sdkapi.BookData{}
+		// f0f0 := &sdkapi.Author{}
+		// f0f0f0 := &sdkapi.Address{}
+		// f0f0f0.SetStreet(*ko.Spec.Author.Address.Street)
+		// f0f0f0.SetCity(*ko.Spec.Author.Address.City)
+		// f0f0f0.SetState(*ko.Spec.Author.Address.State)
+		// f0f0.Address = field0f0f0
+		// f0f0.SetName(*r.ko.Author.Name)
```
vijtrip2, line 293: 'field' -> 'f' ?

pkg/model/crd.go
jaypipes: nice catch, eagle eyes! :)

pkg/model/crd.go
jaypipes: yup. ++

Issue #109, opened 2020-07-26
Develop pattern for "unpacking" Attribute maps into actual CRD Fields


jaypipes:
The SNS API has a [`CreateTopic`](https://docs.aws.amazon.com/sns/latest/api/API_CreateTopic.html) API call.

The HTTP request seems simple. There's a `Name` query string parameter for the topic name and one of more "Attributes" query string parameter pairs. These attribute parameter pairs are a `map[string]*string` that ends up being added as a CRDField with the name `Attributes` in the `Topic` CRD's `TopicSpec` struct.

The problem is that this "Attributes" field isn't a free-form `map[string]string` like AWS Resource Tags. No, no. These are actually specific properties of the Topic. These properties can be set when creating a Topic:

* `DeliveryPolicy` – The policy that defines how Amazon SNS retries failed deliveries to HTTP/S endpoints.
* `DisplayName` – The display name to use for a topic with SMS subscriptions.
* `Policy` – The policy that defines who can access your topic. By default, only the topic owner can publish or subscribe to the topic.
* `KmsMasterKeyId` - The ID of an AWS-managed customer master key (CMK) for Amazon SNS or a custom CMK.

The DeliveryPology` and `Policy` attributes have other constraints on the values of the attribute; in other words, they are Enums.

Even more problematic is that there are a set of **Read Only** attributes that are returned from the `GetTopicAttributes` API call:

* `Owner` – The AWS account ID of the topic's owner.
* `SubscriptionsConfirmed` – The number of confirmed subscriptions for the topic.
* `SubscriptionsDeleted` – The number of deleted subscriptions for the topic.
* `SubscriptionsPending` – The number of subscriptions pending confirmation for the topic.
* `TopicArn` – The topic's ARN.
* `EffectiveDeliveryPolicy` – The JSON serialization of the effective delivery policy, taking system defaults into account.

The reason that these "Attributes" are problematic is that the Go type of the `aws-sdk-go` `CreateTopicInput.Attributes` is just a `map[string]*string` but the contents of this map is actually real, constrained schema properies, not just a free-form map of strings values.

For APIs like this that masquerade real, schema'd properties as a free-form `map[string]*string`, we should make the end-user experience for Kubernetes/ACK users better.

Instead of a `Topic` CR that looks like this:

```yaml
apiVersion: v1alpha1
kind: sns.services.k8s.aws
metadata:
  name: MyTopic
  namespace: my-app
spec:
  name: MyTopic
  attributes:
[...]

PR #112, opened 2020-07-27
introduce framework for handling "special" APIs

jaypipes:
We needed a way of informing the code generator that certain APIs use
"special" handling of certain fields for resources by using an
"Attributes" `map[string]*string` that contains actual, schema'd fields
for the resource.

An example of this type of special API is SNS API which has a
CreateTopic call that defines a Topic resource. The CreateTopic call has
an "Attributes" parameter that contains a set of key/value pairs for
fields such as "DeliveryPolicy" or "DisplayName".

We want to inform the code generator that these fields are contained in
the Attributes parameter and should be "unpacked" into the CRD
definition as top-level fields on the resource.

This patch introduces this functionality by creating a GeneratorConfig
struct that is represented as a YAML file that contains instructions to
the code generator on how to handle these kinds of
special/inconsistent/idiosyncratic API behaviours.

Issue #109

### DIFF
--- a/pkg/model/crd_test.go
+++ b/pkg/model/crd_test.go
@@ -61,34 +61,53 @@ func TestSNSTopic(t *testing.T) {
 	specFields := crd.SpecFields
 	statusFields := crd.StatusFields
 
+	// The SNS Topic uses an "Attributes" map[string]*string to masquerade
+	// real fields. DeliveryPolicy, Policy, KMSMasterKeyID and DisplayName are
+	// all examples of these masqueraded fields...
 	expSpecFieldCamel := []string{
-		"Attributes",
+		"DeliveryPolicy",
+		"DisplayName",
+		"KMSMasterKeyID",
 		"Name",
+		"Policy",
 		"Tags",
 	}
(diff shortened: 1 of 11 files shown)


pkg/model/crd_test.go
```diff
@@ -61,34 +61,53 @@ func TestSNSTopic(t *testing.T) {
 	specFields := crd.SpecFields
 	statusFields := crd.StatusFields
 
+	// The SNS Topic uses an "Attributes" map[string]*string to masquerade
```
vijtrip2, line 64: Dumb question:
Why do we need to break Map attributes into String Fields?
Why not just keep them as a Map in ko.Spec?

pkg/model/helper.go
```diff
@@ -29,6 +29,9 @@ type Helper struct {
 	// A map of operation type and resource name to
 	// aws-sdk-go/private/model/api.Operation structs
 	opMap *OperationMap
+	// Instructions to the code generator how to handle the API and its
+	// resources
+	generatorConfig *GeneratorConfig
```
vijtrip2, line 34: I am guessing this is pathway to adding more customizations to code-generators. Super Cool Stuff! 💯

PR #116, opened 2020-07-29
generate code for GetAttributes input/output

jaypipes:
Adds generation for Go code that sets the Input shape for GetAttributes
operations or sets CRD Status fields from the Output shape of a
GetAttributes operation.

We add special handling for the primary resource ARN and the AWS Owner
Account ID attribute keys, if found in the Attributes map. For these
special cases, we route the returned Output shape attribute key to the
`Status.ACKResourceMetadata.ARN` and
`Status.ACKResourceMetadata.OwnerAccountID` fields respectively.

Issue #109

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -26,11 +26,13 @@ import (
 )
 
 type CRDOps struct {
-	Create   *awssdkmodel.Operation
-	ReadOne  *awssdkmodel.Operation
-	ReadMany *awssdkmodel.Operation
-	Update   *awssdkmodel.Operation
-	Delete   *awssdkmodel.Operation
+	Create        *awssdkmodel.Operation
+	ReadOne       *awssdkmodel.Operation
+	ReadMany      *awssdkmodel.Operation
+	Update        *awssdkmodel.Operation
+	Delete        *awssdkmodel.Operation
+	GetAttributes *awssdkmodel.Operation
+	SetAttributes *awssdkmodel.Operation
 }
 
 // CRDField represents a single field in the CRD's Spec or Status objects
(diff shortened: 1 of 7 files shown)


templates/pkg/crd_sdk.go.tpl
```diff
@@ -26,9 +26,9 @@ import (
 
 // Hack to avoid import errors during build...
 var (
-    _ = &metav1.Time{}
-    _ = &svcsdk.{{ .SDKAPIInterfaceTypeName}}{}
-    _ = &svcapitypes.{{ .CRD.Names.Camel }}{}
+	_ = &metav1.Time{}
+	_ = &svcsdk.{{ .SDKAPIInterfaceTypeName}}{}
+	_ = &svcapitypes.{{ .CRD.Names.Camel }}{}
```
jaypipes, line 31: This fixed up some spaces that I'd accidentally put in the template instead of TABs.

pkg/model/crd.go
```diff
@@ -419,6 +438,85 @@ func (r *CRD) GoCodeSetInput(
...
+			continue
+		}
+
+		cleanMemberNames := names.New(memberName)
+		cleanMemberName := cleanMemberNames.Camel
+
+		sourceVarPath := sourceVarName
+		field, found := r.SpecFields[memberName]
+		if found {
+			sourceVarName = sourceVarPath + ".Spec." + cleanMemberName
```
vijtrip2, line 496: *sourceVarPath= sourceVarPath + ...

jaypipes: nice catch. actually should be `sourceVarPath = sourceVaName + ".Spec." + cleanMemberName`

PR #117, opened 2020-07-29
fix build errors from attribute-map sets

jaypipes:
When constructing the SNS and SQS controllers, ran into some build
failures, including type conversion issues to/from
string/ackv1alpha.AWSResourceName, unused variable errors when there
were no fields being set in an input or output shape, and import issues
with the aws-sdk-go/aws/errors package.

This patch fixes all the above issues and corrects the build of the SNS
and SQS generated service controllers.

You can test the SNS or SQS service controller build by calling:

```
make build-ack-generate
ACK_GENERATE_CONFIG_PATH=pkg/models/testdata/models/apis/sns/0000-00-00/generator.yaml \
./scripts/build-controller.sh sns
make test
```

### DIFF
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -15,6 +15,8 @@ package errors
 
 import (
 	"fmt"
+
+	"github.com/aws/aws-sdk-go/aws/awserr"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


mhausenblas:
Looks like GitHub swallowed that one comment I had: in order to test it, you need to set `ACK_GENERATE_CONFIG_PATH=pkg/model/testdata/models/apis/sns/0000-00-00/generator.yaml` (note `pkg/model` and NOT `pkg/models` ;)

PR #155, opened 2020-08-07
local functional testing

nithu0115:
Issue #, if available: N/A

Description of changes:

- [x] Added sts-assume-role.sh to create temporary credentials for functional testing
- [x] Updated kind-build-test.sh script to include `-r` flag which will users to pass `--role-arn` for functional testing where we make `aws sts assume-role` API call to fetch temp creds and set as environment variable to deployment. This way, we can create AWS resources by leverage the same local KinD cluster. 
- [x] Updated kind-build-test.sh script to fail fast if service name to build parameter is not passed 
- [x] Added makefile config to invoke all these scripts instead of navigating to script dir. Now you can just run `make build-controller SERVICE_TO_BUILD=ecr` commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+SHELL := /bin/bash # Use bash syntax
 GO111MODULE=on
 
 PKGS=$(sort $(dir $(wildcard pkg/*/*/)))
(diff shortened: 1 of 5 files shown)


scripts/lib/sts-assume-role.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: let's place this code into the existing `scripts/lib/aws.sh` file.

PR #159, opened 2020-08-10
Build controllers container images using one Dockerfile

A-Hilaly:
Issue #, if available: N/A

Description of changes:
- Add script to build multiple controllers container images
- Remove `ack-generate` code that generate Dockerfiles
- Add shell utilities arround docker CLI/daemon

Testing:
```bash
SERVICES=ecr
# generate service controller: ./scripts/build-controller.sh $SERVICES
./scripts/build-controllers-image.sh $SERVICES
```

output
```
Sending build context to Docker daemon  181.1MB
Step 1/19 : FROM golang:1.14.1 as builder
 ---> b022a9b82d88
Step 2/19 : ARG service_alias
 ---> Using cache
 ---> 7c25bf56883e
Step 3/19 : ARG work_dir=/github.com/aws/aws-controllers-k8s
 ---> Running in a1b6d377ab2c
Removing intermediate container a1b6d377ab2c
 ---> fb616e458fbe
Step 4/19 : WORKDIR $work_dir
 ---> Running in 4c8726cc0137
Removing intermediate container 4c8726cc0137
 ---> 023b05a8d8f3
Step 5/19 : ENV GOPROXY=direct
 ---> Running in 6a1186f0e5db
Removing intermediate container 6a1186f0e5db
 ---> 9b40ebce95c0
Step 6/19 : ENV GO111MODULE=on
 ---> Running in 7df85a1b08c0
Removing intermediate container 7df85a1b08c0
 ---> 206798c0b58e
Step 7/19 : ENV GOARCH=amd64
 ---> Running in efdaba646ee0
Removing intermediate container efdaba646ee0
 ---> 7397e5da20a1
Step 8/19 : ENV GOOS=linux
 ---> Running in e5f39f31cd8e
Removing intermediate container e5f39f31cd8e
 ---> 6e559a262d2f
Step 9/19 : ENV CGO_ENABLED=0
 ---> Running in 55ede6b65cd7
Removing intermediate container 55ede6b65cd7
 ---> ebc6eeb39132
Step 10/19 : COPY go.mod go.mod
 ---> 63b4c3137758
Step 11/19 : COPY go.sum go.sum
 ---> 240460748c7a
Step 12/19 : RUN  go mod download
 ---> Running in b9ea611ba8f3
Removing intermediate container b9ea611ba8f3
 ---> 1c1cbc942ea4
Step 13/19 : COPY . $work_dir/
 ---> 075c21e1ff82
Step 14/19 : RUN  go build -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
 ---> Running in cc0e201b04a2
Removing intermediate container cc0e201b04a2
 ---> 7bfa618e376e
Step 15/19 : FROM amazonlinux:2
 ---> fa0a6a710ca7
Step 16/19 : ARG work_dir=/github.com/aws/aws-controllers-k8s
 ---> Using cache
 ---> 25ca0bbe041d
Step 17/19 : WORKDIR /
 ---> Using cache
 ---> 137038d9a0ae
Step 18/19 : COPY --from=builder $work_dir/bin/controller /bin/.
 ---> 7f804f1334d4
Step 19/19 : ENTRYPOINT ["/bin/controller"]
 ---> Running in 5b1d864a6fd0
Removing intermediate container 5b1d864a6fd0
 ---> a14f7d17a531
[Warning] One or more build-args [GOPROXY] were not consumed
Successfully built a14f7d17a531
Successfully tagged ack/ecr:latest
```


scripts/build-controllers-image.sh
```diff
@@ -0,0 +1,50 @@
...
+done
+
+if [[ ${#SERVICES[@]} -eq 0 ]]; then
+    echo "ERROR: Specify at least one service to build an container image for" 1>&2
+    echo "$USAGE"
+    exit 1
+fi
+
+for SERVICE in ${SERVICES[@]}; do
+    IMAGE_TAG=ack/$SERVICE
```
jaypipes, line 40: IMAGE_TAG should have a tag suffix (otherwise it defaults to ":latest" which isn't a best practice. Can you default the tag to the git tag or dirty, like the kind-build-test.sh script does:

```
VERSION=$(git describe --tags --always --dirty || echo "unknown")
DEFAULT_AWS_SERVICE_DOCKER_IMG="ack-${AWS_SERVICE}-controller:${VERSION}"
```

scripts/lib/docker.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+
+# ensure_docker [<docker-version>]
```
jaypipes, line 4: meh, there is no handling of an argument for the version?

scripts/lib/docker.sh
```diff
@@ -0,0 +1,26 @@
...
+# 
+# Ensure that docker client is installed and daemon is running.
+ensure_docker () {
+    ensure_docker_client && ensure_docker_daemon_running
+}
+
+ensure_docker_client () {
```
jaypipes, line 11: generally, I've been using the functions prefixed with `ensure_` to actually install the required tools. The problem with Docker, though, is that it's not possible to install and configure it without a) superuser privleges, b) reloading the shell and usergroups and c) doing it in a sane multi-OS way.

Which is why I've typically just put this into the top of the scripts that depend on Docker:

```
source $SCRIPTS/lib/common.sh
...
check_is_installed docker
```

which does the same as what you're doing above :)

A-Hilaly: Haaa i see! thanks for explaining! Let's keep only `check_is_installed` then 👍 . i'll remove the docker.sh lib file

scripts/lib/docker.sh
A-Hilaly: Right. it will be complex to implement for both daemon/cli let's keep it simple for now

scripts/lib/docker.sh
A-Hilaly: @jaypipes i removed the `ensure_` commands. Do you think it will be useful to have `check_docker_daemon_running` or shall we keep it like this ?

Issue #106, opened 2020-07-24
Secrets Integration to Protect CRD Fields


alina-kim:
The AWS RDS API, among others, requires sensitive fields to be displayed in plain text. To protect such fields, we want to:

- Allow users to create Kubernetes Secrets and store their sensitive data in them
- Receive Secret references in place of target fields and pass the Secret data to the desired AWS API

Link to the [Secrets Integration design proposal](https://github.com/alina-kim/aws-controllers-k8s/blob/ack-secrets/docs/design/proposals/secrets/secrets.md)

RedbackThomson:
Added as part of https://github.com/aws-controllers-k8s/code-generator/pull/21

PR #161, opened 2020-08-10
plumb reconciler into resource manager

jaypipes:
We need a way of allowing the resource manager to get Kubernetes object
information for things like Kubernetes Secrets. However, I deliberately
did not want to pass the Kubernetes client objects into the resource
managers (since their job is to handle the SDK linkage, not to query
Kubernetes API).

However, the resource manager will need at least the ability to grab a
value for a SecretReference, and in order to facilitate this, I changed
the `AWSResourceManagerFactory.ManagerFor` interface method to pass in
an object that implements the `AWSResourceReconciler` interface. I then
added a stub `AWSResourceReconciler.SecretValueFromReference()`
interface method (for Alina to implement) into the runtime's Reconciler
concrete struct. The resource managers now have an `rr` field that
points to the Reconciler for the runtime and will allow the resource
managers to call utility methods like `SecretValueFromReference` and not
have to handle calls to the Kubernetes API directly themselves.

Related: #106

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrlrt "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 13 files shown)


pkg/runtime/reconciler.go
```diff
@@ -67,6 +68,15 @@ func (r *reconciler) BindControllerManager(mgr ctrlrt.Manager) error {
 	).Complete(r)
 }
 
+// SecretValueFromReference fetches the value of a Secret given a
+// SecretReference
+func (r *reconciler) SecretValueFromReference(
+	ref *corev1.SecretReference,
+) (string, error) {
+	// TODO(alina-kim): Implement this method :)
+	return "", nil
```
A-Hilaly, line 77: nit: maybe we should return a `"", fmt.Errorf("not implemented")` for now ?

jaypipes: even better. ++

jaypipes: done.

Issue #127, opened 2020-07-31
Panic during api creation


vijtrip2:
**Describe the bug**
Before create call, sdkFind call is causing panic by de-referencing nil objects. 

**Steps to reproduce**
Start the ack controller for APIGatewayV2 and ```kubectl create -f httpapisample.yaml```

*httpapisample.yaml*

```
apiVersion: apigatewayv2.services.k8s.aws/v1alpha1
kind: API
metadata:
  name: httpapi-sample
spec:
  name: httpapi-sample
```
-----------

**Bug#1**
*The aws-request generation is not setting any fields in the input.*
The method below should be creating request by setting an apiId or relevant fields for a successful get.
```
func (rm *resourceManager) newDescribeRequestPayload(
	r *resource,
) (*svcsdk.GetApiInput, error) {
	res := &svcsdk.GetApiInput{}

	return res, nil
}
```
------
**Bug#2**
nil checks should be present in sdkFind() method before dereferencing the response from aws-sdk
resp.CreatedDate is nil in below case and causes the panic
```
// Merge in the information we read from the API call above to the copy of
	// the original Kubernetes object we passed to the function
	ko := r.ko.DeepCopy()

	ko.Status.APIEndpoint = resp.ApiEndpoint
	ko.Status.APIID = resp.ApiId
        ko.Status.CreatedDate = &metav1.Time{*resp.CreatedDate}
```
-------------
**Stack Trace**
```
E0731 13:59:48.699223   75157 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 368 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x22092c0, 0x32be010)
        /Users/vijat/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /Users/vijat/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:48 +0x82
panic(0x22092c0, 0x32be010)
        /usr/local/go/src/runtime/panic.go:967 +0x166
[...]

PR #163, opened 2020-08-11
properly handle NotFoundException in reads

jaypipes:
The templated sdk.go file was looking for the string "NotFoundException"
to identify whether a call to an AWS API was returning a 404 NotFound.
Unfortunately this isn't correct for most APIs, including Elasticache.
The exception codes in Elasticache look like "CacheClusterNotFound"
instead of the generic "NotFoundException". Trying to match on the
string "NotFound" doesn't work either, since calls to, for example, the
DescribeCacheClusters API call legitimately return 400 with a code of
"ServiceLinkedRoleNotFound" which is actually a permissions problem.

This patch adds a `CRD.NotFoundException()` method which looks through
the read-like API calls for Error codes and looks for 404 HTTP status
codes and writes the matched exception code string into the sdk.go
template instead of the hard-coded 'NotFoundException'.

In the process, I also fixed up a shadowing problem where some of the
sdk.go template code referred to the wrong error variable (err instead
of respErr).

Finally, a few cleanups to `scripts/kind-build-test.sh` allow properly
testing with AWS_REGION set and in environments where the default
output encoding for the `aws` CLI tool isn't `json`.

Issue #127 
Issue #157

### DIFF
--- a/scripts/build-controller-image.sh
+++ b/scripts/build-controller-image.sh
@@ -63,7 +63,7 @@ docker build \
   --quiet=${QUIET} \
   -t ${AWS_SERVICE_DOCKER_IMG} \
   -f ${DOCKERFILE} \
-  --build-arg service_alias=${SERVICE} \
+  --build-arg service_alias=${AWS_SERVICE} \
   ${BUILD_CONTEXT}
 
 if [ $? -ne 0 ]; then
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -15,7 +15,10 @@ source "$SCRIPTS_DIR/lib/k8s.sh"
 
 OPTIND=1
 CLUSTER_NAME_BASE="test"
+AWS_ACCOUNT_ID=${AWS_ACCOUNT_ID:-"123456789012"}
+AWS_REGION=${AWS_REGION_ID:-"us-west-2"}
 AWS_ROLE_ARN=""
+ACK_ENABLE_DEVELOPMENT_LOGGING="true"
 DELETE_CLUSTER_ARGS=""
 K8S_VERSION="1.16"
 OVERRIDE_PATH=0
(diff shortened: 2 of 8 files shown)


scripts/kind-build-test.sh
```diff
@@ -15,7 +15,10 @@ source "$SCRIPTS_DIR/lib/k8s.sh"
 
 OPTIND=1
 CLUSTER_NAME_BASE="test"
+AWS_ACCOUNT_ID=${AWS_ACCOUNT_ID:-"123456789012"}
```
nithu0115, line 18: I feel like we should parse the ROLE_ARN passed by user to get ACCOUNT_ID instead of passing a random string to avoid confusion.

scripts/kind-build-test.sh
```diff
@@ -132,6 +135,11 @@ export KUBECONFIG="${TMP_DIR}/kubeconfig"
 trap "exit_and_fail" INT TERM ERR
 trap "clean_up" EXIT
 
+export AWS_ACCOUNT_ID
+export AWS_REGION
+export AWS_ROLE_ARN
+export ACK_ENABLE_DEVELOPMENT_LOGGING
+
```
nithu0115, line 142: is this export statements required ?

scripts/lib/aws.sh
```diff
@@ -60,7 +60,8 @@ generate_aws_temp_creds() {
   JSON=$(aws sts assume-role \
            --role-arn "$AWS_ROLE_ARN"  \
            --role-session-name tmp-role-"$__uuid" \
-           --duration-seconds 900 || exit 1)
+           --duration-seconds 900 \
+           --output json || exit 1)
```
nithu0115, line 64: neat 👍

scripts/build-controller-image.sh
```diff
@@ -63,7 +63,7 @@ docker build \
   --quiet=${QUIET} \
   -t ${AWS_SERVICE_DOCKER_IMG} \
   -f ${DOCKERFILE} \
-  --build-arg service_alias=${SERVICE} \
+  --build-arg service_alias=${AWS_SERVICE} \
```
A-Hilaly, line 66: Nice catch 👍

Issue #180, opened 2020-08-14
S3 service controller build failures


jaypipes:
```
jaypipes@thelio:~/go/src/github.com/aws/aws-controllers-k8s$ make build-ack-generate 
go build -tags codegen -o bin/ack-generate cmd/ack-generate/main.go
jaypipes@thelio:~/go/src/github.com/aws/aws-controllers-k8s$ ./scripts/build-controller.sh s3
Building Kubernetes API objects for s3
Generating deepcopy code for s3
1297:20: missing ',' in argument list
1392:20: missing ',' in argument list
3300:20: missing ',' in argument list
3305:20: missing ',' in argument list
Error: not all generators ran successfully
run `controller-gen object:headerFile=/home/jaypipes/go/src/github.com/aws/aws-controllers-k8s/scripts/../templates/boilerplate.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=/home/jaypipes/go/src/github.com/aws/aws-controllers-k8s/scripts/../templates/boilerplate.txt paths=./... -h` for usage
Generating custom resource definitions for s3
Building service controller for s3
Generating RBAC manifests for s3
```

There are issues generating deepcopy. Need to investigate...

jaypipes:
Part one of investigation...

Line 1297 of `services/s3/apis/v1alpha1/zz_deepcopy.generated.go` looks like this:

```go
// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
func (in *InputSerialization) DeepCopyInto(out *InputSerialization) {
*out = *in
if in.CSV != nil {
in, out := &in.CSV, &out.CSV
*out = new(CSVInput)
(*in).DeepCopyInto(*out)
}
if in.CompressionType != nil {
in, out := &in.CompressionType, &out.CompressionType
*out = new(string)
**out = **in
}
if in.JSON != nil {
in, out := &in.JSON, &out.JSON
*out = new(JSONInput)
(*in).DeepCopyInto(*out)
}
if in.Parquet != nil {
in, out := &in.Parquet, &out.Parquet
line 1297 >>>>>> *out = new(invalid type)
**out = **in
}
}

```

So I looked up `ParquetInput` in `services/s3/apis/v1alpha1/types.go` and found no type definition. So, I looked at the model file and found this:

```json
    "ParquetInput":{
      "type":"structure",
      "members":{
      }
    },
```

[...]

PR #185, opened 2020-08-16
S3 service controller enhancements and fixes

jaypipes:
The S3 CreateBucket API Operation has an Input shape with a field called
"Bucket". This is actually the *name* of the Bucket to be created.
Unfortunately, the ListBuckets API Operation's Output shape (from which
we build the SDK linkage to find a particular object) calls this same
field "Name" (quite appropriately I might add, since that's what the
field represents... the bucket's name).

In order to handle mismatched field names like this, I have added a
`RenamesConfig` struct to the `GeneratorConfig` that holds instructions
on how to rename certain fields in the input and output shapes.

The S3 `generator.yaml` config has been modified with the following
rename instructions:

```yaml
resources:
  Bucket:
    renames:
      operations:
        CreateBucket:
          input_fields:
            Bucket: Name
```

which effectively renames the `BucketSpec.Bucket` field to
`BucketSpec.Name`.

Fixes the last build failure identified in #180.

Issue #180

Adds `ListOperationConfig` struct to the `GeneratorConfig`. This struct
allows us to instruct the code generator that a particular API's List
operation returns all resources of a particular type and that it will
need to add Go code to the `sdkFind()` method of the resource manager to
filter out results from the returned set of objects that do not match
the values of certain fields.

It's easier to talk about this with an example. Let's take the S3 Bucket
API. There is a ListBuckets API call. The Input shape for ListBuckets is
always empty. There is no ability to ask ListBuckets to only return
results for, say, a particular named Bucket. Instead, in order to use
the ListBuckets API call to find a *particular* named Bucket, you need
to ListBuckets and then loop over all the results, looking for a result
with a Name field matching the Bucket.Spec.Name that you are looking
for.

This patch changed the code generator to add Go code to the output shape
processing for the ListBuckets API that looks like this:

```go
if len(resp.Buckets) == 0 {
    return nil, ackerr.NotFound
}
found := false
for _, elem := range resp.Buckets {
    if elem.Name != nil {
        if ko.Spec.Name != nil {
            if *elem.Name != *ko.Spec.Name {
                continue
            }
        }
        r.ko.Spec.Name = elem.Name
    }
    found = true
}
if !found {
    return nil, ackerr.NotFound
}
```

The `generator.yaml` for S3 looks like this now:

```yaml
ignore:
  resource_names:
    - Object
    - MultipartUpload
  shape_names:
    # These shapes are structs with no members...
    - SSES3
resources:
  Bucket:
    renames:
      operations:
        CreateBucket:
          input_fields:
            Bucket: Name
        DeleteBucket:
          input_fields:
            Bucket: Name
    list_operation:
      match_fields:
        - Name
```

The `resources.Bucket.list_operation` element contains the instructions
telling the code generator to output special code to filter out results
that don't match the searched-for Bucket's Name.

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -1215,8 +1215,11 @@ func (r *CRD) GoCodeSetOutput(
 // Note: "ko" is the target variable and represents the thing we'll be
 //		 setting fields on
 //
-//  if len(resp.CacheClusters) == 1 {
-//      elem := resp.CacheClusters[0]
+//  if len(resp.CacheClusters) == 0 {
+//      return nil, ackerr.NotFound
+//  }
+//  found := false
+//  for _, elem := range resp.CacheClusters {
 //      if elem.ARN != nil {
 //          if ko.Status.ACKResourceMetadata == nil {
 //              ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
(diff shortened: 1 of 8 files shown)


pkg/model/crd.go
```diff
@@ -1752,7 +1871,10 @@ func (h *Helper) GetCRDs() ([]*CRD, error) {
 			if h.IsIgnoredShape(memberShapeRef.Shape.ShapeName) {
 				continue
 			}
-			memberNames := names.New(memberName)
+			renamedName, _ := crd.InputFieldRename(
```
nithu0115, line 1874: Question: under what conditions does InputFieldRename return false ?

jaypipes: it actually returns (string, bool). The string part of the return is the renamed field name, or if there was no renaming done, the original field name.

PR #198, opened 2020-08-18
Updated testing docs

nithu0115:
Issue #, if available: N/A

Description of changes: Updated testing docs.
 
### DIFF
--- a/docs/contents/dev-docs/testing.md
+++ b/docs/contents/dev-docs/testing.md
@@ -10,7 +10,7 @@ root directory of your [checked-out source repository](../setup/).
     the step that builds the container image for the target ACK service
     controller can 40 or more minutes. This is because the container image
     contains a lot of dependencies. Once you successfully build the target
-    image this base image layer is cached locally and the build takes a much 
+    image this base image layer is cached locally, and the build takes a much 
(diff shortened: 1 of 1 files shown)


docs/contents/dev-docs/testing.md
```diff
@@ -57,13 +57,17 @@ If you already set this environment variable above during build stage, ignore th
...
-for the respective AWS Service:
+To generate temporary credentials for functional testing, pass in a role ARN. 
 
 ```
-AWS_ROLE_ARN=arn:aws:iam::12345678980:role/Admin-k8
+AWS_ROLE_ARN=arn:aws:iam::<account-id>:role/<role-name>
 ```
 
+!!! warning
+    The role ARN passed should exist in your AWS account, and has necessary permissions to create/describe/delete the resource.
```
mhausenblas, line 67: The role *should* or *must* exist? What happens if it doesn't? Can you also add an example here, as in: if you want to be test read-only from an S3 bucket then use the ARN ...

nithu0115: Good point. Role must exist and have necessary permissions. I will go ahead and add some examples roles. We cannot have a read only role unfortunately as we are create a resource and describe it to make sure it is successfully created as part of E2E tests.

Issue #209, opened 2020-08-18
apigatewayv2: invalid memory address or nil pointer dereference


vijtrip2:
**Describe the bug**
Probable nil check required during reconciliation.

**Sample Config**
```
apiVersion: apigatewayv2.services.k8s.aws/v1alpha1
kind: API
metadata:
  name: httpapi-sample-a
spec:
  name: httpapi-sample-a
```

**StackTrace**
```
E0818 10:06:04.442784     226 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 344 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x2212d60, 0x32d8170)
	/Users/vijat/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:74 +0xa3
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/Users/vijat/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/runtime/runtime.go:48 +0x82
panic(0x2212d60, 0x32d8170)
	/usr/local/go/src/runtime/panic.go:967 +0x166
github.com/aws/aws-controllers-k8s/services/apigatewayv2/pkg/resource/api.(*resourceDescriptor).Equal(0x33201c0, 0x27623e0, 0xc00059a6a0, 0x27623e0, 0x0, 0x27623e0)
	/Users/vijat/Documents/gocode/src/github.com/vijtrip2/aws-controllers-k8s/services/apigatewayv2/pkg/resource/api/descriptor.go:78 +0x9b
github.com/aws/aws-controllers-k8s/pkg/runtime.(*reconciler).sync(0xc00030a580, 0x2758f60, 0xc0000b4008, 0x275fc20, 0xc00038fb30, 0x27623e0, 0xc00059a6a0, 0x275fc20, 0xc00038fb30)
	/Users/vijat/Documents/gocode/src/github.com/vijtrip2/aws-controllers-k8s/pkg/runtime/reconciler.go:157 +0x6a3
github.com/aws/aws-controllers-k8s/pkg/runtime.(*reconciler).reconcile(0xc00030a580, 0xc00070e659, 0x7, 0xc00070e640, 0x10, 0xc01a489f58, 0xb31dd0e5acf7)
[...]

Issue #162, opened 2020-08-11
Input YAML validation


nmvk:
**Is your feature request related to a problem?**
I was able to create request with invalid YAML file (Though syntactically correct)

**Format of invalid YAML**
```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: CacheParameterGroup
metadata:
  name: test-pg
spec:
  properties:
    cacheParameterGroupFamily: redis5.0 
    cacheParameterGroupName: test-pg
    description: test
```

**Format of correct YAML**
```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: CacheParameterGroup
metadata:
  name: test-pg
spec:
    cacheParameterGroupFamily: redis5.0 
    cacheParameterGroupName: test-pg
    description: test
```

You can see that wrapping the `CreateCacheParameter` with `properties` ends up returning `nil` when generated code would call `r.ko.Spec.CacheParameterGroupName`.

**Describe the solution you'd like**
Throw user error, so they can fix the YAML file.

**Describe alternatives you've considered**
N/A

jaypipes:
Yes, we need to add validating webhook generation to ACK.

PR #211, opened 2020-08-18
Handing the validation error during sdkReadOne operation.

vijtrip2:
Issue #209 , if available:

Description of changes:
* During sdkFind operation for a new CR, the GetAPIInput was throwing validation error due to absence of APIId/HTTP Api. The validation failure in GetApiInput is a symptom of resource not present.
* I changed the template to treat Validation failure same way we treat NotFoundException.

P.S. After this change I was able to create an HTTP Api.

### DIFF
--- a/templates/pkg/crd_sdk.go.tpl
+++ b/templates/pkg/crd_sdk.go.tpl
@@ -14,6 +14,7 @@ import (
 	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
 	ackerr "github.com/aws/aws-controllers-k8s/pkg/errors"
 	"github.com/aws/aws-sdk-go/aws"
+	awsreq "github.com/aws/aws-sdk-go/aws/request"
 	svcsdk "github.com/aws/aws-sdk-go/service/{{ .ServiceAlias }}"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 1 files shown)


templates/pkg/crd_sdk.go.tpl
```diff
@@ -46,7 +47,13 @@ func (rm *resourceManager) sdkFind(
 		if awsErr, ok := ackerr.AWSError(respErr); ok && awsErr.Code() == "{{ ResourceExceptionCode .CRD 404 }}" {
 			return nil, ackerr.NotFound
 		}
-		return nil, err
+
+		// If there are validation errors in constructed request
+		// treat it as NotFound
```
jaypipes, line 52: A few things...

First, an invalid parameter isn't a not found :) We should be producing valid Input shapes, so if we're not, let's fix the source of that bug.

Second, we need to handle generic validation of CR input using validating admission webhooks (see #162).

vijtrip2: I agree to above points and treating ValidationException as NotFound will hide validation-failure before calling Create Api.

I will think of better solution but this was a special scenario. APIGatewayV2 api is a bit different from s3 and ecr. ValidationException occurs even with a legit CR input. See points below.

* To create valid input shape(GetApiInput) in above case, there has to be an existing API(Populated KO status). Which will not happen because we have ReadOne before Create operation.
* S3 and ECR APIs use list operations and filter on a certain field while APIGatewayV2 uses Get(w/ api-id) rather than List.


@jaypipes , I will think of a better solution and  run it by you first before implementing.

Issue #208, opened 2020-08-18
ECR service controller


tabern:
## New ACK service controller

Support for Amazon ECR.

### List of API resources

List the API resources in order of importance to you:

* [Repository](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_CreateRepository.html)

mikestef9:
Closing as this service controller has graduated to GA. Separate issues can be opened to discuss specific follow on topics on the controller.

PR #214, opened 2020-08-19
ensure ECR DescribeRepositories properly generated

jaypipes:
There were two issues with the generated ECR service controller code.
The first issue had to do with the generated code not properly
understanding the NotFound exception code for repositories. The solution
to this was adding something to the generatorConfig for ECR to instruct
the generator which exception code to identity NotFound errors.

The second issue had to do with another generatorConfig instruction:
tell the code generator to output Go code to filter out Repository
results in the `DescribeRepositoriesOutput.Repositories` struct field
that did not have a match `RepositoryName` value as the CR's
`Spec.RepositoryName` field.

Adding the `services/ecr/generator.yaml` file and configuring it like so
fixed the above two problems:

```yaml
resources:
  Repository:
    exceptions:
      codes:
        404: RepositoryNotFoundException
    list_operation:
      match_fields:
        - RepositoryName
```

Finally, I updated both the unit test for ECR repository code generation
as well as modified the ECR smoke test to include creation of 3 separate
repositories and then delete those 3 repositories. Doing this triggers
the issue with DescribeRepositories returning multiple results and the
generated Go code not properly filtering out results by matching
RepositoryName.

Issue #208


pkg/model/crd_test.go
```diff
@@ -889,6 +889,60 @@ func TestECRRepository(t *testing.T) {
...
+			ko.Status.ACKResourceMetadata.ARN = &tmpARN
+		}
+		if elem.RepositoryName != nil {
+			if ko.Spec.RepositoryName != nil {
+				if *elem.RepositoryName != *ko.Spec.RepositoryName {
+					continue
+				}
+			}
+			ko.Spec.RepositoryName = elem.RepositoryName
+		}
```
jaypipes, line 934: Note to reviewers. This is the code that is critical to solving the problem @mhausenblas was seeing. He was seeing this (while others were not) because there were already other ECR repositories in the account Michael was using to test with. Those other repositories were being returned in the call to DescribeRepositories but not being filtered out of the list, which resulted in calls to Create not working (because NotFound wasn't being returned)

mhausenblas (approved):

A-Hilaly (approved):

Issue #215, opened 2020-08-19
Smoke test for ElastiCache CacheSubnetGroup


kumargauravsharma:
**Is your feature request related to a problem?**
Test to verify create, delete CacheSubnetGroup using ACK for ElastiCache.

**Describe the solution you'd like**
Automated test (similar to smoke tests available at: https://github.com/aws/aws-controllers-k8s/blob/main/test/e2e/) which: 
* Creates CacheSubnetGroup using ACK for ElastiCache
* Verifies create using aws cli
* Deletes CacheSubnetGroup using ACK for ElastiCache
* Verifies delete using aws cli

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

PR #216, opened 2020-08-19
generator config and smoke test for ElastiCache CacheSubnetGroup #215

kumargauravsharma:
Issue #215 

Description of changes: generator config and smoke test for ElastiCache CacheSubnetGroup 

### DIFF
--- a/test/e2e/elasticache/cache_subnet_group/smoke.sh
+++ b/test/e2e/elasticache/cache_subnet_group/smoke.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-ROOT_DIR="$THIS_DIR/../../.."
+ROOT_DIR="$THIS_DIR/../../../.."
 SCRIPTS_DIR="$ROOT_DIR/scripts"
 
 source "$SCRIPTS_DIR/lib/common.sh"
@@ -30,13 +30,12 @@ fi
 
 
 describe_subnet_json() {
-    jq_expr='.CacheSubnetGroups[] | select(.CacheSubnetGroupName | contains($CacheSubnetGroupName))'
-    aws elasticache describe-cache-subnet-groups --output json | jq -e --arg CacheSubnetGroupName "$aws_resource_name" "$jq_expr"
+    aws elasticache describe-cache-subnet-groups --cache-subnet-group-name "$aws_resource_name"  --output json >/dev/null 2>&1
 }
 
 # PRE-CHECKS
 describe_subnet_json
-if [ $? -ne 4 ]; then
+if [[ $? -ne 255 && $? -ne 254 ]]; then
     echo "FAIL: expected $aws_resource_name to not exist in ${service_name}. Did previous test run cleanup?"
     exit 1
 fi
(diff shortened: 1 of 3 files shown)


test/e2e/elasticache/smoke_cache_subnet_group.sh
```diff
@@ -0,0 +1,83 @@
...
+    echo "no subnet id found to run the test. Ensure that subnet id is available to run the test."
+    exit 1
+else
+    echo "using subnet id: ${aws_subnet_id} for test."
+fi
+
+
+describe_subnet_json() {
+    jq_expr='.CacheSubnetGroups[] | select(.CacheSubnetGroupName | contains($CacheSubnetGroupName))'
+    aws elasticache describe-cache-subnet-groups --output json | jq -e --arg CacheSubnetGroupName "$aws_resource_name" "$jq_expr"
```
jaypipes, line 34: Instead of using the jq query here, since the `DescribeCacheSubnetGroups` API call has a filter for the cache subnet group name, you could more easily and efficiently use this:

```
aws elasticache describe-cache-subnet-groups --output json  --cache-subnet-group-name "$aws_resource_name"
```

which will do the filtering on the server side and matches the technique we use in the ECR smoke test here:

https://github.com/aws/aws-controllers-k8s/blob/4a4aae282ba8ebecf25fc8214bd4d58e82de4907/test/e2e/ecr/smoke.sh#L29-L30

test/e2e/elasticache/smoke_cache_subnet_group.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: Go ahead and rename this file to `test/e2e/elasticache/cache_subnet_group/smoke.sh` and we'll put all the cache subnet group tests in a separate directory. The e2e bash test runner handles this just fine. For the ECR and S3 smoke tests, it's fine to have just a single `smoke.test` file in the main directory since there's only one resource in those APIs (Repository and Bucket, respectively).

We will eventually create a `test/e2e/elasticache/smoke.sh` test file that will test related/dependent object creation, but for these single-resource tests, let's place those in separate directories associated with the specific resource. Sound good?

test/e2e/elasticache/smoke_cache_subnet_group.sh
kumargauravsharma: Thanks for your input.
I have updated the changes.

test/e2e/elasticache/smoke_cache_subnet_group.sh
kumargauravsharma: Updated the file per the comment.

test/e2e/elasticache/cache_subnet_group/smoke.sh
```diff
@@ -0,0 +1,82 @@
...
+debug_msg "executing test: $service_name/$test_name"
+
+aws_resource_name="ack-test-smoke-${service_name}-subnet-gp"
+k8s_resource_name="cachesubnetgroups/$aws_resource_name"
+
+# pre-req: a subnet id to create cache subnet group
+aws_subnet_id="$(aws ec2 describe-subnets --output json | jq -e '.Subnets[0] | .SubnetId')"
+
+if [ $? -eq 4 ] || [ -z $aws_subnet_id ]; then
+    echo "no subnet id found to run the test. Ensure that subnet id is available to run the test."
```
jaypipes, line 25: In a future commit, feel free to make the above failure message start with "FAIL: " to match the same style used below on line 39 and 44.

Also, feel free to move this check down to the section that starts on line 36 below since this is a "PRE-CHECK" in the same way as the other pre-checks.

PR #232, opened 2020-08-21
Handling exception from invalid ReadOneInput

vijtrip2:
Issue #209 , if available:

Description of changes:
`During reconciliation, Ignore ReadOne call before Create call only if ReadOneInput has required parameters from CreateOutput.`

* Created a type ValidationHelper to handle validation logic. This class can include more methods in future.
* Created a validation method which  validates that cause of ValidationException is from all the ignorable fields.
* Used this validation method after creating ReadOneInput and passing all the required field from ko.Status as ignorable fields
* These fields are okay to be ignored because these fields will not available until first create call.

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -371,6 +371,34 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
 	return "UNKNOWN"
 }
 
+/**
+This method return the required input fields names for ReadOne operation which are present in ko.Status .
+During initial creation of aws resource, ReadOne call should be ignored if validation error
+occurs from absence of these fields.
+
+Example output: requiredStatusFields := []string{"GetApiInput.ApiId"}
+*/
+func (r *CRD) RequiredStatusFieldsForReadOneInput(indentLevel int) string {
+	op := r.Ops.ReadOne
+	inputShape := op.InputRef.Shape
+	if inputShape == nil || len(inputShape.Required) == 0 {
+		return ""
+	}
+	requiredFields := inputShape.Required
+	requiredStatusFieldsStr := ""
+	for _, requiredField := range requiredFields {
+		_, found := r.StatusFields[requiredField]
+		if found {
+			requiredStatusFieldsStr += fmt.Sprintf("\"%s.%s\",", inputShape.ShapeName, requiredField)
+		}
+	}
+	if last := len(requiredStatusFieldsStr) - 1; last >= 0 && requiredStatusFieldsStr[last] == ',' {
+		requiredStatusFieldsStr = requiredStatusFieldsStr[:last]
+	}
+	indent := strings.Repeat("\t", indentLevel)
+	return fmt.Sprintf("%srequiredStatusFields := []string{%s}", indent, requiredStatusFieldsStr)
+}
+
 // GoCodeSetInput returns the Go code that sets an input shape's member fields
 // from a CRD's fields.
 //
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -372,7 +372,7 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
 }
 
 /**
-This method return the required input fields names for ReadOne operation which are present in ko.Status .
+This method returns the required fields names for ReadOneInput which are present in ko.Status .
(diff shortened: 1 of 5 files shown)


templates/pkg/crd_sdk.go.tpl
```diff
@@ -40,13 +43,34 @@ func (rm *resourceManager) sdkFind(
 	if err != nil {
 		return nil, err
 	}
+	//If the input implements validate method, invoke Validate() before making ReadOne call
+	inputType := reflect.TypeOf(input)
+	_, validateable := inputType.MethodByName("Validate")
+	if validateable {
+		err := input.Validate()
```
A-Hilaly, line 50: Maybe we can just use interface type assertions here instead of using the `reflect` package ?

pkg/model/validation_helper.go
```diff
@@ -0,0 +1,66 @@
...
+			return false
+		} else {
+			//Make sure all the requiredStatusFieldNames are present in the validationError
+			allIgnorableFieldsPresent := true
+			for _, ignorableFieldName := range ignorableFieldNames {
+				allIgnorableFieldsPresent = allIgnorableFieldsPresent && strings.Contains(strings.ToLower(errInvalidParams.Error()), strings.ToLower(ignorableFieldName))
+			}
+			if !allIgnorableFieldsPresent {
+				return false
+			}
```
A-Hilaly, line 62: nit: we can also just return false when we see the first non ignorable field:
```golang
for _, ignorableFieldName := range ignorableFieldNames {
    if !strings.Contains(strings.ToLower(errInvalidParams.Error()), strings.ToLower(ignorableFieldName)) {
        return false
    }
}
```

templates/pkg/crd_sdk.go.tpl
A-Hilaly: I think it is not possible to directly check (on runtime) if a type implement an interface. But with interface conversation we can achieve that with something like this:

```golang
package main

import "fmt"

type Validatable interface {
	Valid() error
}

type Input struct{}

func (i *Input) Valid() error { return nil }

func isValidatable(t interface{}) bool {
	_, ok := t.(Validatable)
	return ok
}

func main() {
	i := &Input{}
	fmt.Println(isValidatable(i))
}

```

A-Hilaly:
Good work @vijtrip2 ! I Left few questions/suggestions above :)

pkg/model/crd.go
```diff
@@ -371,6 +371,34 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
...
+		if found {
+			requiredStatusFieldsStr += fmt.Sprintf("\"%s.%s\",", inputShape.ShapeName, requiredField)
+		}
+	}
+	if last := len(requiredStatusFieldsStr) - 1; last >= 0 && requiredStatusFieldsStr[last] == ',' {
+		requiredStatusFieldsStr = requiredStatusFieldsStr[:last]
+	}
+	indent := strings.Repeat("\t", indentLevel)
+	return fmt.Sprintf("%srequiredStatusFields := []string{%s}", indent, requiredStatusFieldsStr)
+}
```
jaypipes, line 400: I actually thought that what we were going to do was output Go code that would return NotFound immediately in `sdkFind()` if there were missing fields in the Input shape.

Currently, the top of the `sdkFind()` method looks like this:

```go
// sdkFind returns SDK-specific information about a supplied resource
func (rm *resourceManager) sdkFind(
	ctx context.Context,
	r *resource,
) (*resource, error) {
	input, err := rm.newDescribeRequestPayload(r)
	if err != nil {
		return nil, err
	}
...
```

what I thought we wanted to do was modify the code that generates the `newDescribeRequestPayload()` so that it returns `ackerr.NotFound` when certain fields that comprise required elements of the Input shape are missing.

```go

// newDescribeRequestPayload returns SDK-specific struct for the HTTP request
// payload of the Describe API call for the resource
func (rm *resourceManager) newDescribeRequestPayload(
	r *resource,
) (*svcsdk.DescribeAPIInput, error) {
	res := &svcsdk.DescribeAPIInput{}

        if r.ko.Status.APIID == nil {
                return ackerr.NotFound
        }

	return res, nil
}
```

vijtrip2:
Closing in favor of https://github.com/aws/aws-controllers-k8s/pull/271

PR #243, opened 2020-08-26
Exclude ModifyCacheParameterGroup operation from generator.yaml

nmvk:
Description of changes:

The attributes of CacheParameterGroup are immutable. This API can be used to update individual key, value
parameters that are part of `CacheParameterGroup`. Without excluding the operation we keep on seeing 
```
2020-08-25T19:28:58.593-0700    ERROR   controller-runtime.controller   Reconciler error        {"controller": "cacheparametergroup", "request": "default/testccr", "error": "InvalidParameter: 1 validation error(s) found.\n- missing required field, ModifyCacheParameterGroupInput.ParameterNameValues.\n"}
```
as Modify operation is continuously invoked.

### DIFF
--- a/services/elasticache/generator.yaml
+++ b/services/elasticache/generator.yaml
@@ -6,5 +6,6 @@ resources:
 ignore:
   operations:
     - DeleteSnapshot
+    - ModifyCacheParameterGroup
   resource_names:
     - GlobalReplicationGroup


jaypipes:
> The attributes of CacheParameterGroup are immutable. This API can be used to update individual key, value
> parameters that are part of `CacheParameterGroup`. Without excluding the operation we keep on seeing

@nmvk I'm confused... if the attributes of CacheParameterGroup are immutable, what does the ModifyCacheParameterGroup API call do?

nmvk:
@jaypipes When we create ParameterGroup for an engine version, https://docs.aws.amazon.com/cli/latest/reference/elasticache/create-cache-parameter-group.html we create a list of parameters for that particular engine version. This list of parameters can be described using `describe-cache-parameters`. 

```
aws elasticache describe-cache-parameters \
    --cache-parameter-group-name "myparamgroup"
```

This will return 
```
{
    "Parameters": [
          {
            "ParameterName": "activedefrag",
            "ParameterValue": "yes",
            "Description": "Enabled active memory defragmentation",
            "Source": "user",
            "DataType": "string",
            "AllowedValues": "yes,no",
            "IsModifiable": true,
            "MinimumEngineVersion": "5.0.0",
            "ChangeType": "immediate"
        }

      & many more parameters
```

Using `Modify` we can update only the values of these parameters and not `cache-parameter-group-family` or `description`

```
aws elasticache modify-cache-parameter-group \
    --cache-parameter-group-name "mygroup" \
    --parameter-name-values "ParameterName=activedefrag, ParameterValue=no"
```

jaypipes:
Ah, OK, yeah, understood now. This actually should be a relatively easy fix/solution. We just need to loop over the Input shape members instead of looping over the Spec/Status fields when generating the Go code that constructs the Input shape to Update operations.

Issue #210, opened 2020-08-18
ElastiCache ReplicationGroup multiple update operations support


kumargauravsharma:
**Is your feature request related to a problem?**
Following ElastiCache APIs are not getting included as part of ack-generate:
- ModifyReplicationGroupShardConfiguration
- IncreaseReplicaCount
- DecreaseReplicaCount

These are related to `ReplicationGroup` and can be considered as modify operation.
CRD for ReplicationGroup is getting auto generated thus `ModifyReplicationGroup` is included.

This issue is to enable support for above mentioned modify operations for ReplicationGroup CR.

**Describe the solution you'd like**
Ability to perform 
- ModifyReplicationGroupShardConfiguration
- IncreaseReplicaCount
- DecreaseReplicaCount

operations on `ReplicationGroup` CR

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

PR #245, opened 2020-08-26
Enable multiple update operations support on single resource

kumargauravsharma:
Issue #210 

Description of changes:
The changes in this PR are to enable custom resource update scenario where Update operation on given resource does not support fields that are supported by its Create method, and there exist specialized methods to update such fields. These specialized method response provides the Resource object indicating the Update on the resource.

For example: `create-replication-group` API arguments include: 
```
num-node-groups
replicas-per-node-group
```
however, `modify-replication-group` API arguments does not, instead `modify-replication-group-shard-configuration` and `increase-replica-count`,  `decrease-replica-count` APIs corresponds to the related Update.

The approach taken in this PR is to add the details of such fields and corresponding custom update methods in the `generator.yaml`. For example:
```
resources:
  ReplicationGroup:
    custom_update_operations:
      UpdateShardConfiguration:
        fields:
          - NumNodeGroups
      UpdateReplicaCount:
        fields:
          - ReplicasPerNodeGroup
```
And custom update methods' (`UpdateShardConfiguration` and `UpdateReplicaCount`) implementation is provided in a separate code file: ```aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/custom_update_api.go```

These methods are supplied with ```DiffReporter``` which helps the code in this file determine the specific Update API to invoke based on the difference in desired, current resource details.

Auto generating following:
```
make build-controller SERVICE=sns
make build-controller SERVICE=ecr
make build-controller SERVICE=s3
```
and after updating bookstore, petstore services files. 
`make test` passed.

Changes include autogenerated sns, ect, s3, and manually updated bookstore, petstore service files per the latest APIs.


pkg/model/crd.go
```diff
@@ -1298,6 +1310,163 @@ func (r *CRD) GoCodeSetOutput(
...
+//			fields:
+//			  - ReplicasPerNodeGroup
+//
+// Then invoking this method as {{ GoCodeCustomUpdate .CRD "ctx" "r" "diffReporter" 1 "resp" }} from template
+// should return:
+//
+//	var ReplicationGroup *svcsdk.ReplicationGroup = nil
+//	for _, diff := range diffReporter.Differences {
+//		switch diff.Path {
+//		case strings.Contains(diff.Path, "ReplicasPerNodeGroup"):
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code to check for exact match.

pkg/model/crd.go
```diff
@@ -1298,6 +1310,163 @@ func (r *CRD) GoCodeSetOutput(
...
+//			outputShapeName, respErr := rm.UpdateReplicaCount(ctx, r, diffReporter)
```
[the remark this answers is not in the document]
kumargauravsharma: As it corresponds to custom logic, it is not auto generated, file: aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/custom_update_api.go contains the custom logic methods which generator yaml refers.

pkg/model/crd.go
```diff
@@ -1298,6 +1310,163 @@ func (r *CRD) GoCodeSetOutput(
...
+//			fields:
+//			  - NumNodeGroups
+//		  UpdateReplicaCount:
+//			fields:
+//			  - ReplicasPerNodeGroup
+//
+// Then invoking this method as {{ GoCodeCustomUpdate .CRD "ctx" "r" "diffReporter" 1 "resp" }} from template
+// should return:
+//
+//	var ReplicationGroup *svcsdk.ReplicationGroup = nil
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code. This variable represent the updated resource from API response.

pkg/model/crd.go
```diff
@@ -1298,6 +1310,163 @@ func (r *CRD) GoCodeSetOutput(
...
+//			ReplicationGroup = outputShapeName
+//		case strings.Contains(diff.Path, "NumNodeGroups"):
+//			outputShapeName, respErr := rm.UpdateShardConfiguration(ctx, r, diffReporter)
+//			if respErr != nil {
+//				return nil, respErr
+//			}
+//			ReplicationGroup = outputShapeName
+//		}
+//		if ReplicationGroup != nil {
+//			resp = *svcsdk.ModifyReplicationGroupOutput{ReplicationGroup}
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code, removed the need to construct `resp`.

pkg/util/diff_reporter.go
```diff
@@ -0,0 +1,46 @@
+package util
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code, refactored the file per the comment.

pkg/runtime/reconciler.go
```diff
@@ -164,7 +165,11 @@ func (r *reconciler) sync(
 			"arn", latest.Identifiers().ARN(),
 			"is_adopted", isAdopted,
 		)
-		latest, err = rm.Update(ctx, desired)
+
+		var diffReporter util.DiffReporter
+		cmp.Equal(latest, desired, cmp.Reporter(&diffReporter))
```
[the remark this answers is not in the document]
kumargauravsharma: Kept the original diff logic as is, and added the logic for reporter reference to pass to `rm.Update` thus, did not move it up. Please review.

pkg/util/diff_reporter.go
```diff
@@ -0,0 +1,46 @@
...
+	Path string
+	ValueA reflect.Value
+	ValueB reflect.Value
+}
+
+func (diff *DiffItem) String() string {
+	return fmt.Sprintf("%#v:\n\t-: %+v\n\t+: %+v\n", diff.Path, diff.ValueA, diff.ValueB)
+}
+
+type DiffReporter struct {
```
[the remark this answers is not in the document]
kumargauravsharma: Renamed per the comment.

pkg/util/diff_reporter.go
```diff
@@ -0,0 +1,46 @@
...
+	DiffPath		cmp.Path
```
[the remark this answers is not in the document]
kumargauravsharma: Renamed per the comment.

pkg/generate/config/config.go
```diff
@@ -166,6 +179,10 @@ type ExceptionsConfig struct {
 	Codes map[int]string `json:"codes"`
 }
 
+type CustomUpdateOperation struct {
```
jaypipes, line 182: this is unused, AFAICT?

pkg/generate/config/config.go
```diff
@@ -86,6 +94,11 @@ type ResourceConfig struct {
 	ListOperation *ListOperationConfig `json:"list_operation,omitempty"`
 }
 
+// resource fields that map to custom operation
+type CustomUpdateOperationFields struct {
```
jaypipes, line 98: Let's call this type name `CustomUpdateOperationConfig` to match the other type names.

pkg/generate/config/config.go
```diff
@@ -86,6 +94,11 @@ type ResourceConfig struct {
...
+	Fields []string `json:"fields"`
```
jaypipes, line 99: Please add a docstring comment explaining what should go in the `fields` `[]string`.

pkg/model/crd.go
```diff
@@ -1210,7 +1222,11 @@ func (r *CRD) GoCodeSetOutput(
 	if outputShape.UsedAsOutput && len(outputShape.MemberRefs) == 1 {
 		for memberName, memberRef := range outputShape.MemberRefs {
 			if memberRef.Shape.Type == "structure" {
-				sourceVarName += "." + memberName
+				if sourceVarName == "" {
+					sourceVarName = memberName
+				} else {
+					sourceVarName += "." + memberName
+				}
```
jaypipes, line 1229: This change should be reverted.

pkg/model/crd.go
```diff
@@ -1356,6 +1372,197 @@ func (r *CRD) GoCodeSetOutput(
...
+//				}
+//				ReplicationGroup = outputShapeName
+//			case "Spec.NumNodeGroups":
+//				outputShapeName, respErr := rm.UpdateShardConfiguration(ctx, r, diffReporter)
+//				if respErr != nil {
+//					return nil, respErr
+//				}
+//				ReplicationGroup = outputShapeName
+//			}
+//			if ReplicationGroup != nil {
```
jaypipes, line 1410: What happens if there are mutations to **both** the `Spec.ReplicasPerNodeGroup` **and** `Spec.NumNodeGroups` fields? In this code, only one of those calls to the custom callbacks will be triggered due to this line and `break`, no?

pkg/model/crd.go
```diff
@@ -1356,6 +1372,197 @@ func (r *CRD) GoCodeSetOutput(
...
+//			}
+//		}
+//		if ReplicationGroup == nil {
+//			resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
+//			if respErr != nil {
+//				return nil, respErr
+//			}
+//			ReplicationGroup = resp.ReplicationGroup
+//		}
+
```
jaypipes, line 1421: nit: remove this blank line.

pkg/model/crd.go
```diff
@@ -1356,6 +1372,197 @@ func (r *CRD) GoCodeSetOutput(
...
+//		  UpdateShardConfiguration:
+//			fields:
+//			  - NumNodeGroups
+//		  UpdateReplicaCount:
+//			fields:
+//			  - ReplicasPerNodeGroup
+//
+// Then invoking this method as {{ GoCodeCustomUpdate .CRD "ctx" "r" "diffReporter" 1 }} from template
+// should return:
+//		var ReplicationGroup *svcsdk.ReplicationGroup = nil
```
jaypipes, line 1394: I'm still not clear what this `ReplicationGroup` variable actually is needed for...

pkg/model/crd.go
```diff
@@ -1356,6 +1372,197 @@ func (r *CRD) GoCodeSetOutput(
...
+//			if ReplicationGroup != nil {
+//				break // for loop
+//			}
+//		}
+//		if ReplicationGroup == nil {
+//			resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
+//			if respErr != nil {
+//				return nil, respErr
+//			}
+//			ReplicationGroup = resp.ReplicationGroup
```
jaypipes, line 1419: This block of code is already output in the normal `GoCodeSetOutput` code path for the Update operation... you should not need to have this here.

pkg/types/aws_resource_descriptor.go
```diff
@@ -42,6 +43,13 @@ type AWSResourceDescriptor interface {
 	// should be the same. In other words, the Diff() method should be called
 	// with the same concrete implementing AWSResource type
 	Diff(AWSResource, AWSResource) string
+
+	// DiffReport returns a Reporter which provides the difference between two supplied
+	// AWSResources. The underlying types of the two supplied AWSResources should
+	// be the same. In other words, the DiffReport() method should be called with the
+	// same concrete implementing AWSResource type
+	DiffReport(AWSResource, AWSResource) *ackcompare.Reporter
```
jaypipes, line 51: No need to add another `DiffReport` method to this interface. Simply go ahead and change the existing `Diff` method to return an `*ackcompare.Reporter` instead of a string.

Then, on line 165-171 of reconciler.go:

```go
		diff := r.rd.Diff(desired, latest)
		r.log.V(1).Info(
			"desired resource state has changed",
			"diff", diff,
			"arn", latest.Identifiers().ARN(),
			"is_adopted", isAdopted,
		)
```

simply change it to this:

```go
		diff := r.rd.Diff(desired, latest)
		r.log.V(1).Info(
			"desired resource state has changed",
			"diff", diff.String(),
			"arn", latest.Identifiers().ARN(),
			"is_adopted", isAdopted,
		)
```

services/ecr/apis/v1alpha1/enums.go
```diff
@@ -48,7 +41,6 @@ const (
 	ImageFailureCode_ImageNotFound                 ImageFailureCode = "ImageNotFound"
 	ImageFailureCode_MissingDigestAndTag           ImageFailureCode = "MissingDigestAndTag"
 	ImageFailureCode_ImageReferencedByManifestList ImageFailureCode = "ImageReferencedByManifestList"
-	ImageFailureCode_KmsError                      ImageFailureCode = "KmsError"
```
jaypipes, line 51: Please revert these changes. They were due to you running `make build-controller SERVICE=ecr` with an older version of `aws-sdk-go` :)

jaypipes (changes requested):
Still a number of things that I think need to be fixed up about this, @kumargauravsharma. Please see inline comments.

Also, it looks like you make have run `make build-controller` and had an older version of `aws-sdk-go` Go module installed. You will want to run `make test` before running `make build-controller` again and make sure the latest aws-sdk-go Go module is downloaded.

Finally, I would support checking in a generated version of the Elasticache service controller in a separate PR **before this one**. That way, we can better see the changes to the generated service controller code that this change introduces. Would you like me to check the Elasticache service controller code in and you then rebase onto main?

templates/pkg/crd_sdk.go.tpl
```diff
@@ -188,22 +189,38 @@ func (rm *resourceManager) newCreateRequestPayload(
...
+    {{ if .CRD.HasCustomUpdateOperations }}
+
+    {{ $setCode := GoCodeSetUpdateOutput .CRD "" "ko.Status" 1 }}
+    {{ GoCodeCustomUpdate .CRD "ctx" "r" "diffReporter" 1 }}
+    // Merge in the information we read from the API call above to the copy of
+    // the original Kubernetes object we passed to the function
+    ko := r.ko.DeepCopy()
+    {{ $setCode }}
+
+    {{ else }}
```
jaypipes, line 209: Here, I was thinking you could just use `{{ end }}` instead of  `{{ else }}`. That way, you could place a good chunk of the static Go code generation here, for example, you could do something like this:

```
    {{ if .CRD.HasCustomUpdateOperations }}
		for _, diff := range diffReporter.Differences {
			switch diff.Path {
        {{- range $diffPath, $customMethod := .CRD.GetCustomUpdateOperations }}
			case "{{ $diffPath }}":
				if err := rm.UpdateReplicaCount(ctx, r, diffReporter); err != nil {
					return nil, err
				}
         {{- end }}
			}
    {{ end }}
```

And you wouldn't need any of the `CRD.GoCodeCustomOutput` method at all...

templates/pkg/crd_sdk.go.tpl
```diff
@@ -188,22 +189,38 @@ func (rm *resourceManager) newCreateRequestPayload(
...
 		return nil, respErr
 	}
-	// Merge in the information we read from the API call above to the copy of
-	// the original Kubernetes object we passed to the function
-	ko := r.ko.DeepCopy()
-{{ $setCode }}
+    // Merge in the information we read from the API call above to the copy of
+    // the original Kubernetes object we passed to the function
+    ko := r.ko.DeepCopy()
+    {{ $setCode }}
```
jaypipes, line 219: Make sure you replace four spaces with `\t` characters when in templates :)

pkg/generate/config/config.go
kumargauravsharma: Removed the unused block of code.

pkg/generate/config/config.go
kumargauravsharma: Updated the type name.

pkg/generate/config/config.go
kumargauravsharma: Added docstring.

pkg/model/crd.go
kumargauravsharma: Reverted.

pkg/model/crd.go
kumargauravsharma: Updated the code to `return` after invoking custom update method. Next reconciler invocation would lead to next custom update method call (if resource is in modifiable state).

pkg/model/crd.go
kumargauravsharma: Code removed.

pkg/model/crd.go
kumargauravsharma: Code removed.

pkg/model/crd.go
kumargauravsharma: Code removed.

pkg/types/aws_resource_descriptor.go
kumargauravsharma: Code updated.

services/ecr/apis/v1alpha1/enums.go
kumargauravsharma: The `make build-controller SERVICE=ecr` was producing this even after clearing go mod caches.

I find that setting `ACK_GENERATE_CACHE_DIR` environment variable to a value resulted in resolving this issue.
* The default value `~/.cache/aws-controllers-k8s` (set inside `build-controller.sh`) did not create any `.cache` directory inside user `home`.

latest PR includes files generated using 
```
make build-controller SERVICE=elasticache && make build-controller SERVICE=sns && make build-controller SERVICE=ecr && make build-controller SERVICE=s3
```

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Thanks for the suggestion. I have applied it in the code.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Code updated.

pkg/compare/reporter.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func (reporter *Reporter) String() string {
+	var diffs []string
+	for _, diff := range reporter.Differences {
+		diffs = append(diffs, diff.String())
+	}
+	return strings.Join(diffs, "\n")
+}
```
jaypipes, line 45: :+1:

pkg/runtime/reconciler.go
```diff
@@ -163,14 +163,14 @@ func (r *reconciler) sync(
...
+		diffReporter := r.rd.Diff(desired, latest)
 		r.log.V(1).Info(
 			"desired resource state has changed",
-			"diff", diff,
+			"diff", diffReporter.String(),
 			"arn", latest.Identifiers().ARN(),
 			"is_adopted", isAdopted,
 		)
-		latest, err = rm.Update(ctx, desired)
+		latest, err = rm.Update(ctx, desired, diffReporter)
```
jaypipes, line 173: :+1:

pkg/types/aws_resource_descriptor.go
```diff
@@ -37,11 +38,11 @@ type AWSResourceDescriptor interface {
...
-	// Diff returns a string representing the difference between two supplied
-	// AWSResources/ The underlying types of the two supplied AWSResources
-	// should be the same. In other words, the Diff() method should be called
-	// with the same concrete implementing AWSResource type
-	Diff(AWSResource, AWSResource) string
+	// Diff returns a Reporter which provides the difference between two supplied
+	// AWSResources. The underlying types of the two supplied AWSResources should
+	// be the same. In other words, the Diff() method should be called with the
+	// same concrete implementing AWSResource type
+	Diff(AWSResource, AWSResource) *ackcompare.Reporter
```
jaypipes, line 45: :+1:

pkg/generate/config/config.go
```diff
@@ -86,6 +96,14 @@ type ResourceConfig struct {
 	ListOperation *ListOperationConfig `json:"list_operation,omitempty"`
 }
 
+// Custom resource fields paths list that map to custom operation
+type CustomUpdateOperationConfig struct {
+	// string paths to fields on the custom resource
+	// Example:
+	//  For field ReplicationGroup.Spec.ReplicasPerNodeGroup, the string value would be: Spec.ReplicasPerNodeGroup
+	DiffPaths []string `json:"diff_paths"`
```
jaypipes, line 104: Is there a reason we want DiffPaths to be a slice of string instead of just a string? Are there situations where a single update custom update method would be fired for **multiple** DiffPaths?

If there are no situations like that, probably would be simpler to get rid of `CustomUpdateOperationConfig` entirely and just have `ResourceConfig.CustomUpdateOperations` Go type be `map[string]string` where the map keys are the names of the custom methods on the `resourceManager` and the map values are the DiffPath that triggers that call to a custom method.

Thoughts?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -0,0 +1,388 @@
...
+				if respErr != nil {
+					fmt.Printf("Failed to DecreaseReplicaCountWithContext. Error: %v\n", respErr)
+					return nil, respErr
+				}
+				return provideUpdatedResource(r, resp.ReplicationGroup)
+			}
+
+			break
+		}
+	}
```
jaypipes, line 88: just a stylistic suggestion... you can remove at least one level of indentation in this function by just "searching" for the `DiffItem` and returning an error early if you don't find it. Something like this:

```go
func (rm *resourceManager) UpdateReplicaCount(
	ctx context.Context,
	r *resource,
	diffReporter *ackcompare.Reporter,
) (*resource, error) {

        var di *ackcompare.DiffItem
	for _, diff := range diffReporter.Differences {
		if diff.Path == "Spec.ReplicasPerNodeGroup" {
                        di = diff
			break
		}
	}			
        if di == nil {
                 return nil, fmt.Errorf("expected different ReplicasPerNodeGroup.")
        }

        desired, err1 := strconv.Atoi(diff.ValueA)
        latest, err2 := strconv.Atoi(diff.ValueB)

	if err1 != nil || err2 != nil {
		return nil, fmt.Errorf("UpdateReplicaCount failed: invalid values")
	}

	if latest < desired { // increase
		fmt.Printf("Requesting Increase Replica Count. Old value: %v, New Value: %v\n", latest, desired)
		input, err := rm.newIncreaseReplicaCountRequestPayload(r)
		if err != nil {
			fmt.Printf("Error occurred: %v\n", err)
			return nil, err
		}
		resp, respErr := rm.sdkapi.IncreaseReplicaCountWithContext(ctx, input)
		if respErr != nil {
			fmt.Printf("Failed to IncreaseReplicaCountWithContext. Error: %v\n", respErr)
			return nil, respErr
		}
		return provideUpdatedResource(r, resp.ReplicationGroup)
	}
        // decrease
	input, err := rm.newDecreaseReplicaCountRequestPayload(r)
	fmt.Printf("Requesting Decrease Replica Count. Old value: %v, New Value: %v\n", latest, desired)
	if err != nil {
		fmt.Printf("Error occurred: %v\n", err)
		return nil, err
	}
	resp, respErr := rm.sdkapi.DecreaseReplicaCountWithContext(ctx, input)
	if respErr != nil {
		fmt.Printf("Failed to DecreaseReplicaCountWithContext. Error: %v\n", respErr)
		return nil, respErr
	}
	return provideUpdatedResource(r, resp.ReplicationGroup)
}
```

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -0,0 +1,388 @@
...
+
+			if err1 != nil || err2 != nil {
+				return nil, fmt.Errorf("UpdateReplicaCount failed: invalid values")
+			}
+
+			if latest < desired { // increase
+				fmt.Printf("Requesting Increase Replica Count. Old value: %v, New Value: %v\n", latest, desired)
+				input, err := rm.newIncreaseReplicaCountRequestPayload(r)
+				if err != nil {
+					fmt.Printf("Error occurred: %v\n", err)
```
jaypipes, line 62: Don't print the error, just return it up the call chain. The reconciler will log the error appropriately. :)

Here and throughout this file...

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -0,0 +1,388 @@
...
+				}
+				resp, respErr := rm.sdkapi.IncreaseReplicaCountWithContext(ctx, input)
+				if respErr != nil {
+					fmt.Printf("Failed to IncreaseReplicaCountWithContext. Error: %v\n", respErr)
+					return nil, respErr
+				}
+				return provideUpdatedResource(r, resp.ReplicationGroup)
+			} else { // decrease
+				input, err := rm.newDecreaseReplicaCountRequestPayload(r)
+				fmt.Printf("Requesting Decrease Replica Count. Old value: %v, New Value: %v\n", latest, desired)
```
jaypipes, line 73: this is fine for now. Eventually we'll plumb a logger down into the resourceManager so you can use `rm.log.V(1).Info()` here for debug log statements.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+		res.SetReplicaConfiguration(f13)
+	}
+
+	return res, nil
+}
+
+// This method copies the data from given replicationGroup by populating it into copy of supplied resource
+// and returns it.
+func provideUpdatedResource(
```
jaypipes, line 234: :+1: I know it's a bit of a copy/paste pain right now, but we'll work together to auto-generate these kind of helper setter methods in a future PR :)

pkg/generate/config/config.go
kumargauravsharma: There are multiple DiffPaths that can lead to a custom method. For example, replication group create API supports:
```
[--num-node-groups   <value>]
[--replicas-per-node-group   <value>]
```
values for these are numbers and service handles creation of node-groups (shards), replicas in each shard per the provided number. Service handles determining the fine-grained details for shards, replicas itself.

However, replication group create API also supports fine grained option to configure each shard during create:
```
[--node-group-configuration   <value>]
```
here the value is an array where each entry in array provide fine-grained values for the node-groups (shards) and replicas:
```
[
{
    "NodeGroupId":   "string",
    "Slots": "string",
    "ReplicaCount": integer,
    "PrimaryAvailabilityZone":   "string",
    "ReplicaAvailabilityZones":   ["string", ...]
},
...
]
```

Thus, there are multiple fields that can lead to custom update methods.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: 👍 updated code.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: During debugging it helped to have error message discoverable in code as the reconciler error logs did not provide location where error occurred. Thus, added print messages with error and messages.

Updated code and removed the printf statements related to errors.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: 👍

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: 👍

jaypipes (approved):
Good to go. I still think we might be able to simplify/clarify the CustomUpdateOperations struct slightly but meh, we can always improve it over time.

Issue #102, opened 2020-07-24
Cross account resource management (CARM)


A-Hilaly:
This issue is about AWS accounts management within a single Kubernetes cluster. 

We want to:

- Allow users to securely create resources across multiple AWS accounts without having to own/manage multiple Kubernetes clusters.
- Help users maintain/administer controllers access to stored AWS accounts.
- Limit controllers permissions when calling AWS API.
- Prevent unauthorized users from making CRUD operations on resources on a different AWS account.

Link to [CARM design](https://github.com/aws/aws-controllers-k8s/blob/main/docs/design/proposals/carm/cross-account-resource-management.md) proposal

Related to: #3 and #25

PR #255, opened 2020-08-30
CARM cache implementation

A-Hilaly:
Issue #, if available: #102 

Description of changes:

Adding functionalities to help caching and accessing CARM configmap and namespace ACK related annotations.

- [x] Namespace annotations cache
- [x] CARM ConfigMap (`ack-role-account-map`) cache
- [x] Caches unit tests

### DIFF
--- a/apis/core/v1alpha1/annotations.go
+++ b/apis/core/v1alpha1/annotations.go
@@ -25,7 +25,7 @@ const (
 	// CR, that means the user expects the ACK service controller to create the
 	// backend AWS service API resource.
 	AnnotationARN = AnnotationPrefix + "arn"
-	// AnnotationOwnerAccountID is an annotation whoe value is the identifier
+	// AnnotationOwnerAccountID is an annotation whose value is the identifier
 	// for the AWS account to which the resource belongs.  If this annotation
 	// is set on a CR, the Kubernetes user is indicating that the ACK service
 	// controller should create/patch/delete the resource in the specified AWS
@@ -36,4 +36,20 @@ const (
 	// TODO(jaypipes): Link to documentation on cross-account resource
 	// management
 	AnnotationOwnerAccountID = AnnotationPrefix + "owner-account-id"
+	// AnnotationRegion is an annotation whose value is the identifier for the
+	// the AWS region in which the resources should be created. If this annotation
+	// is set on a CR metadata, that means the user is indicating to the ACK service
+	// controller that the CR should be created on specific region. ACK service
+	// controller will not override the resource region if this annotation is set.
+	AnnotationRegion = AnnotationPrefix + "region"
+	// AnnotationDefaultRegion is an annotation whose value is the identifier
+	// for the default AWS region in which resources should be created. If this
+	// annotation is set on a namespace, the Kubernetes user is indicating that
+	// the ACK service controller should set the regions in which the resource
+	// should be created, if a region annotation is not set on the CR metadata.
+	// If this annotation - and AnnotationRegion - are not set, ACK service
+	// controllers look for controller binary flags and environment variables
+	// injected by POD IRSA, to decide in which region the resources should be
+	// created.
+	AnnotationDefaultRegion = AnnotationPrefix + "default-region"
 )
--- a/pkg/runtime/cache/account.go
+++ b/pkg/runtime/cache/account.go
@@ -1,0 +1,115 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package cache
+
+import (
+	"sync"
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	informersv1 "k8s.io/client-go/informers/core/v1"
+	kubernetes "k8s.io/client-go/kubernetes"
+	k8scache "k8s.io/client-go/tools/cache"
+)
+
+const (
+	// ACKRoleAccountMap is the name of the configmap map object storing
+	// all the AWS Account IDs associated with their AWS Role ARNs.
+	ACKRoleAccountMap = "ack-role-account-map"
+)
+
+// AccountCache is responsible for caching the CARM configmap
+// data. It is listening to all the events related to the CARM map and
+// make the changes accordingly.
+type AccountCache struct {
+	sync.RWMutex
+
+	log logr.Logger
+
+	// ConfigMap informer
+	informer k8scache.SharedInformer
+	roleARNs map[string]string
+}
+
+// NewAccountCache makes a new AccountCache from a client.Interface
+// and a logr.Logger
+func NewAccountCache(clientset kubernetes.Interface, log logr.Logger) *AccountCache {
+	sharedInformer := informersv1.NewConfigMapInformer(
+		clientset,
+		currentNamespace,
+		informerResyncPeriod,
+		k8scache.Indexers{},
+	)
+	return &AccountCache{
+		informer: sharedInformer,
+		log:      log.WithName("AccountCache"),
+		roleARNs: make(map[string]string),
+	}
+}
+
+// resourceMatchACKRoleAccountConfigMap verifies if a resource is
+// the CARM configmap. It verifies the name, namespace and object type.
+func resourceMatchACKRoleAccountsConfigMap(raw interface{}) bool {
+	object, ok := raw.(*corev1.ConfigMap)
+	return ok && object.ObjectMeta.Name == ACKRoleAccountMap
+}
+
+// Run adds the default event handler functions to the SharedInformer and
+// runs the informer to begin processing items.
+func (c *AccountCache) Run(stopCh <-chan struct{}) {
+	c.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if resourceMatchACKRoleAccountsConfigMap(obj) {
+				object := obj.(*corev1.ConfigMap).DeepCopy()
+				c.log.V(1).Info("ack-role-account-map has been created")
+				c.updateAccountRoleData(object.Data)
+				c.log.V(1).Info("cached ack-role-account-map data")
+			}
+		},
+		UpdateFunc: func(orig, desired interface{}) {
+			if resourceMatchACKRoleAccountsConfigMap(desired) {
+				object := desired.(*corev1.ConfigMap).DeepCopy()
+				c.log.V(1).Info("ack-role-account-map has been updated")
+				//TODO(a-hilaly): compare data checksum before updating the cache
+				c.updateAccountRoleData(object.Data)
+				c.log.V(1).Info("cached ack-role-account-map data")
+			}
+		},
+		DeleteFunc: func(obj interface{}) {
+			if resourceMatchACKRoleAccountsConfigMap(obj) {
+				c.log.V(1).Info("ack-role-account-map has been deleted")
+				newMap := make(map[string]string)
+				c.updateAccountRoleData(newMap)
+				c.log.V(1).Info("cleaned up role account map")
+			}
+		},
+	})
+	go c.informer.Run(stopCh)
+}
+
+// GetAccountRoleARN queries the AWS accountID associated Role ARN
+// from the cached CARM configmap. This function is thread safe.
+func (c *AccountCache) GetAccountRoleARN(accountID string) (string, bool) {
+	c.RLock()
+	defer c.RUnlock()
+	roleARN, ok := c.roleARNs[accountID]
+	return roleARN, ok
+}
+
+// updateAccountRoleData updates the CARM map. This function is thread safe.
+func (c *AccountCache) updateAccountRoleData(data map[string]string) {
+	c.Lock()
+	defer c.Unlock()
+	c.roleARNs = data
+}
--- a/pkg/runtime/cache/account_test.go
+++ b/pkg/runtime/cache/account_test.go
@@ -1,0 +1,143 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package cache_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zapcore"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	k8sfake "k8s.io/client-go/kubernetes/fake"
+	k8stesting "k8s.io/client-go/testing"
+	ctrlrtzap "sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	ackrtcache "github.com/aws/aws-controllers-k8s/pkg/runtime/cache"
+)
+
+const (
+	testNamespace = "ack-system"
+
+	testAccount1    = "012345678912"
+	testAccountARN1 = "arn:aws:iam::012345678912:role/S3Access"
+	testAccount2    = "219876543210"
+	testAccountARN2 = "arn:aws:iam::012345678912:role/root"
+)
+
+func TestAccountCache(t *testing.T) {
+	accountsMap1 := map[string]string{
+		testAccount1: testAccountARN1,
+	}
+
+	accountsMap2 := map[string]string{
+		testAccount1: testAccountARN1,
+		testAccount2: testAccountARN2,
+	}
+
+	// create a fake k8s client and a fake watcher
+	k8sClient := k8sfake.NewSimpleClientset()
+	watcher := watch.NewFake()
+	k8sClient.PrependWatchReactor("configMaps", k8stesting.DefaultWatchReactor(watcher, nil))
+
+	zapOptions := ctrlrtzap.Options{
+		Development: true,
+		Level:       zapcore.InfoLevel,
+	}
+	fakeLogger := ctrlrtzap.New(ctrlrtzap.UseFlagOptions(&zapOptions))
+
+	// initlizing account cache
+	accountCache := ackrtcache.NewAccountCache(k8sClient, fakeLogger)
+	stopCh := make(chan struct{})
+	accountCache.Run(stopCh)
+
+	// Test create events
+	k8sClient.CoreV1().ConfigMaps(testNamespace).Create(
+		context.Background(),
+		&corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "random-map",
+			},
+			Data: accountsMap1,
+		},
+		metav1.CreateOptions{},
+	)
+
+	time.Sleep(time.Second)
+
+	_, ok := accountCache.GetAccountRoleARN("random-account")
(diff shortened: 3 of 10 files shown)


pkg/runtime/reconciler.go
```diff
@@ -37,11 +37,13 @@ import (
...
-	rmf acktypes.AWSResourceManagerFactory
-	rd  acktypes.AWSResourceDescriptor
-	log logr.Logger
-	cfg Config
+	kc           client.Client
+	rmf          acktypes.AWSResourceManagerFactory
+	rd           acktypes.AWSResourceDescriptor
+	log          logr.Logger
+	cfg          Config
+	nsaCache     *namespaceAnnotationsCache
```
jaypipes, line 45: probably cleaner to have a `pkg/runtime/cache.Caches` object which itself contains all the different caches.

Something like this:

```go
type Caches struct {
    Namespaces *NamespaceCache
    Accounts *AccountCache
}
```

along with a `pkg/runtime/cache.New` function that returns an initialized `pkg/runtime/cache.Caches` object:

```go
func New(kc clientset.Interface, log logr,Log) Caches {
    return Caches{
        Namespaces: newNamespaceCache(kc. log),
        Accounts: newAccountCache(kc, log),
    }
}
```

and then the `reconciler` can use it like so:

```go
import (
...
    ackrtcache "github.com/aws/aws-controllers-k8s/pkg/runtime/cache"
...
type reconciler struct {
...
    cache ackrtcache.Caches
}
...
    if r.cache.Namespaces.GetByKey(...) {
...
```

pkg/runtime/cache.go
```diff
@@ -0,0 +1,254 @@
+package runtime
```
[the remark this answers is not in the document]
A-Hilaly: yep +1

pkg/runtime/cache.go
```diff
@@ -0,0 +1,254 @@
...
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
+)
+
+const (
+	// ACKDefaultNamespace is the ACK default namespace, where
+	// service controllers and CARM configmap are created
+	ACKDefaultNamespace = "ack-system"
```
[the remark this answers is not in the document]
A-Hilaly: I'm currently experimenting this: trying to query the current namespace and use it to track the Accounts configMap. It is possible to inject the namespace as an environment variable using the [downward](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api). WDYT @jaypipes  ?

pkg/runtime/cache.go
```diff
@@ -0,0 +1,254 @@
...
+	// informerDefaultResyncPeriod is the period at which ShouldResync
+	// is considered.
+	// TODO: find... and set a default resync period
+	informerDefaultResyncPeriod = 0 * time.Second
+)
+
+// awsAccountRoleARNCache is responsible for caching the CARM configmap
+// data. It is listening to all the events related to the CARM map and
+// make the changes accordingly.
+type awsAccountRoleARNCache struct {
```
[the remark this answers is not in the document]
A-Hilaly: Agreed. Simpler is better :)

pkg/runtime/cache.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+}
+
+// resourceMarchCARMConfigMap verifies if a resource is the CARM
+// configmap. It verifies the name, namespace and object type.
+func resourceMatchCARMConfigMap(raw interface{}) bool {
+	object, ok := raw.(*corev1.ConfigMap)
+	return ok &&
+		object.ObjectMeta.Name == CARMConfigMapName &&
+		object.ObjectMeta.Namespace == ACKDefaultNamespace
```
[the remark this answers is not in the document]
A-Hilaly: Aaaah! 👍

pkg/runtime/cache.go
```diff
@@ -0,0 +1,254 @@
...
+		informer:                 sharedInformer,
+		log:                      log.WithName("nsAnnotationsCache"),
+		namespacesACKAnnotations: make(map[string]*nsAnnotations),
+	}
+}
+
+// isNotIgnoredNamespace returns true if an object is of type
+// corev1.Namespace and is not one of 'ack-system', 'kube-system',
+// or 'kube-public'
+func isNotIgnoredNamespace(raw interface{}) bool {
```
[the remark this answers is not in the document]
A-Hilaly: Ohhh! bien vue! 😄

pkg/runtime/reconciler.go
A-Hilaly: Great will do that!

apis/core/v1alpha1/annotations.go
```diff
@@ -36,4 +36,20 @@ const (
 	// TODO(jaypipes): Link to documentation on cross-account resource
 	// management
 	AnnotationOwnerAccountID = AnnotationPrefix + "owner-account-id"
+	// AnnotationRegion is an annotation whose value is the identifier for the
+	// the AWS region in which the resources should be created. If this annotation
+	// is set on a CR metadata, that means the user is indicating to the ACK service
+	// controller that the CR should be created on specific region. ACK service
+	// controller will not override the resource region if this annotation is set.
+	AnnotationRegion = AnnotationPrefix + "aws-region"
```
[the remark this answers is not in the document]
A-Hilaly: Agreed! It is not necessary. Removing it.

pkg/runtime/cache.go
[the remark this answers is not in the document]
A-Hilaly: ^^ 
https://github.com/A-Hilaly/aws-controllers-k8s/blob/carm-impl/templates/config/controller/deployment.yaml.tpl#L44-L48
https://github.com/A-Hilaly/aws-controllers-k8s/blob/carm-impl/pkg/runtime/cache/cache.go#L35-L44

pkg/runtime/cache/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
```
jaypipes, line 13: Please rename to `pkg/runtime/cache/account.go` (singular)

Sorry about that!

pkg/runtime/cache/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+		log:      log.WithName("AccountCache"),
+		roleARNs: make(map[string]string),
+	}
+}
+
+// resourceMatchACKRoleAccountConfigMap verifies if a resource is
+// the CARM configmap. It verifies the name, namespace and object type.
+func resourceMatchACKRoleAccountsConfigMap(raw interface{}) bool {
+	object, ok := raw.(*corev1.ConfigMap)
+	return ok && object.ObjectMeta.Namespace == currentNamespace
```
jaypipes, line 66: Unless I'm missing something, you need to check the ObjectMeta.Name, not the ObjectMeta.Namespace :)

pkg/runtime/cache/accounts.go
```diff
@@ -0,0 +1,125 @@
...
+				c.log.Info("cleaned up cache")
+			}
+		},
+	})
+
+	go c.informer.Run(stopCh)
+}
+
+// GetByKey returns the account role ARN from the cached configmap
+func (c *AccountCache) GetByKey(key string) (string, error) {
```
jaypipes, line 104: Maybe call this `GetRoleARN()`? And change the variable name "key" to "accountID"

pkg/runtime/cache/cache.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// New creates a new Caches object from a clientset.Interface and
+// a logr.Logger
+func New(kc clientset.Interface, log logr.Logger) Caches {
+	return Caches{
+		Accounts:   NewAccountCache(kc, log),
+		Namespaces: NewNamespaceCache(kc, log),
+	}
+}
```
jaypipes, line 62: :+1:

pkg/runtime/cache/namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+// isIgnoredNamespace returns verifies if an object is of type
+// corev1.Namespace and is not one of 'ack-system', 'kube-system',
+// or 'kube-public'
+func isIgnoredNamespace(raw interface{}) bool {
+	object, ok := raw.(*corev1.Namespace)
+	return ok &&
+		object.ObjectMeta.Name != "ack-system" &&
```
jaypipes, line 86: these checks should be `==` with `||`, not `!=` with `&&`

pkg/runtime/cache/namespaces.go
```diff
@@ -0,0 +1,169 @@
...
+func NewNamespaceCache(kc clientset.Interface, log logr.Logger) *NamespaceCache {
+	sharedInformer := informersv1.NewNamespaceInformer(
+		kc,
+		informerResyncPeriod,
+		k8scache.Indexers{},
+	)
+	return &NamespaceCache{
+		informer:       sharedInformer,
+		log:            log.WithName("NamespaceCache"),
+		namespaceInfos: make(map[string]*namespaceInfo),
```
jaypipes, line 76: how is the `namespaceInfos` map going to be populated on startup of the service controller?

pkg/runtime/reconciler.go
```diff
@@ -327,13 +329,31 @@ func (r *reconciler) getOwnerAccountID(
...
+	if err != nil {
+		return ackv1alpha1.AWSRegion(r.cfg.Region)
+	}
+
+	nsDefaultRegion := nsAnnotations.GetDefaultRegion()
+	if nsDefaultRegion == "" {
+		return ackv1alpha1.AWSRegion(r.cfg.Region)
+	}
+
+	return ackv1alpha1.AWSRegion(nsDefaultRegion)
```
jaypipes, line 356: :+1:

pkg/runtime/reconciler.go
```diff
@@ -342,6 +362,8 @@ func NewReconciler(
 	log logr.Logger,
 	cfg Config,
 ) acktypes.AWSResourceReconciler {
+	// TODO(a-hilaly): initialize cache, and figure out how to
+	// bring a k8s.Client to this level
```
jaypipes, line 366: You can initialize the caches in BindControllerManager(), after line 64.

pkg/runtime/cache.go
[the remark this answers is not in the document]
jaypipes: Yeah, that would definitely work, Amine. ++

jaypipes (changes requested):
This is getting very close, thank you @A-Hilaly! Few more suggestions inline...

pkg/runtime/cache/account_test.go
```diff
@@ -0,0 +1,145 @@
...
+		&corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "random-map",
+			},
+			Data: accountsMap1,
+		},
+		metav1.CreateOptions{},
+	)
+
+	time.Sleep(time.Second)
```
A-Hilaly, line 82: I'm not really comfortable using `time.Sleep()` calls in this test, and `SharedInformer.HasSynced` isn't really useful since it will always return `true` after the first informer sync.

Multiple other solutions are possible but all of them imply changes the Cache public functions signatures:

1. expose `SharedInformer.LastSyncResourceVersion()` and use something like
```go
lastKnownVersion := informer.LastSyncResourceVersion()
for informer.LastSyncResourceVersion() == lastKnownVersion {}
// informer has synced at this point
```

2. introduce a new parameter in `AccountCache.Run` that will take a `func()` type and will execute the function at each call
```go
func (c *AccountCache) Run(stopCh <-chan struct{}, fn func()) {
	c.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			defer fn()
			if resourceMatchACKRoleAccountsConfigMap(obj) {
				//...
			}
		},
		// ....
	})
	go c.informer.Run(stopCh)
}
```

In the test file we can use it like this:
```go
accountCache := ackrtcache.NewAccountCache(k8sClient, fakeLogger)
stopCh := make(chan struct{})
eventCh := make(chan struct{})
accountCache.Run(stopCh, func() { eventCh <- struct{}{} })
k8sClient.CoreV1().ConfigMaps(testNamespace).Create(...)
<- eventCh // instead of time.Sleep(time.Second)
```

*NOTE*: i don't feel that this is the right approach

3. introduce a new parameter that will take am optional list of handlers and append them to the shared informer ResourceEventHandlers
```go
func (c *NamespaceCache) Run(stopCh <-chan struct{}, handlers ...k8scache.ResourceEventHandlerFuncs) {
	c.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
		// default handler
	})
	for _, handler := range handlers {
		c.informer.AddEventHandler(handler)
	}
	go c.informer.Run(stopCh)
}
```

using the same testing method in the previous option, we can append a new `EventResourceHandlerFunc` that will send to `eventCh` a `struct{}{}` value whenever the handlers are called.

Finally, IMHO i believe that putting these tests on the same package name `cache` will simplify a things a lot, at least we won't worry about making changes in the public functions

A-Hilaly: I'm also starting to think if we really need to use `ResourceEventHandlerFuncs`... `SharedInformer` exposes a `GetStore` function that return a ["thread safe" store](https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L132-L140) containing all the resources/methods we'll need to use. 
But anyways testing the cache without `time.Sleep` will require using the `SharedInformer.AddEventHandler` if we go for option .2 or .3

pkg/runtime/service_controller_test.go
```diff
@@ -87,6 +89,12 @@ func TestServiceController(t *testing.T) {
 	sc := ackrt.NewServiceController("bookstore", "bookstore.services.k8s.aws")
 	require.NotNil(sc)
 
+	zapOptions := ctrlrtzap.Options{
+		Development: true,
+		Level:       zapcore.InfoLevel,
+	}
+	fakeLogger := ctrlrtzap.New(ctrlrtzap.UseFlagOptions(&zapOptions))
```
A-Hilaly, line 96: Not sure but maybe it's better to pass a fake `io.Writer` to `zapOptions` to prevent cache/controller from logging during unit test phase

pkg/runtime/service_controller_test.go
```diff
@@ -54,7 +56,7 @@ func (m *fakeManager) AddMetricsExtraHandler(path string, handler http.Handler)
 func (m *fakeManager) AddHealthzCheck(name string, check healthz.Checker) error       { return nil }
 func (m *fakeManager) AddReadyzCheck(name string, check healthz.Checker) error        { return nil }
 func (m *fakeManager) Start(<-chan struct{}) error                                    { return nil }
-func (m *fakeManager) GetConfig() *rest.Config                                        { return nil }
+func (m *fakeManager) GetConfig() *rest.Config                                        { return &rest.Config{} }
```
A-Hilaly, line 59: This is needed to avoid panics during unit test phase

pkg/runtime/cache/account.go
```diff
@@ -0,0 +1,115 @@
...
+	c.informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			if resourceMatchACKRoleAccountsConfigMap(obj) {
+				object := obj.(*corev1.ConfigMap).DeepCopy()
+				c.log.V(1).Info("ack-role-account-map has been created")
+				c.updateAccountRoleData(object.Data)
+				c.log.V(1).Info("cached ack-role-account-map data")
+			}
+		},
+		UpdateFunc: func(old, new interface{}) {
```
jaypipes, line 80: teeny nit: `new` is a keyword in Go and naming variables the same as keywords (something called "variable shadowing" ) is generally discouraged. Might be better to name these `orig` and `desired`?

pkg/runtime/cache/account.go
```diff
@@ -0,0 +1,115 @@
...
+	go c.informer.Run(stopCh)
+}
+
+// GetAccountRoleARN queries the AWS accountID associated Role ARN
+// from the cached CARM configmap. This function is thread safe.
+func (c *AccountCache) GetAccountRoleARN(accountID string) (string, bool) {
+	c.RLock()
+	defer c.RUnlock()
+	roleARN, ok := c.roleARNs[accountID]
+	return roleARN, ok
```
jaypipes, line 107: :+1:

pkg/runtime/cache/account_test.go
jaypipes: We can certainly improve this in the future and remove the sleep calls, but this is perfectly fine for now. Nice work on this, @A-Hilaly

pkg/runtime/reconciler.go
```diff
@@ -60,6 +63,12 @@ func (r *reconciler) BindControllerManager(mgr ctrlrt.Manager) error {
 	if r.rmf == nil {
 		return ackerr.NilResourceManagerFactory
 	}
+	clusterConfig := mgr.GetConfig()
+	kc, err := kubernetes.NewForConfig(clusterConfig)
+	if err != nil {
+		return err
+	}
+	r.cache = ackrtcache.New(kc, r.log)
 	r.kc = mgr.GetClient()
```
jaypipes, line 72: Any reason why you did the above instead of just:

```go
r.kc = mgr.GetClient()
r.cache = ackrtcache.New(r.kc, r.log)
```

pkg/runtime/registry.go
```diff
@@ -39,7 +39,7 @@ func (r *Registry) GetResourceManagerFactories() []types.AWSResourceManagerFacto
 	return res
 }
 
-// RegisterManagerFactory registers a resource manager factory with the
+// RegisterResourceManagerFactory registers a resource manager factory with the
```
jaypipes, line 42: thanks :)

pkg/runtime/service_controller_test.go
jaypipes: Maybe. Can do later :)

jaypipes (approved):
I'm good with this as-is. Awesome work, @A-Hilaly. @mhausenblas and @vijtrip2 would you mind doing a thorough review of this patch before I merge please?

vijtrip2 (approved):
LGTM

mhausenblas (approved):
LGTM as well. Meta comment: we need to review the resource requests (non blocking).

Let's ship it!

pkg/runtime/cache/account.go
A-Hilaly: Ahh, nice catch! I will change that

pkg/runtime/reconciler.go
A-Hilaly: my bad for using `kc` as a variable name ... `clientset` makes much more sense. 
`ackrtcache.New` first parameter takes a `client-go/kubernetes.Interface` , while `mgr.GetClient` returns a `client.Client` (from `sigs.k8s.io/controller-runtime/pkg/client`)

pkg/runtime/cache/account.go
A-Hilaly: Done.

pkg/runtime/reconciler.go
A-Hilaly: Changed `kc` to `clientset`

jaypipes: Sorry, you misunderstood me :) I was saying that the clientset object is already being created on line 71 with:

```go
r.kc = mgr.GetClient()
```

so you don't have to create another one with the call to `mgr.GetConfig() and then `kubernetes.NewForConfig()`. You can just pass the reconciler's `kc` attribute (which is the kubernetes clientset) to your `ackrtcache.New()` function.

jaypipes (changes requested):
See inline, sorry I wasn't clear.

pkg/runtime/reconciler.go
A-Hilaly: Ahhh i see now... i think i tried this approach before but something wasn't compiling

A-Hilaly: i think it's because: `client-go/kubernetes.NewForConfig` returns a `kubernetes.Interface`, `mgr.Client` return a `ctrlrt/client.Client` and `client.Client` doesn't implement the `kubernetes.Interface` interface

jaypipes: hmmm. interesting... OK, soon as I'm done with a customer call I'll look into perhaps a cleaner way to do this.

jaypipes: Just closing this out, but I've asked @A-Hilaly to create an issue with a TODO about this and we're going to merge this as-is.

jaypipes (approved):

PR #258, opened 2020-08-31
changing to zap log

nithu0115:
Issue #, if available: N/A

Description of changes:

Switching to zap core logging instead of `klog`. Part of the motivation for `logr` was the idea that V level is a bad idea - log lines are all INFO or ERROR. But there are real world scenario when we want know more on what is happening and it is required. We should limit to V(1) (Debug logging) and should not go over that level. 

Changes:
1) Added log-level arg which will either take `debug` or `info` log levels 
2) Modified manager.go template to pass in region value to generate create temp credentials to be consistent with config region flag


nithu0115:
(Note to reviewer: Change to templates/pkg/crd_manager.go.tpl will require us to re-generate Service controllers)

A-Hilaly (approved):
/lgtm 👍 

Shall we regenerate the service controllers in this PR ?

pkg/runtime/config.go
```diff
@@ -28,6 +29,7 @@ const (
 	flagEnableDevLogging     = "enable-development-logging"
 	flagAWSAccountID         = "aws-account-id"
 	flagAWSRegion            = "aws-region"
+	flagLogLevel			 = "log-level"
```
A-Hilaly, line 32: nit: please run go fmt :)

pkg/runtime/session.go
```diff
@@ -13,12 +13,16 @@
 
 package runtime
 
-import "github.com/aws/aws-sdk-go/aws/session"
+import (
+	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
```
A-Hilaly, line 19: nit: let's keep internal and external imports separated

pkg/runtime/reconciler.go
```diff
@@ -15,7 +15,6 @@ package runtime
 
 import (
 	"context"
-
```
A-Hilaly, line 18: the new line was used to keep standard imports and external imports separated

A-Hilaly (commented):
leaving few comments inline

nithu0115:
> Shall we regenerate the service controllers in this PR ?

I am happy to re-generate but wanted to check in with @jaypipes to see if we want to introduce any process around this. Fwiw, we should maintain 2 branches (main and dev/staging). This way, whenever we make such breaking changes, we can push to dev and cherry pick to merge into main once we cut new branches.

nithu0115:
> > Shall we regenerate the service controllers in this PR ?
> 
> I am happy to re-generate but wanted to check in with @jaypipes to see if we want to introduce any process around this. Fwiw, we should maintain 2 branches (main and dev/staging). This way, whenever we make such breaking changes, we can push to dev and cherry pick to merge into main once we cut new releases.

Spoke to @jaypipes offline and come to a consensus to re-generate service controllers as this is a minimal change. If the change is breaking one or a big one, we will go ahead and merge the PR first without re-generating service controller(s) and then open new PRs for each controller.

pkg/runtime/reconciler.go
```diff
@@ -98,11 +98,12 @@ func (r *reconciler) reconcile(req ctrlrt.Request) error {
 	acctID := r.getOwnerAccountID(res)
 	region := r.getRegion(res)
 
-	r.log.WithValues(
+	log := r.log.WithValues(
 		"account_id", acctID,
 		"region", region,
 		"kind", r.rd.GroupKind().String(),
 	)
+	log.Info("starting reconcilation")
```
jaypipes, line 106: This should be DEBUG, not INFO :)

pkg/runtime/session.go
```diff
@@ -13,12 +13,17 @@
...
-func NewSession() (*session.Session, error) {
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
+
+func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
 	// NOTE(jaypipes): session.NewSession() is needed for the STS::AssumeRole
 	// stuff we will need to do...
-	sess, err := session.NewSession()
+	sess, err := session.NewSession(aws.NewConfig().WithRegion(string(region)))
```
jaypipes, line 26: This change is not related to logging. Would you mind removing this change (put it into a separate PR please)

services/bookstore/pkg/resource/book/manager.go
```diff
@@ -150,7 +150,7 @@ func newResourceManager(
 	id ackv1alpha1.AWSAccountID,
 	region ackv1alpha1.AWSRegion,
 ) (*resourceManager, error) {
-	sess, err := ackrt.NewSession()
+	sess, err := ackrt.NewSession(region)
```
jaypipes, line 153: ditto.

services/petstore/pkg/resource/pet/manager.go
```diff
@@ -227,7 +227,7 @@ func newResourceManager(
 	id ackv1alpha1.AWSAccountID,
 	region ackv1alpha1.AWSRegion,
 ) (*resourceManager, error) {
-	sess, err := ackrt.NewSession()
+	sess, err := ackrt.NewSession(region)
```
jaypipes, line 230: ditto

services/ecr/pkg/resource/repository/manager.go
```diff
@@ -153,7 +153,7 @@ func newResourceManager(
 	id ackv1alpha1.AWSAccountID,
 	region ackv1alpha1.AWSRegion,
 ) (*resourceManager, error) {
-	sess, err := ackrt.NewSession()
+	sess, err := ackrt.NewSession(region)
```
jaypipes, line 156: ditto.

services/sns/config/controller/deployment.yaml
```diff
@@ -27,11 +27,11 @@ spec:
         - ./bin/controller
         args:
         - --aws-account-id
-        - "$(AWS_ACCOUNT_ID)"
+        - "${AWS_ACCOUNT_ID}"
         - --aws-region
-        - "$(AWS_REGION)"
+        - "${AWS_REGION}"
         - --enable-development-logging
-        - "$(ACK_ENABLE_DEVELOPMENT_LOGGING)"
```
jaypipes, line 34: These changes need to be undone. The parentheses are necessary. The curly braces are incorrect for Docker args.

templates/config/controller/deployment.yaml.tpl
```diff
@@ -27,11 +27,11 @@ spec:
...
         args:
         - --aws-account-id
-        - "$(AWS_ACCOUNT_ID)"
+        - "${AWS_ACCOUNT_ID}"
         - --aws-region
-        - "$(AWS_REGION)"
+        - "${AWS_REGION}"
         - --enable-development-logging
-        - "$(ACK_ENABLE_DEVELOPMENT_LOGGING)"
+        - "${ACK_ENABLE_DEVELOPMENT_LOGGING}"
```
jaypipes, line 34: These changes need to be undone. The parentheses are necessary. The curly braces are incorrect for Docker args.

jaypipes (changes requested):
@nithu0115 this needs to be redone. I support the changes for logging, but the remaining changes either belong in a separate PR (NewSession(region)) or are incorrect (the curly braces instead of parens)

pkg/runtime/reconciler.go
nithu0115: Done :)

templates/config/controller/deployment.yaml.tpl
nithu0115: I wanted to keep this consistent with what we currently have - https://github.com/aws/aws-controllers-k8s/blob/main/services/ecr/config/controller/deployment.yaml#L34. Thus, I made this change.

nithu0115:
> @nithu0115 this needs to be redone. I support the changes for logging, but the remaining changes either belong in a separate PR (NewSession(region)) or are incorrect (the curly braces instead of parens)

Addressed your comments.

jaypipes (approved):
:+1:

PR #271, opened 2020-09-03
Handling ValidationException in HTTP API creation & status patching.

vijtrip2:
Generates Go code that checks for required fields in ReadOne
Input shapes in the sdkFind method.

The generate content now looks like
```
func (rm *resourceManager) sdkFind(
	r *resource,
) (*resource, error) {
         // If all the required fields in the input shape are missing, AWS
         // resource is not created yet. Return NotFound here to indicate
         // to callers that the resource isn't yet created.
         if rm.requiredFieldsMissingFromReadOneInput(r) {
                return nil, ackerr.NotFound
         }
...
}
...
func (rm *resourceManager) requiredFieldsMissingFromReadOneInput(
    r *resource,
) bool {
{{ GoCodeRequiredFieldsFromReadOneInput "r" }}
}
```
Issue #127

### DIFF
--- a/apis/core/v1alpha1/resource_metadata.go
+++ b/apis/core/v1alpha1/resource_metadata.go
@@ -24,7 +24,9 @@ type ResourceMetadata struct {
 	// when it has verified that an "adopted" resource (a resource where the
 	// ARN annotation was set by the Kubernetes user on the CR) exists and
 	// matches the supplied CR's Spec field values.
-	ARN *AWSResourceName `json:"arn"`
+	//TODO(vijat@): Find a better strategy for resources that do not have ARN in CreateOutputResponse
+	// https://github.com/aws/aws-controllers-k8s/issues/270
+	ARN *AWSResourceName `json:"arn,omitempty"`
 	// OwnerAccountID is the AWS Account ID of the account that owns the
 	// backend AWS service API resource.
 	OwnerAccountID *AWSAccountID `json:"ownerAccountID"`
--- a/apis/core/v1alpha1/resource_metadata.go
+++ b/apis/core/v1alpha1/resource_metadata.go
@@ -24,9 +24,7 @@ type ResourceMetadata struct {
 	// when it has verified that an "adopted" resource (a resource where the
 	// ARN annotation was set by the Kubernetes user on the CR) exists and
 	// matches the supplied CR's Spec field values.
-	//TODO(vijat@): Find a better strategy for resources that do not have ARN in CreateOutputResponse
-	// https://github.com/aws/aws-controllers-k8s/issues/270
-	ARN *AWSResourceName `json:"arn,omitempty"`
+	ARN *AWSResourceName `json:"arn"`
 	// OwnerAccountID is the AWS Account ID of the account that owns the
 	// backend AWS service API resource.
 	OwnerAccountID *AWSAccountID `json:"ownerAccountID"`
--- a/pkg/generate/apigwv2_test.go
+++ b/pkg/generate/apigwv2_test.go
@@ -14,6 +14,7 @@
 package generate_test
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -226,4 +227,13 @@ func TestAPIGatewayV2_Route(t *testing.T) {
 	}
 `
 	assert.Equal(expCreateOutput, crd.GoCodeSetOutput(model.OpTypeCreate, "resp", "ko.Status", 1))
+
+	expRequiredStatusFieldsMissingFromReadOneInput := `
+	if r.ko.Status.RouteID == nil  {
+		return true
+	} else {
+		return false
+	}
+`
+	assert.Equal(expRequiredStatusFieldsMissingFromReadOneInput, fmt.Sprintf("\n%s\n", crd.GoCodeRequiredStatusFieldsMissingFromReadOneInput("r.ko", 1)))
 }
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -359,6 +359,36 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
 	return "UNKNOWN"
 }
 
+// Return true if all the required status fields are missing from ReadOneInput. Else, return false
+// Sample Output:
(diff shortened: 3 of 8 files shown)


apis/core/v1alpha1/resource_metadata.go
```diff
@@ -24,7 +24,9 @@ type ResourceMetadata struct {
 	// when it has verified that an "adopted" resource (a resource where the
 	// ARN annotation was set by the Kubernetes user on the CR) exists and
 	// matches the supplied CR's Spec field values.
-	ARN *AWSResourceName `json:"arn"`
+	//TODO(vijat@): Find a better strategy for resources that do not have ARN in CreateOutputResponse
+	// https://github.com/aws/aws-controllers-k8s/issues/270
+	ARN *AWSResourceName `json:"arn,omitempty"`
 	// OwnerAccountID is the AWS Account ID of the account that owns the
 	// backend AWS service API resource.
 	OwnerAccountID *AWSAccountID `json:"ownerAccountID"`
```
jaypipes, line 32: Probably want to omitempty on this one too, but let's handle that in a separate PR.

jaypipes (changes requested):
@vijtrip2 can we have separate PRs , one for the status patching changes and one for the validation code changes? For the status patching PR, we'd want to regenerate the checked in service controllers after making those changes to the templates.

apis/core/v1alpha1/resource_metadata.go
vijtrip2: I thought OwnerAccountId will always be present in the resourceManager but yeah for the worst case we can make it optional too. I don't have any strong argument against it.

pkg/model/crd.go
```diff
@@ -448,6 +466,24 @@ func (r *CRD) GoCodeSetInput(
 	out := "\n"
 	indent := strings.Repeat("\t", indentLevel)
 
+	// During initial creation of aws resource, return NotFoundError if all the required Status fields are missing for ReadOne call.
+	// Status fields will be populated after object creation
+	if OpTypeGet == opType {
+		var requiredKoStatusFields = r.RequiredStatusFieldsForReadOneInput()
+		if len(requiredKoStatusFields) > 0 {
+			allRequiredKoStatusFieldsMissingComment := "// If all the requiredKoStatusFields are missing, AWS resource is not created yet. Return NotFound exception\n"
```
jaypipes, line 474: For this kind of thing where there is a long comment and an if condition on the operation type, I'd prefer to put this kind of thing into the template and not the Go code generator.

In other words, here:

https://github.com/aws/aws-controllers-k8s/blob/1cabf4d5a2a159a68ffe7f44d8a4965fe3034d47/templates/pkg/crd_sdk.go.tpl#L39

You would not need to have an `if` conditional because you're already in the "ReadOne" input shape block.

You could add this into the template:

```go
// If all the required fields in the input shape are missing, AWS
// resource is not created yet. Return NotFound here to indicate
// to callers that the resource isn't yet created.
if rm.requiredFieldsMissingFromReadOneInput(r) {
    return nil, ackerr.NotFound
}
```

and then later, in this section:

https://github.com/aws/aws-controllers-k8s/blob/1cabf4d5a2a159a68ffe7f44d8a4965fe3034d47/templates/pkg/crd_sdk.go.tpl#L103-L113

Add this:

```go
func (rm *resourceManager) requiredFieldsMissingFromReadOneInput(
    r *resource,
) bool {
{{ GoCodeRequiredFieldsFromReadOneInput "r" }}
}
```

And then all you need to do is implement a separate `CRD.GoCodeRequiredFieldsFromReadOneInput` method.

jaypipes (changes requested):
@vijtrip2 see inline for a better description of what I was looking for. Also, needs a unit test added to `pkg/generate/apigw_test.go` :)

pkg/generate/apigwv2_test.go
```diff
@@ -226,4 +227,13 @@ func TestAPIGatewayV2_Route(t *testing.T) {
...
 	assert.Equal(expCreateOutput, crd.GoCodeSetOutput(model.OpTypeCreate, "resp", "ko.Status", 1))
+
+	expRequiredStatusFieldsMissingFromReadOneInput := `
+	if r.ko.Status.RouteID == nil  {
+		return true
+	} else {
+		return false
+	}
+`
+	assert.Equal(expRequiredStatusFieldsMissingFromReadOneInput, fmt.Sprintf("\n%s\n", crd.GoCodeRequiredStatusFieldsMissingFromReadOneInput("r.ko", 1)))
```
jaypipes, line 238: Instead of the above, a cleaner approach might be:

```go
got := crd.GoCodeRequiredStatusFieldsMissingFromReadOneInput("r.ko", 1)
assert.Equal(strings.Trim(expRequiredStatusFieldsMissingFromReadOneInput), got)
```

pkg/model/crd.go
```diff
@@ -359,6 +359,54 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
 	return "UNKNOWN"
 }
 
+// Return true if all the required status fields are missing from ReadOneInput. Else, return false
```
jaypipes, line 362: top tip for Golang...

Godoc expects docstring comments for all targets (methods, functions, types and fields) to be in the form `{{ Target }} blah blah`. So, for this method, the docstring should be like this:

```go
// GoCodeRequiresStatusFieldsMissingFromReadOneInput returns true if all the
// required status fields are missing...
```

This expected format helps godoc generate documentation for various targets and keeps the docstrings consistent.

But, remember that this function doesn't actually return true or false :) It returns a **string** with Go code that returns true or false ;) So, that docstring isn't quite true (pun intended).

templates/pkg/crd_sdk.go.tpl
```diff
@@ -101,6 +108,12 @@ func (rm *resourceManager) sdkFind(
 }
 
 {{- if .CRD.Ops.ReadOne }}
+func (rm *resourceManager) requiredStatusFieldsMissingFromReadOneInput(
```
jaypipes, line 111: no docstring comment for this function? :)

jaypipes (approved):
I've left a few notes for cleanups you can do in a future PR. This looks great for now, thank you @vijtrip2! :)

Issue #229, opened 2020-08-20
Maintenance window updates


nmvk:
**Describe the bug**
Elasticache allows customers to perform some modifications in maintenance window. This can be done by not specifying `--apply-immediately`.

Generated code for the update does not seem to have a way to accept this parameter. So when I requested to update a replication group from ACK the request was accepted and ACK describe showed modification as applied but modification was not applied in Elasticache as it was going to be updated during maintenance window.

**Steps to reproduce**
Create a replication group with t3.micro.

```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: ReplicationGroup
metadata:
  name: test-rg
spec:
    engine: redis 
    replicationGroupID: test-k8
    replicationGroupDescription: test
    automaticFailover: true
    cacheNodeType: cache.t3.micro
    numNodeGroups: 3
    replicasPerNodeGroup: 2
```
Update the instance type to t3.medium.
```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: ReplicationGroup
metadata:
  name: test-rg
spec:
    engine: redis 
    replicationGroupID: test-k8
    replicationGroupDescription: test
    automaticFailover: true
    cacheNodeType: cache.t3.medium
    numNodeGroups: 3
    replicasPerNodeGroup: 2
```
**Expected outcome**
A concise description of what you expected to happen.

**Environment**

* Kubernetes version : 1.17
* Using EKS (yes/no), if so version? Yes
* AWS service targeted : Elasticache

jaypipes:
I had a chat with @nmvk about this on Friday. I think to make things simpler, ACK would probably not want to support the maintenance window feature and instead always use the `--apply-immediately` flag. It's going to be "interesting" to figure out how we handle status polling for both RDS and Elasticache resources already, and having to handle another unique maintenance window technique would be difficult.

nmvk:
We would require a way to overwrite the value of input parameter in this case. My proposal is to allow specifying the default values for an operations parameter in `generator.yaml`

```
resources:
  CacheSubnetGroup:
    exceptions:
      codes:
        404: CacheSubnetGroupNotFoundFault
operations:
  ModifyReplicationGroup:
    default_values:
      ApplyImmediately: true
```

jaypipes:
> We would require a way to overwrite the value of input parameter in this case. My proposal is to allow specifying the default values for an operations parameter in `generator.yaml`
> 
> ```
> resources:
[...]

PR #272, opened 2020-09-03
Implement Override values of Operations

nmvk:
Some operations input parameters are requiredto be overriden. This change would allow them to be configuredfrom generator.yaml. ElastiCache will use this to override SetApplyImmediately while modifying CacheCluster and ReplicationGroup

This would not solve the maintenance window handling, but allows us to disable it for now.

Issue #, if available: https://github.com/aws/aws-controllers-k8s/issues/229

### DIFF
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -30,6 +30,14 @@ type Config struct {
 	Resources map[string]ResourceConfig `json:"resources"`
 	// CRDs to ignore. ACK generator would skip these resources.
 	Ignore IgnoreSpec `json:"ignore"`
+	// Contains generator instructions for individual API operations.
+	Operations map[string]OperationConfig `json:"operations"`
+}
+
+// OperationConfig represents instructions to the ACK code generator to
+// specify the overriding values for API operation parameters
+type OperationConfig struct {
+	OverrideValues map[string]string `json:"override_values"`
 }
 
 // IgnoreSpec represents instructions to the ACK code generator to
@@ -215,6 +223,18 @@ func (c *Config) IsIgnoredShape(shapeName string) bool {
 		return false
 	}
 	return util.InStrings(shapeName, c.Ignore.ShapeNames)
+}
+
+// OverrideValues gives list of member values to override.
+func (c *Config) OverrideValues(operationName string) (map[string]string, bool) {
+	if c == nil {
+		return nil, false
+	}
+	oConfig, ok := c.Operations[operationName]
+	if !ok {
+		return nil, false
+	}
+	return oConfig.OverrideValues, ok
 }
 
 // IsIgnoredResource returns true if Operation Name is configured to be ignored
--- a/pkg/generate/elasticache_test.go
+++ b/pkg/generate/elasticache_test.go
@@ -650,3 +650,84 @@ func TestElasticache_Ignored_Resources(t *testing.T) {
 	crd := getCRDByName("GlobalReplicationGroup", crds)
 	require.Nil(crd)
 }
+
+func TestElasticache_Override_Values(t *testing.T)  {
+	require := require.New(t)
+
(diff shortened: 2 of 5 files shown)


pkg/generate/config/config.go
```diff
@@ -217,6 +224,18 @@ func (c *Config) IsIgnoredShape(shapeName string) bool {
 	return util.InStrings(shapeName, c.Ignore.ShapeNames)
 }
 
+// OverrideValues gives list of member values to override.
+func (c *Config)  OverrideValues(operationName string) (map[string]string, bool) {
```
jaypipes, line 228: there's an extra space in between `*Config)` and `OverrideValues(` :) Please run `go fmt` on your Go code to catch these little things; alternately, enable an auto-go-fmt plugin for your IDE or editor.

pkg/generate/elasticache_test.go
```diff
@@ -650,3 +650,84 @@ func TestElasticache_Ignored_Resources(t *testing.T) {
...
+	if r.ko.Spec.SnapshotWindow != nil {
+		res.SetSnapshotWindow(*r.ko.Spec.SnapshotWindow)
+	}
+	if r.ko.Status.SnapshottingClusterID != nil {
+		res.SetSnapshottingClusterId(*r.ko.Status.SnapshottingClusterID)
+	}
+`
+	assert := assert.New(t)
+	assert.Equal(expected, crd.GoCodeSetInput(model.OpTypeUpdate, "r.ko", "res", 1))
+}
```
jaypipes, line 733: ++ nice unit test.

pkg/model/crd.go
```diff
@@ -494,10 +494,21 @@ func (r *CRD) GoCodeSetInput(
...
+
 	for memberIndex, memberName := range inputShape.MemberNames() {
 		if r.UnpacksAttributesMap() && memberName == "Attributes" {
 			continue
 		}
+
+		if override {
+			value, ok := opConfig[memberName]
+			if ok {
+				out += fmt.Sprintf("%s%s.Set%s(%s)\n", indent, targetVarName, memberName, value)
```
jaypipes, line 507: Hmm, this works for `SetSetImmediately(true)` only because the underlying data type is boolean... But this will blow up for any string type, which would expect `SetField("value")`.

We are going to need to a type switch here that inspects the Shape.Type for the InputShape field in question here and outputs Go code that will handle string types properly.

jaypipes (changes requested):
@nmvk great start on this! Please see inline for reasons why you will need to handle string data types differently :)

pkg/generate/config/config.go
nmvk: Ah very nice catch! Enabled this in IDE thank you!

pkg/model/crd.go
nmvk: Agreed, Added a new switch statement to handle `string`, `integer`, `boolean`.

nmvk:
@jaypipes Can you review again, due to my force push did not mark the comments did not get marked as resolved.

jaypipes (changes requested):
Couple minor renames requested and this is good to go, thank you @nmvk! :)

pkg/generate/config/config.go
```diff
@@ -30,6 +30,14 @@ type Config struct {
 	Resources map[string]ResourceConfig `json:"resources"`
 	// CRDs to ignore. ACK generator would skip these resources.
 	Ignore IgnoreSpec `json:"ignore"`
+	// Contains generator instructions for individual API operations.
+	Operations map[string]OperationsConfig `json:"operations"`
+}
+
+// OperationsConfig represents instructions to the ACK code generator to
+// specify the overriding values for API operation parameters
+type OperationsConfig struct {
```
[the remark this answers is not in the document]
nmvk: Done

jaypipes (approved):
perfecto. :)

PR #285, opened 2020-09-08
Rely on aws-region cfg or region annotation instead of AWS_REGION variable

nithu0115:
Issue #, if available: N/A

Description of changes: This change will help us generate session with `ackv1alpha1.AWSRegion` passed via annotation instead of rely on user to pass AWS_REGION environment variable to generate session. 

Along with this, added:
1) Regional endpoint parameter instead of global endpoint. 
* Here we are going to rely on VPC private links and do not allow "custom" endpoint (not the default PrivateLink override to the regional sts endpoint) 

### DIFF
--- a/pkg/runtime/session.go
+++ b/pkg/runtime/session.go
@@ -13,16 +13,22 @@
 
 package runtime
 
-import "github.com/aws/aws-sdk-go/aws/session"
+import (
+	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
 
-func NewSession() (*session.Session, error) {
-	// NOTE(jaypipes): session.NewSession() is needed for the STS::AssumeRole
-	// stuff we will need to do...
-	sess, err := session.NewSession()
+func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
+	awsCfg := aws.Config{
+		Region:              aws.String(string(region)),
+		STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
+	}
+	sess, err := session.NewSession(&awsCfg)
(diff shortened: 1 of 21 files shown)


pkg/runtime/session.go
```diff
@@ -13,16 +13,19 @@
...
-	// NOTE(jaypipes): session.NewSession() is needed for the STS::AssumeRole
-	// stuff we will need to do...
-	sess, err := session.NewSession()
-	if err != nil {
-		return nil, err
-	}
-	// TODO(jaypipes): Handling all common region endpoint, throttling
-	// configuration, TLS, etc
+func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
+	// NOTE(jaypipes): session.must() is needed for the STS::AssumeRole
```
jaypipes, line 24: why is `session.Must()` required for STS::AssumeRole, versus `session.NewSession()`?

The problem with `Must()` is that it will panic upon errors instead of returning the error "upstream" to the caller...

nithu0115:
Note to reviewers: When I re-generated API Gatewayv2 controller code, because of latest changes to API Gatewayv2 SDK, those new fields are introduced in `apigatewayv2 struct`.

vijtrip2 (approved):

pkg/runtime/session.go
```diff
@@ -13,16 +13,23 @@
...
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
 
-func NewSession() (*session.Session, error) {
-	// NOTE(jaypipes): session.NewSession() is needed for the STS::AssumeRole
-	// stuff we will need to do...
-	sess, err := session.NewSession()
+func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
+	// NOTE(jaypipes): session.must() is needed for the STS::AssumeRole
```
A-Hilaly, line 24: nit: comment not needed anymore

A-Hilaly (approved):
/lgtm 👍

jaypipes (approved):
++ thank you @nithu0115

Issue #287, opened 2020-09-09
missing required fields in GetAttributes Input shape


jaypipes:
Running the SQS integration tests, getting this:

```
2020-09-09T15:40:59.982Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "queue", "request": "default/ack-test-smoke-sqs", "error": "InvalidParameter: 1 validation error(s) found.\n- missing required field, GetQueueAttributesInput.QueueUrl.\n"}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:258
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.6.0/pkg/internal/controller/controller.go:211
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.Until
	/go/pkg/mod/k8s.io/apimachinery@v0.18.6/pkg/util/wait/wait.go:90
```

in the controller logs. This is because the `GetQueueAttributesInput` shape has a `QueueUrl` required field and we aren't supplying any value (for a new Queue resource, we first call `sdkFind` to see if a Queue exists).

Basically, the code that @vijtrip2 wrote for #271 needs to be added to the GetXXXAttributes code paths...

PR #288, opened 2020-09-09
NotFound when any required field missing in shape

jaypipes:
In #271, @vijtrip2 added methods that output Go code that checked if the
Input shape for ReadOne operations for a resource had required fields
and if so, checked that the CR's `Status.{Field}` had a non-nil value,
otherwise the `resourceManager.sdkFind()` method would return
`NotFound`, indicating to callers (in particular when trying to
determine if a newly-created resource already existed) that the backend
AWS resource had not been created yet.

This patch makes that work generic in two ways:

* adds support for when required fields in an Input shape are in the
  CR's `Spec` struct in addition to the `Status` struct.
* adds support for any Input shape, not just the ReadOne operation's
  Input shape.

For SQS, the GetQueueAttributes' Input shape has required fields
(QueueUrl) that, if not present in the CR's `Status` struct, indicate
the resource has not been created yet.

Issue #287

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -359,52 +359,79 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
 	return "UNKNOWN"
 }
 
-// Return true if all the required status fields are missing from ReadOneInput. Else, return false
+// GoCodeRequiredFieldsMissingFromShape returns Go code that contains a
+// condition checking that the required fields in the supplied Shape have a
+// non-nil value in the corresponding CR's Spec or Status substruct.
+//
 // Sample Output:
-//if r.ko.Status.APIID == nil {
-//	return true
-//} else {
-//	return false
-//}
-func (r *CRD) GoCodeRequiredStatusFieldsMissingFromReadOneInput(koVarName string, indentLevel int) string {
-	out := ""
-	indent := strings.Repeat("\t", indentLevel)
-
-	var requiredKoStatusFields = r.RequiredStatusFieldsForReadOneInput()
-	if len(requiredKoStatusFields) > 0 {
-		allRequiredKoStatusFieldMissingCondition := ""
-		for _, fieldName := range requiredKoStatusFields {
-			// Use '&&' because all the requiredStatusFields should be missing if object is not created yet
-			allRequiredKoStatusFieldMissingCondition += fmt.Sprintf("%s.Status.%s == nil &&", koVarName, fieldName.Names.Camel)
-		}
-		allRequiredKoStatusFieldMissingCondition = strings.TrimSuffix(allRequiredKoStatusFieldMissingCondition, "&&")
-		out += fmt.Sprintf("%sif %s {\n", indent, allRequiredKoStatusFieldMissingCondition)
-		out += fmt.Sprintf("%s\treturn true\n", indent)
-		out += fmt.Sprintf("%s} else {\n", indent)
-		out += fmt.Sprintf("%s\treturn false\n", indent)
-		out += fmt.Sprintf("%s}", indent)
-	} else {
-		out += fmt.Sprintf("%sreturn false", indent)
+//
+// return r.ko.Spec.APIID == nil || r.ko.Status.RouteID != nil
+func (r *CRD) GoCodeRequiredFieldsMissingFromShape(
+	opType OpType,
+	koVarName string,
+	indentLevel int,
+) string {
+	var op *awssdkmodel.Operation
+	switch opType {
+	case OpTypeGet:
+		op = r.Ops.ReadOne
+	case OpTypeGetAttributes:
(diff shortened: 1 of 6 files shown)


pkg/model/crd.go
```diff
@@ -359,52 +359,89 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
...
+	indent := strings.Repeat("\t", indentLevel)
+	if shape == nil || len(shape.Required) == 0 {
+		return fmt.Sprintf("%sreturn false", indent)
 	}
-	requiredFieldNames := inputShape.Required
-	for _, requiredFieldName := range requiredFieldNames {
-		koStatusField, found := r.StatusFields[requiredFieldName]
+	out := ""
+
+	// Loop over the required member fields in the shape and identity whether
```
kumargauravsharma, line 407: nitpick: identity => identify

pkg/model/crd.go
```diff
@@ -359,52 +359,89 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
...
+				)
+				panic(msg)
+			}
+			resVarPath = resVarPath + ".Status." + cleanMemberName
 		}
-	}
-	return requiredStatusFields
+		missing = append(missing, fmt.Sprintf("%s == nil", resVarPath))
+	}
+	// Use '||' because if any of the required fFields are missing the object
```
kumargauravsharma, line 436: nitpick: fFields => fields

vijtrip2:
LGTM.. For my understanding can you clarify if the following create flow is expected.

a) If "any" required field is missing (using OR operator ) then ReadOne will not be performed before Create. Ex: ApiId for Route resource
b) In that case if any required-field for CreateShape is missing from spec, User will see ValidationError from createCall in their controller logs Ex: "ApiId Missing" on CreateRoute call.
c) Then ACK user will add required CreateInput field in their spec for remedy.

vijtrip2 (approved):

pkg/model/crd.go
```diff
@@ -359,52 +359,89 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
...
+	// is not created yet
+	missingCondition := strings.Join(missing, " || ")
+	out += fmt.Sprintf("%sif (%s) {\n", indent, missingCondition)
+	out += fmt.Sprintf("%s\treturn true\n", indent)
+	out += fmt.Sprintf("%s} else {\n", indent)
+	out += fmt.Sprintf("%s\treturn false\n", indent)
+	out += fmt.Sprintf("%s}", indent)
```
A-Hilaly, line 443: nit: Maybe we can just return the `missingCondition` "value"
```go
out += fmt.Sprintf("%s\treturn %s\n", indent, missingCondition)
```

test/e2e/sqs/smoke.sh
```diff
@@ -0,0 +1,66 @@
...
+
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+ROOT_DIR="$THIS_DIR/../../.."
+SCRIPTS_DIR="$ROOT_DIR/scripts"
+
+source "$SCRIPTS_DIR/lib/common.sh"
+source "$SCRIPTS_DIR/lib/k8s.sh"
+source "$SCRIPTS_DIR/lib/testutil.sh"
+
+check_is_installed jq
```
A-Hilaly, line 11: `jq` is not called anywhere in this script

A-Hilaly (approved):
/lgtm

Leaving few comments

pkg/model/crd.go
jaypipes: fixed.

pkg/model/crd.go
jaypipes: Sorry, I have a stutter. :P

pkg/model/crd.go
jaypipes: ya, that's a good suggestion, @A-Hilaly.

test/e2e/sqs/smoke.sh
jaypipes: copy/pasta :) removed.

jaypipes:
> a) If "any" required field is missing (using OR operator ) then ReadOne will not be performed before Create. Ex: ApiId for Route resource

Correct.

> b) In that case if any required-field for CreateShape is missing from spec, User will see ValidationError from createCall in their controller logs Ex: "ApiId Missing" on CreateRoute call.

Correct. For now. We will be using validating admission webhooks (see #162)  to handle validation of the new resources before it gets to the main reconciler.sync loop.

> c) Then ACK user will add required CreateInput field in their spec for remedy.

Correct. For now. Again, when the webhooks are done, there will be more direct/immediate feedback to the user from the kube-apiserver (which calls the webhook when `kubectl create/apply` is called.

Issue #282, opened 2020-09-08
Map computed properties from output shape to CRD's status


kumargauravsharma:
**Is your feature request related to a problem?**
While populating CRD's status field for `ReplicationGroup` output shape, the properties like `ReplicasPerNodeGroup` are not populated/computed. It can be computed using `Status.NodeGroups` from output shape.

And as the `GoCodeSetOutput` logic copies the `r.ko` (desired state) into returned value `ko`
```
        // Merge in the information we read from the API call above to the copy of
	// the original Kubernetes object we passed to the function
	ko := r.ko.DeepCopy()
```
the reconciler fails to identify change in such properties when applied using `kubectl apply -f updated_resource.yaml`
and no update call is made.

**Describe the solution you'd like**
Enhance generator.yaml to specify the mapping of output shape properties to CRD's status properties or allowing custom code invocation during `GoCodeSetOutput` to help services implement the custom logic for such scenarios.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

kumargauravsharma:
Possible solution:
Enhance `generator.yaml` to specify custom code invocation during `GoCodeSetOutput `to help services implement the custom logic. This is generic solution which can be used for any service in ACK.

For ElastiCache ReplicationGroup, the updated `generator.yaml` will look like:
```
resources:
  ReplicationGroup:
    custom_set_output_operation:
      GoCodeCustomSetOutput
```
and an implementation for `GoCodeCustomSetOutput` operation will be provided inside `aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/custom_set_output.go` file and it will be invoked from `crd.go::GoCodeSetOutput`.

Thus the fields that are part of resource spec (due to create API input but do not have direct mapping in describe API output) will be properly retrieved from response object and improve the reconciler update code flow to identify changes/diffs in `desired`, `latest` states.

kumargauravsharma:
Considering that different operations (create, read, update) on a custom resource can have different output shapes, the custom set output needs to process output shape; thus updated generator.yaml would look like:
```
resources:
  ReplicationGroup:
    output_shapes:
      ReplicationGroup:
        custom_set_output_operation: CustomSetReplicationGroupOutput
```

This can be read as: For resource `ReplicationGroup` the output shape `ReplicationGroup` has custom set output operation `CustomSetReplicationGroupOutput`.

kumargauravsharma:
Alternative generator.yaml
```
[...]

PR #307, opened 2020-09-15
Enable custom set output operations

kumargauravsharma:
Issue #282 

Description of changes:
Added support for custom set output operations using generator.yaml configuration.
Example:
```
operations:
  DescribeReplicationGroups:
    custom_set_output_operation: CustomDescribeReplicationGroupsSetOutput
```

### DIFF
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -38,6 +38,8 @@ type Config struct {
 // specify the overriding values for API operation parameters
 type OperationConfig struct {
 	OverrideValues map[string]string `json:"override_values"`
+	// CustomSetOutputOperation provides custom operation to process operation's output shape.
+	CustomSetOutputOperation string `json:"custom_set_output_operation,omitempty"`
 }
 
 // IgnoreSpec represents instructions to the ACK code generator to
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -38,8 +38,10 @@ type Config struct {
 // specify the overriding values for API operation parameters
 type OperationConfig struct {
 	OverrideValues map[string]string `json:"override_values"`
-	// CustomSetOutputOperation provides custom operation to process operation's output shape.
-	CustomSetOutputOperation string `json:"custom_set_output_operation,omitempty"`
+	// SetOutputCustomMethodName provides the name of the custom method on the
+	// `resourceManager` struct that will set fields on a `resource` struct
+	// depending on the output of the operation.
+	SetOutputCustomMethodName string `json:"set_output_custom_method_name,omitempty"`
 }
 
 // IgnoreSpec represents instructions to the ACK code generator to
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -304,23 +304,23 @@ func (r *CRD) IsPrimaryARNField(fieldName string) bool {
 		strings.EqualFold(fieldName, r.Names.Original+"arn")
 }
 
-// GetCustomSetOutputOperation returns custom set output operation as string for
+// SetOutputCustomMethodName returns custom set output operation as *string for
 // given operation on custom resource, if specified in generator config
-func (r *CRD) GetCustomSetOutputOperation(
+func (r *CRD) SetOutputCustomMethodName(
 	// The operation to look for the Output shape
 	op *awssdkmodel.Operation,
-) string {
+) *string {
 	if op == nil {
-		return ""
+		return nil
 	}
 	if r.genCfg == nil {
-		return ""
+		return nil
 	}
 	resGenConfig, found := r.genCfg.Operations[op.Name]
 	if !found {
-		return ""
+		return nil
 	}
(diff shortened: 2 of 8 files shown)


pkg/generate/config/config.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 // specify the overriding values for API operation parameters
 type OperationConfig struct {
 	OverrideValues map[string]string `json:"override_values"`
+	// CustomSetOutputOperation provides custom operation to process operation's output shape.
+	CustomSetOutputOperation string `json:"custom_set_output_operation,omitempty"`
```
jaypipes, line 42: Probably just call this `SetOutputCustomMethod` or `SetOutputCustomMethodName` and change the description to say:

```
// SetOutputCustomMethod provides the name of the custom method on the
// `resourceManager` struct that will set fields on a `resource` struct
// depending on the output of the operation.
```

pkg/model/crd.go
```diff
@@ -304,6 +304,25 @@ func (r *CRD) IsPrimaryARNField(fieldName string) bool {
 		strings.EqualFold(fieldName, r.Names.Original+"arn")
 }
 
+// GetCustomSetOutputOperation returns custom set output operation as string for
+// given operation on custom resource, if specified in generator config
+func (r *CRD) GetCustomSetOutputOperation(
```
jaypipes, line 309: You could just call this method `SetOutputCustomMethodName()`

Additionally, consider making the attribute a `*string` instead of `string` and returning `nil` here instead of `""` to indicate there is no custom method name...

templates/pkg/crd_sdk.go.tpl
```diff
@@ -94,6 +94,7 @@ func (rm *resourceManager) sdkFind(
 		return nil, err
 	}
 {{ $setCode := GoCodeSetReadManyOutput .CRD "resp" "ko" 1 }}
+{{ $setCustomOutputOperation := .CRD.GetCustomSetOutputOperation .CRD.Ops.ReadMany }}
```
jaypipes, line 97: any reason to place this here instead of below where it is used?

templates/pkg/crd_sdk.go.tpl
```diff
@@ -180,6 +185,7 @@ func (rm *resourceManager) sdkCreate(
 		return nil, err
 	}
 {{ $createCode := GoCodeSetCreateOutput .CRD "resp" "ko.Status" 1 }}
+{{ $setCustomOutputOperation := .CRD.GetCustomSetOutputOperation .CRD.Ops.Create }}
```
jaypipes, line 188: ditto...

templates/pkg/crd_sdk.go.tpl
```diff
@@ -228,6 +238,7 @@ func (rm *resourceManager) sdkUpdate(
 {{ end }}
 
 {{ $setCode := GoCodeSetUpdateOutput .CRD "resp" "ko.Status" 1 }}
+{{ $setCustomOutputOperation := .CRD.GetCustomSetOutputOperation .CRD.Ops.Update }}
```
jaypipes, line 241: ditto.

jaypipes (changes requested):
Some relatively minor squibbles, overall very nice work, @kumargauravsharma! :)

pkg/generate/config/config.go
kumargauravsharma: Renamed to `SetOutputCustomMethodName` and updated the code comment per the suggestion.

pkg/model/crd.go
kumargauravsharma: Updated the method name.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Moved the code and used `{{ if $setOutputCustomMethodName := ` to combine into one statement.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Moved the code and used `{{ if $setOutputCustomMethodName := ` to combine into one statement.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Moved the code and used `{{ if $setOutputCustomMethodName := ` to combine into one statement.

pkg/generate/config/config.go
```diff
@@ -38,6 +38,10 @@ type Config struct {
 // specify the overriding values for API operation parameters
 type OperationConfig struct {
 	OverrideValues map[string]string `json:"override_values"`
+	// SetOutputCustomMethodName provides the name of the custom method on the
+	// `resourceManager` struct that will set fields on a `resource` struct
+	// depending on the output of the operation.
```
jaypipes, line 43: In a future PR, we should consider documenting here the interface / function signature of the SetOutput custom method.

jaypipes (approved):
Nice work @kumargauravsharma :)

Issue #296, opened 2020-09-11
sns: GENERATION FAILURE! there's a required field PlatformApplicationArn in Shape GetPlatformApplicationAttributesInput that isn't in either the CR's Spec or Status structs


kumargauravsharma:
**Describe the bug**
`make build-controller SERVICE=sns` fails with error:
```
Error: template: pkg/crd_sdk:157:3: executing "pkg/crd_sdk" at <GoCodeRequiredFieldsMissingFromGetAttributesInput .CRD "r.ko" 1>: error calling GoCodeRequiredFieldsMissingFromGetAttributesInput: GENERATION FAILURE! there's a required field PlatformApplicationArn in Shape GetPlatformApplicationAttributesInput that isn't in either the CR's Spec or Status structs!
```

**Steps to reproduce**
1. Checkout commit level: `29360d667ddb3d898623ccb052e6179a0cbc25c9` 
2. Run `make build-controller SERVICE=sns`
3. Observe the error. 
Error Logs:

```
$ make build-controller SERVICE=sns
go build -tags codegen -ldflags "-X main.version="v0.0.0" -X main.buildHash=29360d667ddb3d898623ccb052e6179a0cbc25c9 -X main.buildDate=2020-09-11T09:23:07Z" -o bin/ack-generate cmd/ack-generate/main.go
./scripts/build-controller.sh sns
Building Kubernetes API objects for sns
Generating deepcopy code for sns
Generating custom resource definitions for sns
Building service controller for sns
Error: template: pkg/crd_sdk:157:3: executing "pkg/crd_sdk" at <GoCodeRequiredFieldsMissingFromGetAttributesInput .CRD "r.ko" 1>: error calling GoCodeRequiredFieldsMissingFromGetAttributesInput: GENERATION FAILURE! there's a required field PlatformApplicationArn in Shape GetPlatformApplicationAttributesInput that isn't in either the CR's Spec or Status structs!
Usage:
  ack-generate controller <service> [flags]

Flags:
  -h, --help            help for controller
  -o, --output string   path to root directory to create generated files. Defaults to /$service

Global Flags:
      --cache-dir string               Path to directory to store cached files (including clone'd aws-sdk-go repo) (default "/Users/kumargsh/.cache/ack-generate")
      --dry-run                        If true, outputs all files to stdout
      --generator-config-path string   Path to file containing instructions for code generation to use
[...]

PR #312, opened 2020-09-16
fix issues with Get/SetAttributes Go code generation

jaypipes:
This patch adds to the code generator for the SDK linkage for APIs that
use the SetAttributes-flavor of updating a resource. Previously, the
`sdk.go` files for these APIs had empty `sdkUpdate` operations with a
big TODO that ended up breaking the service controller (quite
predictably).

While implementing this patch, I stumbled upon yet another unexplainable
inconsistency between two SetAttributes operations within the SNS
service API. The SNS `SetPlatformApplicationAttributes` API call accepts a
field in its Input shape called `Attributes` that is a map of key/value
pairs for attributes to set on the platform application. Unfortunately,
the similarly-named SNS `SetTopicAttributes` API call apparently does
*NOT* work the same way. Instead, there is a single `AttributeName` and
`AttributeValue` field in the Input shape and you need to call the
`SetTopicAttributes` API call once for each modified attribute. :(

In fact, the [official documentation][0] for the SNS SetTopicAttributes
API call says that the `AttributeName` field in the Input shape is
actually a "*map* of attributes with their corresponding values". But
that isn't actually the case...

So, the code in this patch for now just contains a TODO in the
`CRD.GoCodeSetAttributesSetInput()` method for SetAttributes APIs that
can only operate on a single attribute at a time. We will use the
CustomOperation functionality as a temporary workaround for these APIs
while we come up with a more permanent code-generated solution.

There were a series of problems that I uncovered when investigating the
cause of Issue #296:

* The code returned from GoCodeGetAttributesForOutput was failing to set
  `Status.ACKResourceMetadata.ARN` and
  `Status.ACKResourceMetadata.OwnerAccountID` when the attribute fields
  corresponded to the primary resource ARN or the owner ID. This was the
  direct cause of Issue #296
* The code in `templates/pkg/crd_sdk.go.tpl` that ran at the end of the
  `sdkCreate()` call was inadvertently overwriting any setters of
  `Status.ACKResourceMetadata` that had previously executed
* The code returned from APIs with GetAttributes operations was *always*
  returning a nil-guard and constructor for the ACKResourceMetadata
  struct, even when no attribute fields actually set either ARN or
  OwnerAccountID. When I removed the check in the template for Status
  fields being required for the `resp` variable to be defined, this was
  causing `"resp" variable unused` compilation failures

Issue #296

### DIFF
--- a/pkg/generate/testdata/models/apis/sns/0000-00-00/generator.yaml
+++ b/pkg/generate/testdata/models/apis/sns/0000-00-00/generator.yaml
@@ -1,6 +1,7 @@
 resources:
   Topic:
     unpack_attributes_map:
+      set_attributes_single_attribute: true
       fields:
         DeliveryPolicy:
         DisplayName:
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -161,6 +161,13 @@ type UnpackAttributesMapConfig struct {
 	// information to the ACK service controller that is useful for determining
 	// observed versus desired state -- then do NOT list that attribute here.
 	Fields map[string]FieldConfig `json:"fields"`
+	// SetAttributesSingleAttribute indicates that the SetAttributes API call
+	// doesn't actually set multiple attributes but rather must be called
+	// multiple times, once for each attribute that needs to change. See SNS
+	// SetTopicAttributes API call, which can be compared to the "normal" SNS
+	// SetPlatformApplicationAttributes API call which accepts multiple
+	// attributes and replaces the supplied attributes map key/values...
+	SetAttributesSingleAttribute bool `json:"set_attributes_single_attribute"`
 }
 
 // FieldConfig contains instructions to the code generator about how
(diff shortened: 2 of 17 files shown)


pkg/generate/testdata/models/apis/sns/0000-00-00/generator.yaml
```diff
@@ -1,6 +1,7 @@
 resources:
   Topic:
     unpack_attributes_map:
+      set_attributes_single_attribute: true
```
kumargauravsharma, line 4: As the 'single attribute' behavior is related to operation on the resource, should the configuration be set at operation level in the config? The resultant generator.yaml would look like:
```
operations:
    SetTopicAttributes:
         single_attributes: true
```

jaypipes: Yeah, it's an interesting question. I did it this way because this is more of a variation on the "UnpacksAttributesMap" behaviour than it is for anything else. That said, I'm going to be using the OperationConfig.CustomOperation functionality to sidestep the weirdness in the SetTopicAttributes API call entirely for now, so the point may be moot (for the timebeing at least)

kumargauravsharma (approved):

A-Hilaly (approved):
/lgtm 👍 
Was fun reading the comments :)

PR #318, opened 2020-09-19
update replication group - modify shard configuration and replica count

kumargauravsharma:
Issue #210 

Description of changes:
Implementation for modify replication group shard configuration and replica count

### DIFF
--- a/services/elasticache/pkg/resource/replication_group/custom_set_output.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_set_output.go
@@ -55,11 +55,4 @@ func (rm *resourceManager) customSetOutput(
 	ko *svcapitypes.ReplicationGroup,
 ) {
 	// TODO: custom code
-	if ko.Status.NodeGroups != nil {
-		for _, nodegroup := range ko.Status.NodeGroups {
-			membersCount := int64(len(nodegroup.NodeGroupMembers))
-			ko.Spec.ReplicasPerNodeGroup = &membersCount
-			break
-		}
-	}
 }
--- a/services/elasticache/pkg/resource/replication_group/custom_update_api.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_update_api.go
@@ -16,11 +16,11 @@ package replication_group
 import (
 	"context"
 	"fmt"
+
 	ackcompare "github.com/aws/aws-controllers-k8s/pkg/compare"
 	svcapitypes "github.com/aws/aws-controllers-k8s/services/elasticache/apis/v1alpha1"
 	svcsdk "github.com/aws/aws-sdk-go/service/elasticache"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"strconv"
 )
 
 // Implements specialized logic for replication group updates.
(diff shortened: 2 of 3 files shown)


services/elasticache/pkg/resource/replication_group/custom_set_output.go
```diff
@@ -55,11 +55,4 @@ func (rm *resourceManager) customSetOutput(
 	ko *svcapitypes.ReplicationGroup,
 ) {
 	// TODO: custom code
-	if ko.Status.NodeGroups != nil {
-		for _, nodegroup := range ko.Status.NodeGroups {
-			membersCount := int64(len(nodegroup.NodeGroupMembers))
-			ko.Spec.ReplicasPerNodeGroup = &membersCount
-			break
-		}
-	}
```
jaypipes, line 64: ++ I presume you removed this because it's not appropriate to have controller code modifying the `Spec` attributes (only the `Status` attributes)

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -30,76 +30,146 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
+			if desiredReplicaCount := int(*desiredSpec.ReplicasPerNodeGroup); desiredReplicaCount != latestReplicaCount {
+				nodeGroupID := ""
+				if latestShard.NodeGroupID != nil {
+					nodeGroupID = *latestShard.NodeGroupID
+				}
+				return desiredReplicaCount - latestReplicaCount, fmt.Sprintf(
+					"ReplicasPerNodeGroup differs for NodeGroup: %s, desired = %d, latest: %d",
+					nodeGroupID, int(*desiredSpec.ReplicasPerNodeGroup), latestReplicaCount)
+			}
+		}
```
jaypipes, line 84: might be nice to break the above lines 68-84 into its own `diffReplicasPerNodeGroup` function and lines 86-115 into its own `diffReplicasNodeGroupConfiguration` function that you can then call:

```go
if desiredSpec.ReplicasPerNodeGroup != nil {
    return rm.diffReplicasPerNodeGroup(desired, latest)
} else if desiredSpec.NodeGroupConfiguration != nil {
    return rm.diffReplicasNodeGroupConfiguration(desired, latest)
}
return 0
```

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -30,76 +30,146 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
+// -ve number if desired replica count is less than latest replica count
+// 0 otherwise
+func (rm *resourceManager) replicaCountDifference(
+	desired *resource,
+	latest *resource,
+) (int, string) {
+	desiredSpec := desired.ko.Spec
+	latestStatus := latest.ko.Status
+
+	if desiredSpec.ReplicasPerNodeGroup != nil {
```
jaypipes, line 67: Please place a comment into this function that basically says that there are two ways of setting replica counts for NodeGroups in Elasticache ReplicationGroups. The first way is to have the same replica count for all node groups. In this case, the `Spec.ReplicasPerNodeGroup` field is set to a non-nil-value integer pointer. The second way is to set different replica counts per node group. In this case, the `Spec.NodeGroupConfiguration` field is set to a non-nil `NodeGroupConfiguration` struct which contains a `NodeGroups` slice of `NodeGroup` structs that each have a `ReplicatCount` non-nil-value integer pointer field that contains the number of replicas for that particular node group.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -30,76 +30,146 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 	latest *resource,
 	diffReporter *ackcompare.Reporter,
```
jaypipes, line 31: So the diffReporter variable is no longer being used, right?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -110,144 +180,184 @@ func (rm *resourceManager) updateReplicaCount(
...
-	// TODO: 	optional - only if -  NumNodeGroups decreases shards
-	//			refer 'rl' to find out
-	// res.SetNodeGroupsToRemove() or res.SetNodeGroupsToRetain()
-	res.SetNodeGroupsToRetain(nodegroupsToRetain)
-
 	return res, nil
 }
 
-// new(IncreaseReplicaCount)RequestPayload returns an SDK-specific struct for the HTTP request
+// new(DecreaseReplicaCount)RequestPayload returns an SDK-specific struct for the HTTP request
```
jaypipes, line 245: Can you remove the parens around DecreaseReplicaCount above?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -110,144 +180,184 @@ func (rm *resourceManager) updateReplicaCount(
...
 	return res, nil
 }
 
-// new(DecreaseReplicaCount)RequestPayload returns an SDK-specific struct for the HTTP request
-// payload of the Create API call for the resource
-func (rm *resourceManager) newDecreaseReplicaCountRequestPayload(
-	r *resource,
-) (*svcsdk.DecreaseReplicaCountInput, error) {
-	res := &svcsdk.DecreaseReplicaCountInput{}
+// newUpdate(ShardConfiguration)RequestPayload returns an SDK-specific struct for the HTTP request
```
jaypipes, line 291: Can you remove the parens around ShardConfiguration above?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -110,144 +180,184 @@ func (rm *resourceManager) updateReplicaCount(
...
+		return nil, err
+	}
+	resp, respErr := rm.sdkapi.ModifyReplicationGroupShardConfigurationWithContext(ctx, input)
+	if respErr != nil {
+		return nil, respErr
+	}
+	return provideUpdatedResource(desired, resp.ReplicationGroup)
+}
+
+// new(IncreaseReplicaCount)RequestPayload returns an SDK-specific struct for the HTTP request
```
jaypipes, line 199: Can you remove the parens around IncreaseReplicaCount above?

services/elasticache/pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -0,0 +1,713 @@
...
+					found = true
+					break
+				}
+			}
+			assert.True(found, "Expected node group id %s not found in payload", desiredNodeGroup.NodeGroupID)
+		}
+		assert.Nil(payload.NodeGroupsToRemove)
+		assert.Nil(err)
+	})
+}
```
jaypipes, line 713: Nice unit tests, thanks @kumargauravsharma! :)

jaypipes (changes requested):
couple relatively minor cleanups requested inline. Knock those out and we'll merge this today.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -30,76 +30,146 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
 	}
+
+	// no updates
 	return nil, nil
 }
 
-func (rm *resourceManager) updateShardConfiguration(
+// replicaConfigurationsDifference returns
+// +ve number if desired replica count is greater than latest replica count
+// -ve number if desired replica count is less than latest replica count
```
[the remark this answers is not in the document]
kumargauravsharma: updated the code files.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -30,76 +30,146 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
-		case "Spec.ReplicasPerNodeGroup":
-			return rm.updateReplicaCount(ctx, desired, latest, diffReporter)
+	// Order of operations when diffs map to multiple updates APIs:
+	// 1. updateReplicaCount() is invoked Before updateShardConfiguration()
+	//	  because both accept availability zones, however the number of
+	//	  values depend on replica count.
+
+	// increase/decrease replica count
+	if diff, message := rm.replicaCountDifference(desired, latest); diff != 0 {
+		fmt.Print(message)
```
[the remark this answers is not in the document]
kumargauravsharma: Updated function to return int, kept printf statements in the called methods to help in debugging. Once logger is available inside `rm` log level would be applied.

services/elasticache/pkg/resource/replication_group/custom_set_output.go
kumargauravsharma: yes, as both `desired` and `latest` resources are available in update method call, there was no need to update ` Spec`  on latest.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Thanks for the suggestion. Updated source code.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Thanks for the suggestion. Updated source code and added code comment.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: As both `latest` and `desired` objects are available, diffReporter is not required.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated source file.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated source file.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated source file.

jaypipes (approved):
Rock on @kumargauravsharma :)

Issue #207, opened 2020-08-18
APIGatewayv2 service controller


tabern:
## New ACK service controller

Support for Amazon API Gateway v2.

### List of API resources

List the API resources in order of importance to you:

* [API](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis.html#CreateApi)
* [VPCLink](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/vpclinks.html#CreateVpcLink)
* [APIMapping](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/domainnames-domainname-apimappings.html#CreateApiMapping)
* [Authorizer](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-authorizers.html#CreateAuthorizer)
* [Deployment](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-deployments.html#CreateDeployment)
* [DomainName](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/domainnames.html#CreateDomainName)
* [Integration](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-integrations.html#CreateIntegration)
* [IntegrationResponse](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-integrations-integrationid-integrationresponses.html#CreateIntegrationResponse)
* [Model](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-models.html#CreateModel)
* [Route](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-routes.html#CreateRoute)
* [RouteResponse](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-routes-routeid-routeresponses.html#CreateRouteResponse)
* [Stage](https://docs.aws.amazon.com/apigatewayv2/latest/api-reference/apis-apiid-stages.html#CreateStage)

wszarmach-koia:
Could someone tell me whether there are plans to consider the controller as generally available any time soon ?

vijtrip2:
> Could someone tell me whether there are plans to consider the controller as generally available any time soon ?

Hi @wszarmach-koia , thanks for reaching out! We definitely want to graduate all our service controllers to GA in future. 
[...]

PR #331, opened 2020-09-24
apigwv2 developer-preview

vijtrip2:
"Updating Spec and Status from ReadOne & ReadMany calls, using shapes used to construct Spec and Status."

Issue #207

Description of changes:
* modified all tests
* Added a todo to consider renamed spec fields
* generated new controller code for all services
* tested that updates are working now for apigatewayv2
* updated rds test to ignore DBSecurityGroup field because the shape for same field name is different in Create, ReadMany outputs.
* added an e2e test for apigatewayv2
* changed apigatewayv2 status to developer preview

```
vijat@186590cff3cf aws-controllers-k8s % ./scripts/kind-build-test.sh -s apigatewayv2 -p -r arn:aws:iam::309117047740:role/Admin-k8s -c /Users/vijat/Documents/gocode/src/github.com/vijtrip2/aws-controllers-k8s/scripts/../build/tmp-test-a9ac43f6
Building apigatewayv2 docker image
Building 'apigatewayv2' controller docker image with tag: ack-apigatewayv2-controller:10a36b4-dirty
sha256:82e945028ac74112140f14a57b7a9589dcf4d2b20b684c78af354c32fb2a8859
Loading the images into the cluster
Loading CRD manifests for apigatewayv2 into the cluster
customresourcedefinition.apiextensions.k8s.io/apimappings.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/apis.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/authorizers.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/deployments.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/domainnames.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/integrationresponses.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/integrations.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/models.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/routeresponses.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/routes.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/stages.apigatewayv2.services.k8s.aws configured
customresourcedefinition.apiextensions.k8s.io/vpclinks.apigatewayv2.services.k8s.aws configured
Loading RBAC manifests for apigatewayv2 into the cluster
clusterrole.rbac.authorization.k8s.io/ack-controller configured
clusterrolebinding.rbac.authorization.k8s.io/ack-controller-rolebinding unchanged
Loading service controller Deployment for apigatewayv2 into the cluster
namespace/ack-system unchanged
deployment.apps/ack-apigatewayv2-controller configured
Running aws sts assume-role --role-arn arn:aws:iam::309117047740:role/Admin-k8s, --role-session-name tmp-role-0a63a1c2  --duration-seconds 900,
Temporary credentials generated
deployment.apps/ack-apigatewayv2-controller env updated
Added AWS Credentials to env vars map
======================================================================================================
To poke around your test manually:
export KUBECONFIG=/Users/vijat/Documents/gocode/src/github.com/vijtrip2/aws-controllers-k8s/scripts/../build/tmp-test-a9ac43f6/kubeconfig
kubectl get pods -A
======================================================================================================
Successful apigatewayv2 e2e test
e2e took 278 second(s)
To resume test with the same cluster use: "-c /Users/vijat/Documents/gocode/src/github.com/vijtrip2/aws-controllers-k8s/scripts/../build/tmp-test-a9ac43f6"
```


docs/contents/services.md
```diff
@@ -4,7 +4,7 @@ The following services are currently supported by ACK and for details see the [S
 
 |AWS Service|Status|Controller|
 |-----------|------|----------|
-|Amazon [API Gateway V2](https://aws.amazon.com/api-gateway/)|`BUILD`|`apigateway`|
+|Amazon [API Gateway V2](https://aws.amazon.com/api-gateway/)|`DEVELOPER PREVIEW`|`apigateway`|
```
jaypipes, line 7: please "linkify" the controller name like the ECR and S3 ones

pkg/generate/apigwv2_test.go
```diff
@@ -222,11 +222,64 @@ func TestAPIGatewayV2_Route(t *testing.T) {
...
 	}
+	if resp.RouteKey != nil {
+		ko.Spec.RouteKey = resp.RouteKey
+	}
+	if resp.RouteResponseSelectionExpression != nil {
+		ko.Spec.RouteResponseSelectionExpression = resp.RouteResponseSelectionExpression
+	}
+	if resp.Target != nil {
+		ko.Spec.Target = resp.Target
+	}
```
jaypipes, line 280: OK, so what I'm realizing from the above is that you are setting the `Spec` fields -- presumably to a **default** value that the API Gateway v2 service sets on the `Route` resource? If that is the case, we need to think carefully here about whether we might be inadvertently overwriting field values that were set in the desired `Spec` by the Kubernetes user.

Put another way, if `ko.Spec.Target` was set to some non-nil value already by the Kubernetes user in their CR, then the above Go code will overwrite the `Target` field with whatever was returned in the `CreateRoute` response. There's no way to tell if the values were different.

Potentially we could place some guard expression like this?

```go
        if resp.Target != nil && ko.Spec.Target == nil {
		ko.Spec.Target = resp.Target
	}
```

jaypipes (commented):
@vijtrip2 I'm a little concerned about this one... see my comment inline about the danger of overwriting desired CR state with the value returned in a Create response payload.

pkg/generate/apigwv2_test.go
vijtrip2: That's fair. I will work on it.
* I am thinking of avoiding writing spec fields for Create
* I will write spec fields for latestKO when we are comparing desiredKO and latestKO objects in case of Update.

Does the rest of PR look okay to you except handling the CreateResponse??

docs/contents/services.md
vijtrip2: I will. Thanks!

vijtrip2:
(a) I did the easiest/simplest thing to handle the concern mentioned by you. I added a boolean parameter 'performSpecUpdate' to GoCodeSetOutput and control that only Read operation update the spec fields from template.

Above works because I know we are using readOperation to create latest Image in reconciler and not actually to change the desired object.

(b) Another option is making this change top-down and changing interface of resourceManager to  something like
latest, err = rm.Create(ctx, desired, performSpecUpdate=false)
latest, err := rm.ReadOne(ctx, desired, performSpecUpdate=true)

Please let me know if (a) works(in the PR) or I should go for (b).
If you have any other suggestions, I am all ears too.

jaypipes (commented):
I'm a little concerned about this and will need to test this patch out locally and regenerate the existing service controllers to ensure that the sdk.go linkage didn't break (see my question inline about the target var changing to the source var)

vijtrip2:
>"I'm a little concerned about this and will need to test this patch out locally and regenerate the existing service controllers to ensure that the sdk.go linkage didn't break (see my question inline about the target var changing to the source var)"

I did regenerate all the existing service controllers and build them that there are no compilation errors.

pkg/model/crd.go
```diff
@@ -2173,23 +2202,24 @@ func (r *CRD) goCodeSetOutputForContainer(
 			//		f0elem := string{}
 			out += r.goCodeVarEmptyConstructorK8sType(
 				valVarName,
-				shape.ValueRef.Shape,
+				targetShape.ValueRef.Shape,
```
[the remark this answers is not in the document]
vijtrip2: Because targetShape is OutputShape of CreateInput(spec) or CreateOutput(status) so we should be creating constructors based on targetShape.

--------------
The issue was surfaced during creation of APIGatewayV2 Integration sdk.go file
Integration has a field called "TlsConfig".
The outputShape of ReadOne has the name "TlsConfig", while the inputShape of Create object (spec) has the field called "TlsConfigInput"

If we use the sourceShape to create constructor, i.e. TlsConfig, we get compile time errors that TlsConfig does not exist in ko.Spec

----------------

My approach here is to
a) Introduce targetShape alongside sourceShape in every method. TargetShape is either spec or status shape. while sourceshape can also be readOne or readMany shape.
b) If any member of sourceShape member does not exist in targetShape, ignore it.
c) Perform constructor creation based on targetShape and not source Shape.

vijtrip2:
> > > "I'm a little concerned about this and will need to test this patch out locally and regenerate the existing service controllers to ensure that the sdk.go linkage didn't break (see my question inline about the target var changing to the source var)"
> > 
> > 
> > I did regenerate all the existing service controllers and build them that there are no compilation errors.
> 
> I'm concerned about the change of a sourceShape into a targetShape in one of the generator routines.

replied inline for that comment.

jaypipes (approved):
Walked through my questions with @vijtrip2 on a Chime call. I think he uncovered a latent bug with the target/source var name thing, and fixed it here :)

Issue #321, opened 2020-09-21
`ack-generate` git pull the aws-sdk-go version mentioned in `go.mod` file  


A-Hilaly:
Currently `ack-generate` execute a `git pull` on the cached `aws-sdk-go` repository before starting the code generation. This can cause some issues if the version of `aws-sdk-go` mentioned in the `go.mod` file is different from the cached repository version.

Ideally we should parse and pull the version mentioned in the `go.mod` file: [example](https://github.com/aws/aws-controllers-k8s/blob/main/go.mod#L6)

A-Hilaly:
The library Go uses to parse `go.mod` and `go.sum` files: https://github.com/golang/mod/tree/master/modfile.

PR #333, opened 2020-09-25
Use the same aws-sdk-go version mentionned in the go.mod file to generate code

A-Hilaly:
Issue #321

Description of changes:
- Add `getSDKVersion` function to extract the aws-sdk-go version to use
- Remove `--depth=1` flag from the `git pull` command to allow easier switches between versions/tags
- Use `git checkout` instead of `git pull`

### DIFF
--- a/cmd/ack-generate/command/common.go
+++ b/cmd/ack-generate/command/common.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"golang.org/x/mod/modfile"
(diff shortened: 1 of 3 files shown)


vijtrip2 (approved):
LGTM.

A-Hilaly (commented):
Note to reviewers: Users will need to remove the old cache, because `git checkout` doesn't work as expected after a `git clone --depth=1` . 
I'm thinking about adding some logic to handle this specific case. Ideally delete the cache if number of commits is 1...

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.5.1
 	go.uber.org/zap v1.10.0
+	golang.org/x/mod v0.2.0
```
jaypipes, line 20: Please update the `ATTRIBUTION.md` file to include the BSD3 license attribution for this new dependency:

https://pkg.go.dev/golang.org/x/mod/modfile?tab=licenses

jaypipes (changes requested):
Code looks great, thanks @A-Hilaly! Need to update the ATTRIBUTION.md file though. :)

jaypipes:
> Note to reviewers: Users will need to remove the old cache, because `git checkout` doesn't work as expected after a `git clone --depth=1` .
> I'm thinking about adding some logic to handle this specific case. Ideally delete the cache if number of commits is 1...

++

go.mod
A-Hilaly: Nice catch! Done.

cmd/ack-generate/command/common.go
```diff
@@ -74,18 +79,76 @@ func ensureSDKRepo(cacheDir string) error {
...
+		cmd := exec.Command("git", "clone", "-b", sdkVersion, sdkRepoURL, clonePath)
 		return clonePath, cmd.Run()
 	}
 	return clonePath, nil
 }
+
+// removeOldRepoCache deletes old aws-sdk-go repositories. It deletes a
+// repository if the number of commits is equal to 0
+func removeOldRepoCache(repoDir string) error {
+	b, err := exec.Command("git", "-C", repoDir, "rev-list", "--count", "HEAD").Output()
```
jaypipes, line 120: heh, very cool. :)

jaypipes (approved):
Nice work on this Amine! :)

PR #334, opened 2020-09-25
An extra space after EOF should be removed

iaoiui:
heredoc cannot end with extra space after EOF.
So, how about removing this extra space?


jaypipes (approved):
Thanks for catch this @iaoiui :)

PR #336, opened 2020-09-28
CARM implementation

A-Hilaly:
Issue #102 

Description of changes:

- [x] CARM implementation
- [x] Smoke tests
- [ ] Github action step

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -70,6 +70,7 @@ func (r *reconciler) BindControllerManager(mgr ctrlrt.Manager) error {
 	}
 	r.kc = mgr.GetClient()
 	r.cache = ackrtcache.New(clientset, r.log)
+	r.cache.Run()
 	rd := r.rmf.ResourceDescriptor()
 	return ctrlrt.NewControllerManagedBy(
 		mgr,
@@ -106,14 +107,20 @@ func (r *reconciler) reconcile(req ctrlrt.Request) error {
 
 	acctID := r.getOwnerAccountID(res)
 	region := r.getRegion(res)
+	roleARN := r.getRoleARN(acctID)
+	sess, err := NewSession(region, roleARN)
+	if err != nil {
+		return err
+	}
 
 	r.log.WithValues(
-		"account_id", acctID,
+		"account", acctID,
+		"role", roleARN,
 		"region", region,
 		"kind", r.rd.GroupKind().String(),
 	).V(1).Info("starting reconcilation")
 
-	rm, err := r.rmf.ManagerFor(r, acctID, region)
+	rm, err := r.rmf.ManagerFor(r, sess, acctID, region)
 	if err != nil {
 		return err
 	}
@@ -336,10 +343,32 @@ func (r *reconciler) getOwnerAccountID(
 	if acctID != nil {
 		return *acctID
 	}
-	// OK, it's a new resource. Look for an override account ID annotation,
-	// which indicates a cross-account resource request
-	// TODO(jaypipes)
+
+	// look for owner account id in the CR annotations
+	annotations := res.MetaObject().GetAnnotations()
+	accID, ok := annotations[ackv1alpha1.AnnotationOwnerAccountID]
+	if ok {
+		return ackv1alpha1.AWSAccountID(accID)
+	}
+
+	// look for owner account id in the namespace annotations
+	namespace := res.MetaObject().GetNamespace()
+	accID, ok = r.cache.Namespaces.GetOwnerAccountID(namespace)
+	if ok {
+		return ackv1alpha1.AWSAccountID(accID)
+	}
+
+	// use controller configuration
 	return ackv1alpha1.AWSAccountID(r.cfg.AccountID)
+}
+
+// getRoleARN return the Role ARN that should be assumed in order to manage
+// the resources.
+func (r *reconciler) getRoleARN(
+	acctID ackv1alpha1.AWSAccountID,
+) ackv1alpha1.AWSResourceName {
+	roleARN, _ := r.cache.Accounts.GetAccountRoleARN(string(acctID))
+	return ackv1alpha1.AWSResourceName(roleARN)
 }
 
 // getRegion returns the AWS region that the given resource is in or should be
--- a/pkg/runtime/session.go
+++ b/pkg/runtime/session.go
@@ -16,11 +16,19 @@ package runtime
 import (
 	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 )
 
-func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
+// NewSession returns a new session object. Buy default the returned session is
+// created using pod IRSA environment variables. If assumeRoleARN is not empty,
+// NewSession will call STS::AssumeRole and use the returned credentials to create
+// the session.
+func NewSession(
+	region ackv1alpha1.AWSRegion,
+	assumeRoleARN ackv1alpha1.AWSResourceName,
+) (*session.Session, error) {
 	awsCfg := aws.Config{
 		Region:              aws.String(string(region)),
 		STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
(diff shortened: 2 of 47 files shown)


test/e2e/cross-account/smoke.sh
```diff
@@ -0,0 +1,79 @@
...
+SCRIPTS_DIR="$ROOT_DIR/scripts"
+
+source "$SCRIPTS_DIR/lib/common.sh"
+source "$SCRIPTS_DIR/lib/k8s.sh"
+source "$SCRIPTS_DIR/lib/testutil.sh"
+
+
+# AWS_ACCOUNT_ID_2 should be configured to allow cross account resources management
+# from AWS_ACCOUNT_ID
+AWS_ACCOUNT_ID_2=""
```
A-Hilaly, line 14: I'm still working on this script. I need to figure out a way to switch accounts for github action steps

apis/core/v1alpha1/common.go
```diff
@@ -21,3 +21,6 @@ type AWSAccountID string
 
 // AWSResourceName represents an AWS Resource Name (ARN)
 type AWSResourceName string
+
+// AWSRoleARN represents an AWS Role ARN
+type AWSRoleARN string
```
jaypipes, line 26: Just use the AWSResourceName as the data type, please.

pkg/runtime/session.go
```diff
@@ -16,11 +16,19 @@ package runtime
...
 )
 
-func NewSession(region ackv1alpha1.AWSRegion) (*session.Session, error) {
+// NewSession returns a new session object. Buy default the returned session is
+// created using pod IRSA environment variables. If assumeRoleARN is not empty,
+// NewSession will call STS::AssumeRole and use the returned credentials to create
+// the session.
+func NewSession(
+	region ackv1alpha1.AWSRegion,
+	assumeRoleARN ackv1alpha1.AWSRoleARN,
```
jaypipes, line 30: just use `AWSResouceName` data type.

pkg/types/aws_resource_manager.go
```diff
@@ -45,7 +48,12 @@ type AWSResourceManager interface {
 	// higher-level reonciler determines whether or not the desired differs
 	// from the latest observed and decides whether to call the resource
 	// manager's Update method
-	Update(context.Context, /*desired*/ AWSResource, /*latest*/ AWSResource, *ackcompare.Reporter) (AWSResource, error)
+	Update(
+		context.Context,
+		AWSResource, // desired
+		AWSResource, // latest
+		*ackcompare.Reporter,
+	) (AWSResource, error)
```
jaypipes, line 56: FYI, it's better to leave style-only changes like this to a separate style-only PR. Makes reviewing the code easier since it separates functional code changes from non-functional code changes.

services/s3/config/crd/bases/s3.services.k8s.aws_buckets.yaml
```diff
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.1-0.20200716001835-4a903ddb7005
+    controller-gen.kubebuilder.io/version: (devel)
```
jaypipes, line 7: hmm.... not sure this is correct.

services/s3/config/crd/bases/s3.services.k8s.aws_buckets.yaml
```diff
@@ -15,122 +15,122 @@ spec:
...
+    subresources:
+      status: {}
 status:
   acceptedNames:
     kind: ""
     plural: ""
-  conditions: null
-  storedVersions: null
+  conditions: []
+  storedVersions: []
```
jaypipes, line 136: all these changes to CRD YAML files should be in a separate PR. If these changes were from an updated kubebuilder, then we should have a separate PR that only updates the service controller generated files due to that upgraded KB version.

test/e2e/cross-account/smoke.sh
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env bash
```
jaypipes, line 1: Place this file in the `test/e2e/s3` directory and call it `cross-account.sh` :) That way, when we run the S3 e2e tests, we'll run the CARM tests.

test/e2e/cross-account/smoke.sh
```diff
@@ -0,0 +1,79 @@
...
+
+source "$SCRIPTS_DIR/lib/common.sh"
+source "$SCRIPTS_DIR/lib/k8s.sh"
+source "$SCRIPTS_DIR/lib/testutil.sh"
+
+
+# AWS_ACCOUNT_ID_2 should be configured to allow cross account resources management
+# from AWS_ACCOUNT_ID
+AWS_ACCOUNT_ID_2=""
+DEFAULT_REGION="eu-west-2"
```
jaypipes, line 15: Note that there is already an `AWS_REGION` environment variable defined in the `kind-build-test.sh` script:

https://github.com/aws/aws-controllers-k8s/blob/b39842d4ce92595478c4e9e6af4a9667b33985e4/scripts/kind-build-test.sh#L20

So, probably best to call this something like `AWS_REGION_ALT`?

test/e2e/cross-account/smoke.sh
jaypipes, line 14: How about calling this env var `AWS_ACCOUNT_ID_ALT`

jaypipes: Note that the `aws` CLI tool supports named profiles for exactly this purpose:

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

A-Hilaly: Oh yeah let's use this!

pkg/types/aws_resource_manager.go
A-Hilaly: Agreed 👍 gofmt was automatically doing weird comment style change... i'll make this change in a different PR

services/s3/config/crd/bases/s3.services.k8s.aws_buckets.yaml
A-Hilaly: My bad, i was used the latest controller-gen version, i just regenerated with `v3.1.0`

services/s3/config/crd/bases/s3.services.k8s.aws_buckets.yaml
A-Hilaly: Yes this change was made by the latest version of `controller-gen`

test/e2e/cross-account/smoke.sh
A-Hilaly: ++

pkg/runtime/reconciler.go
```diff
@@ -106,14 +107,20 @@ func (r *reconciler) reconcile(req ctrlrt.Request) error {
 
 	acctID := r.getOwnerAccountID(res)
 	region := r.getRegion(res)
+	roleARN, _ := r.cache.Accounts.GetAccountRoleARN(string(acctID))
```
jaypipes, line 110: How about we hide the cache implementation further and create a `reconciler.getRoleARN()` function similar to the `getOwnerAccountID` and `getRegion` functions.

pkg/runtime/reconciler.go
```diff
@@ -106,14 +107,20 @@ func (r *reconciler) reconcile(req ctrlrt.Request) error {
...
+	sess, err := NewSession(region, ackv1alpha1.AWSResourceName(roleARN))
+	if err != nil {
+		return err
+	}
 
 	r.log.WithValues(
 		"account_id", acctID,
+		"assumed_role_arn", roleARN,
```
jaypipes, line 118: just call it `role` instead of `assumed_role_arn`. And let's call the `account_id`log message key just `account`.

jaypipes (changes requested):
two tiny requests inline. :)

pkg/runtime/reconciler.go
A-Hilaly: +1 That's a good idea! Done

pkg/runtime/reconciler.go
A-Hilaly: Done!

test/e2e/s3/cross-account.sh
```diff
@@ -0,0 +1,80 @@
...
+
+source "$SCRIPTS_DIR/lib/common.sh"
+source "$SCRIPTS_DIR/lib/k8s.sh"
+source "$SCRIPTS_DIR/lib/testutil.sh"
+
+
+# AWS_ACCOUNT_ID_ALT should be configured to allow cross account resources management
+# from AWS_ACCOUNT_ID
+AWS_ACCOUNT_ID_ALT=${AWS_ACCOUNT_ID_ALT:-""}
+AWS_PROFILE_ALT=${AWS_PROFILE_ALT:-""}
```
jaypipes, line 15: please work with @smuralee to get the self-hosted runners set up with an appropriate alternate profile/account ID for the e2e tests that will run on new PRs. :)

jaypipes (approved):
\o/

test/e2e/s3/cross-account.sh
A-Hilaly: Sure! Will reach out ASAP.

PR #340, opened 2020-09-28
Avoid creating directories if a service is not found in aws-sdk-go cache

A-Hilaly:
Issue N/A

Description of changes:
- Avoid creating directories if a service is not found in aws-sdk-go cache
- Clarify error message when service is not found in aws-sdk-go cache
- Silence usage when user enter correct flags and arguments but command exit with an error
- Avoid double printing the same error, cobra commands already do when use `RunE`: https://github.com/spf13/cobra/blob/master/command.go#L940-L943


A-Hilaly:
Still need to do the same for `ack-generate controller`

jaypipes (approved):
Nice catch on this, @A-Hilaly :)

PR #341, opened 2020-09-28
allow total override of Update op

jaypipes:
Some APIs, like ECR, have no Update operation at all for some resources
(like Repository). Instead, there are a series of API calls like
`PutImageScanningConfiguration`, `PutImageTagMutability`, or
`SetRepositoryPolicy` that set individual attributes for a particular
resource.

This patch adds the ability to override the implementation of
`sdkUpdate` **entirely** to a custom method name of the resourceManager
struct, thereby allowing total control over the various update
operations. This fixes a bug where, for the ECR sdk.go file, the
sdkUpdate method was empty and returning `(nil, nil)` which was causing
segfaults in the service controller.

### DIFF
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -37,8 +37,8 @@ type Config struct {
 // OperationConfig represents instructions to the ACK code generator to
 // specify the overriding values for API operation parameters and its custom implementation.
 type OperationConfig struct {
-	CustomImplementation string `json:"custom_implementation,omitempty"`
-	OverrideValues map[string]string `json:"override_values"`
+	CustomImplementation string            `json:"custom_implementation,omitempty"`
+	OverrideValues       map[string]string `json:"override_values"`
 	// SetOutputCustomMethodName provides the name of the custom method on the
 	// `resourceManager` struct that will set fields on a `resource` struct
 	// depending on the output of the operation.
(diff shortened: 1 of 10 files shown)


A-Hilaly (approved):
/lgtm

I think it's fair to go this way with custom code.

A-Hilaly:
What about `LifecyclePolicy` and `RepositoryPolicy` specs ? They only exist on `Update` operation inputs. Should we deal with this in a seperate PR ?

jaypipes:
> /lgtm
> 
> I think it's fair to go this way with custom code.

Long term, I'd like to auto-generate this stuff based on a pattern. For now, though, it was more expedient to just do a custom callbak.

jaypipes:
> What about `LifecyclePolicy` and `RepositoryPolicy` specs ? They only exist on `Update` operation inputs. Should we deal with this in a seperate PR ?

Yeah, I still need to do those and add e2e tests...

A-Hilaly:
> Long term, I'd like to auto-generate this stuff based on a pattern. For now, though, it was more expedient to just do a custom callbak.

++

> Note that LifecyclePolicy and RepositoryPolicy is VERY weird. It isn't settable on the Create of the Repository resource.

I see. Yeah i was playing with ECR API this morning.. mega weird x(

jaypipes:
@A-Hilaly @kumargauravsharma @nithu0115 @vijtrip2 OK, this is ready to review now.

A-Hilaly (approved):

mhausenblas:
I'm overall fine with this, only worry I have is the temporary nature (custom callback). These provisional solutions tend to stick around for longer than one anticipates and, if not properly documented, can yield to really nasty prod troubleshooting sessions. So, I suppose what I'm asking is: towards beta/GA, where will we document this in the user docs?

nithu0115 (approved):
/lgtm

kumargauravsharma (approved):

PR #345, opened 2020-09-29
use aws CLI in Docker within e2e tests

jaypipes:
The AWS CLI tool version 1.x does not support a variety of more recent
AWS API calls. Case in point: the ECR DescribeRepositories API call does
not return information about the image scanning configuration.

In order to ensure that we have modern versions of the AWS CLI tool
usable in our e2e tests, I've added a `daws` (for Docker AWS) Bash
function that can be used to execute a specific version of the AWS CLI
tool within a Docker container.


vijtrip2 (approved):

A-Hilaly (approved):
/lgtm

Issue #349, opened 2020-09-29
Wrong CustomResourceDefinition for buckets.s3.services.k8s.aws


ricardo-larosa:
**Describe the bug**
Changes in https://github.com/aws/aws-controllers-k8s/pull/317 are not the same when you run `make build-controller SERVICE=s3` and revert version of `controller-gen.kubebuilder.io/version` from `v0.4.0` (latest) to `v0.3.1-0.20200716001835-4a903ddb7005`
For more context check https://github.com/aws/aws-controllers-k8s/pull/295

**Steps to reproduce**

Run: `make build-controller SERVICE=s3` 

**Expected outcome**
No changes in s3.services.k8s.aws_buckets.yaml

* AWS service targeted 
S3

PR #350, opened 2020-09-29
Support kustomize for S3

ricardo-larosa:
Make possible to install ACK S3 via Kustomize and also fix https://github.com/aws/aws-controllers-k8s/issues/349

Run:
```sh
kustomize build services/s3/config/default
```
Please note it uses latest
```
go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.0
```


jaypipes:
We're going to need to use a specific version of `controller-gen`/`controller-tools` otherwise this is going to keep happening. :(

services/s3/config/crd/bases/s3.services.k8s.aws_buckets.yaml
```diff
@@ -1,10 +1,10 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.3.1-0.20200716001835-4a903ddb7005
+    controller-gen.kubebuilder.io/version: v0.4.0
```
ricardo-larosa, line 7: @jaypipes Do you think it is enough if annotated here?

jaypipes: It's not really about the annotation (that is created by the `controller-gen crd` command). It's that we need to fix ALL the controllers, not just S3 :)

And the root of the fix is in how we install a specific version of `controller-gen`. We're currently using `go get` which has the very unfortunate side-effect of modifying the `go.mod` and `go.sum` file for the source repository, which isn't what we want.

I'll take a stab at trying to fix this in a less instrusive way, @ricardo-larosa.

ricardo-larosa: Cool! For reference https://github.com/kubernetes-sigs/kubebuilder/issues/1351

jaypipes:
@ricardo-larosa I think we're good to close this PR out, yeah? All service controllers have now been updated for controller-gen 0.4.0 and we've ensured in our build scripts that controller-tools@v0.4.0 is installed.

ricardo-larosa:
thanks @jaypipes

Issue #353, opened 2020-09-30
ACK service controllers RBAC needs access to ConfigMap and Namespace resources


jaypipes:
When running e2e tests, seeing the following in the service controller logs:

```
E0930 19:39:50.727439       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:ack-system:default" cannot list resource "namespaces" in API group "" at the cluster scope
E0930 19:39:51.627514       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:ack-system:default" cannot list resource "configmaps" in API group "" in the namespace "ack-system"
```

The RBAC RoleBinding and ClusterRoleBinding that is generated needs to give the service account's Role the ability to read ConfigMap and Namespace resources.

This should be a quick fix. Some lines that instructs the `controller-gen rbac` command to create the role permissions to allow reads from ConfigMap and Namespace resources needs to be added here:

https://github.com/aws/aws-controllers-k8s/blob/377df95cf40b4a1e5a6151811939cdcf690c532b/templates/pkg/crd_manager.go.tpl#L18

Something like this should work:

```go
// +kubebuilder:rbac:resources=namespaces,verbs=get;list;watch
// +kubebuilder:rbac:resources=configmaps,verbs=get;list;watch
```

and then each controller needs to be regenerated of course...

vaibhavp:
@jaypipes I would like to work on it. Can you please assign it to me?

jaypipes:
> @jaypipes I would like to work on it. Can you please assign it to me?

Hi @vaibhavp, I think @ChaturvediSulabh has already begun work on this one :)

If you're interested in contributing, let me find another good first issue for you? Feel free to find me on Kubernetes Slack community on #provider-aws and we can chat.

PR #354, opened 2020-10-01
ACK service controllers RBAC issue fixed

vaibhavp:
Issue #353 

Description of changes:
As per the description, I have updated the file with instruction for the controller-gen RBAC command to create the role permissions to allow reads from ConfigMap and Namespace resources. 

Also regenerated controller for following services. 
- s3
- sns
- apigatewayv2
- elasticache
- ecr

### DIFF
--- a/services/apigatewayv2/pkg/resource/api/manager.go
+++ b/services/apigatewayv2/pkg/resource/api/manager.go
@@ -30,6 +30,8 @@ import (
 
 // +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=apis,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=apis/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch
 
 // resourceManager is responsible for providing a consistent way to perform
 // CRUD operations in a backend AWS service API for Book custom resources.
--- a/test/e2e/s3/cross-account.sh
+++ b/test/e2e/s3/cross-account.sh
@@ -48,7 +48,6 @@ cat <<EOF | kubectl apply -f -
 apiVersion: s3.services.k8s.aws/v1alpha1
 kind: Bucket
 metadata:
-  namespace: testing
   name: $bucket_name
   annotations:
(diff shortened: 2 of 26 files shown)


services/apigatewayv2/pkg/resource/api/manager.go
```diff
@@ -30,6 +30,8 @@ import (
 
 // +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=apis,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=apis/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
```
jaypipes, line 33: ah, good catch on the `groups=""` thing :) I'd forgotten about that!

test/e2e/s3/cross-account.sh
```diff
@@ -48,7 +48,6 @@ cat <<EOF | kubectl apply -f -
 apiVersion: s3.services.k8s.aws/v1alpha1
 kind: Bucket
 metadata:
-  namespace: testing
```
jaypipes, line 51: Hmm, this seems unrelated. Can you comment why this change was necessary?

jaypipes (commented):
Fantastic @vaibhavp! One little question inline for you, otherwise this looks ready to merge.

test/e2e/s3/cross-account.sh
vaibhavp: When we run the test for s3 service, at that time this resource look for testing namespace which is not present in out testing or we are not creating any such namespace in our testing script anywhere due to which test case for s3  was falling. SO fix this is I have removed this line so that Bucket resource can be created in default namespace or select name.

jaypipes: Ah, OK.

@A-Hilaly, unless I'm mistaken, the cross-account resource management e2e test *requires* a Kubernetes `Namespace` to be created (and subsequently annotated), right? Was this e2e test assuming someone had already created a 'testing' `Namespace`? If so, we need to modify the test script to create the 'testing' `Namespace` and ensure it is annotated appropriately.

A-Hilaly: Exactly, I already had a 'testing' `Namespace` on my cluster and i missed script the steps to create it..

> If so, we need to modify the test script to create the 'testing' Namespace and ensure it is annotated appropriately.

Yes, we will ideally need an annotated `namespace` to test the override/default account/region features

Issue #368, opened 2020-10-04
Error from server (NotFound): error when creating "STDIN": namespaces "testing" not found


saikrishnaa:
**issue:**  

Below issue occurred while configuring ACk for S3 service.
 
make kind-test SERVICE=$SERVICE

To poke around your test manually:
export KUBECONFIG=/home/ubuntu/aws-controllers-k8s/scripts/../build/tmp-test-944e9b8d/kubeconfig
kubectl get pods -A

Error from server (NotFound): error when creating "STDIN": namespaces "testing" not found
FAIL: expected ack-test-smoke-s3--11513 to have been created in S3
2020-10-04T15:30:09.164Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": "0.0.0.0:8080"}
2020-10-04T15:30:09.167Z	INFO	setup	initializing service controller	{"aws.service": "s3"}
2020-10-04T15:30:09.264Z	INFO	setup	starting manager	{"aws.service": "s3"}
2020-10-04T15:30:09.265Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "bucket", "source": "kind source: /, Kind="}
2020-10-04T15:30:09.265Z	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}
E1004 15:30:09.267968       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.ConfigMap: configmaps is forbidden: User "system:serviceaccount:ack-system:default" cannot list resource "configmaps" in API group "" in the namespace "ack-system"
E1004 15:30:09.271967       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:ack-system:default" cannot list resource "namespaces" in API group "" at the cluster scope
2020-10-04T15:30:09.366Z	INFO	controller-runtime.controller	Starting Controller	{"controller": "bucket"}
2020-10-04T15:30:09.366Z	INFO	controller-runtime.controller	Starting workers	{"controller": "bucket", "worker count": 1}
[...]

PR #371, opened 2020-10-05
Create testing namespace and ack-role-account-map ConfigMap before running CARM smoke tests

A-Hilaly:
Issue #368 and follow-up to https://github.com/aws/aws-controllers-k8s/pull/354#discussion_r498401505

Description of changes:
- Create testing namespace and ack-role-account-map ConfigMap before running CARM smoke tests

### DIFF
--- a/test/e2e/s3/cross-account.sh
+++ b/test/e2e/s3/cross-account.sh
@@ -8,12 +8,18 @@ source "$SCRIPTS_DIR/lib/common.sh"
 source "$SCRIPTS_DIR/lib/k8s.sh"
 source "$SCRIPTS_DIR/lib/testutil.sh"
 
-
 # AWS_ACCOUNT_ID_ALT should be configured to allow cross account resources management
 # from AWS_ACCOUNT_ID
 AWS_ACCOUNT_ID_ALT=${AWS_ACCOUNT_ID_ALT:-""}
 AWS_PROFILE_ALT=${AWS_PROFILE_ALT:-""}
 AWS_REGION_ALT="eu-west-2"
+TESTING_NAMESPACE="testing-$RANDOM"
+ASSUME_POLICY_ARN=${ASSUME_POLICY_ARN:="s3FullAccess"}
+
+if [[ -z "$AWS_ACCOUNT_ID_ALT" ]] || [[ -z "$AWS_REGION_ALT" ]] || [[ -z "$AWS_PROFILE_ALT" ]]; then
+    echo "skipping cross-account tests due to missing credentials"
+    exit 0
+fi
 
 test_name="$( filenoext "${BASH_SOURCE[0]}" )"
 service_name="s3"
@@ -22,6 +28,31 @@ debug_msg "executing test: $service_name/$test_name"
 
 bucket_name="ack-test-smoke-$service_name-$AWS_ACCOUNT_ID_ALT-$RANDOM"
 resource_name="buckets/$bucket_name"
+
+# Create the testing namespace.
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: $TESTING_NAMESPACE
+  annotations:
+    services.k8s.aws/default-region: "$AWS_REGION_ALT"
+    services.k8s.aws/owner-account-id: "$AWS_ACCOUNT_ID_ALT"
+EOF
+
+# Create the ack-role-account-map ConfigMap.
+# See CARM design: https://github.com/aws/aws-controllers-k8s/blob/main/docs/design/proposals/carm/cross-account-resource-management.md#storing-aws-role-arns
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ack-role-account-map
+  namespace: ack-system
+data:
+  "$AWS_ACCOUNT_ID_ALT": arn:aws:iam::$AWS_ACCOUNT_ID_ALT:role/$ASSUME_POLICY_ARN
+EOF
+
+sleep 5
 
 check_is_installed jq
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -32,7 +32,7 @@ all: test
 build-ack-generate:	## Build ack-generate binary
 	go build ${GO_TAGS} ${GO_LDFLAGS} -o bin/ack-generate cmd/ack-generate/main.go
 
-test: | mocks	## Run code tests
```
A-Hilaly, line 35: Just triggering the CI... I should remove this before merge

A-Hilaly:
Having few issues with github secrets ... currently testing https://github.com/marketplace/actions/configure-aws-credentials-action-for-github-actions

A-Hilaly:
Well `secrets are not passed to the runner when a workflow is triggered from a forked repository.`
^^ https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow

test/e2e/s3/cross-account.sh
```diff
@@ -8,12 +8,18 @@ source "$SCRIPTS_DIR/lib/common.sh"
...
 source "$SCRIPTS_DIR/lib/testutil.sh"
 
-
 # AWS_ACCOUNT_ID_ALT should be configured to allow cross account resources management
 # from AWS_ACCOUNT_ID
 AWS_ACCOUNT_ID_ALT=${AWS_ACCOUNT_ID_ALT:-""}
 AWS_PROFILE_ALT=${AWS_PROFILE_ALT:-""}
 AWS_REGION_ALT="eu-west-2"
+TESTING_NAMESPACE=${TESTING_NAMESPACE:-"testing"}
+ASSUME_POLICY_ARN=${RECOMMENDED_ROLE_ARN:="s3FullAccess"}
```
jaypipes, line 17: Where does RECOMMENDED_ROLE_ARN come from?

test/e2e/s3/cross-account.sh
```diff
@@ -23,6 +29,32 @@ debug_msg "executing test: $service_name/$test_name"
 bucket_name="ack-test-smoke-$service_name-$AWS_ACCOUNT_ID_ALT-$RANDOM"
 resource_name="buckets/$bucket_name"
 
+# Create the testing namespace.
+# Note: this is step is not necessary.
```
jaypipes, line 33: I would rather that the namespace under test here be created and deleted by the test itself. In other words, let's make the test entirely idempotent and self-isolated. Create a randomly-named Namespace automatically at the start of the test and delete it at the end.

test/e2e/s3/cross-account.sh
```diff
@@ -23,6 +29,32 @@ debug_msg "executing test: $service_name/$test_name"
...
+# See CARM design: https://github.com/aws/aws-controllers-k8s/blob/main/docs/design/proposals/carm/cross-account-resource-management.md#storing-aws-role-arns
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ack-role-account-map
+  namespace: ack-system
+data:
+  "$AWS_ACCOUNT_ID_ALT": arn:aws:iam::$AWS_ACCOUNT_ID_ALT:role/$ASSUME_POLICY_ARN
+EOF
```
jaypipes, line 54: I **think** that `kubectl apply` against a ConfigMap will no a strategic merge patch against the `data` field, right? In other words, by default it will not replace the `data` field with what you are passing here and instead **merge-update** the keys in the `data` map field.

test/e2e/s3/cross-account.sh
```diff
@@ -49,9 +81,6 @@ apiVersion: s3.services.k8s.aws/v1alpha1
 kind: Bucket
 metadata:
   name: $bucket_name
-  annotations:
-    services.k8s.aws/default-region: $AWS_REGION_ALT
-    services.k8s.aws/owner-account-id: $AWS_ACCOUNT_ID_ALT
```
jaypipes, line 54: :+1:  :)

test/e2e/s3/cross-account.sh
A-Hilaly: Thank you for pointing this, It’s a very interesting point! I think that It depends on how the previous configuration was created/updated. There are some cases where `kubectl apply` can completely replace a part of the `data` field (if not all of it).

Two rules i'm aware of:
- running `kubectl create` then `kubectl apply` = will do a **merge-update**
- running `kubectl apply` then another `kubectl apply` = will replace the previous `apply`-ed configuration

A-Hilaly: I just tested this:

```bash
cat <<EOF | kubectl create -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: testmap
data:
  key1: value1
EOF

# This `kubectl apply` will perform a strategic merge
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: testmap
data:
  key2: value2
EOF

# This `kubectl apply` will replace the previous `kubectl apply`-ed configuration
cat <<EOF | kubectl create -f -
apiVersion: v1
kind: ConfigMap
metadata:
  name: testmap
data:
  key3: value3
EOF
```

at the end the ConfigMap looks like:

```yaml
key1: value1
key3: value3
```

A-Hilaly: @jaypipes I'm not sure if this is an issue, but I guess this can cause some unintended behavior if a user runs the smoke tests against a "real" cluster (since there are no rollout mechanisms for ConfigMap) Do you think that we should use `kubectl patch` instead ?

test/e2e/s3/cross-account.sh
A-Hilaly: I was thinking about reading this role from `$service/config/iam/recommended-policy-arn` file, but i don't think it's correct anymore. Changing to `ASSUME_POLICY_ARN=${ASSUME_POLICY_ARN:="s3FullAccess"}`

test/e2e/s3/cross-account.sh
A-Hilaly: ++ Done.

test/e2e/s3/cross-account.sh
jaypipes: I'm OK with this for now. We just need to be aware of this behaviour and perhaps do a safer technique in a followup patch. :)

jaypipes (approved):
:+1:

Issue #380, opened 2020-10-07
ElastiCache e2e tests for replication group, subnet group


kumargauravsharma:
**Is your feature request related to a problem?**
Request to have e2e tests for replication group, subnet group custom resources.

**Describe the solution you'd like**
e2e tests that cover following scenarios:

**Replication Group**: 
  * create
  * read
   * update
   * delete
   * increase node groups
   * decrease node groups
   * increase replicas
   * decrease replicas

**Subnet Group**: 
  * create
  * read
   * update
   * delete

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

PR #381, opened 2020-10-07
ElastiCache e2e tests for replication group, subnet group

kumargauravsharma:
Issue #380 

Description of changes:
* e2e tests for replication group and subnet group
* update in replication group custom update code to handle missing configuration scenario for decrease node groups

### DIFF
--- a/scripts/lib/aws_elasticache_testutil.sh
+++ b/scripts/lib/aws_elasticache_testutil.sh
@@ -1,0 +1,297 @@
+#!/usr/bin/env bash
+
+#################################################
+# functions for tests
+#################################################
+
+# print_k8s_ack_controller_pod_logs prints kubernetes ack controller pod logs
+# this function depends upon testutil.sh
+print_k8s_ack_controller_pod_logs() {
+  local ack_ctrl_pod_id=$( controller_pod_id )
+  kubectl logs -n ack-system "$ack_ctrl_pod_id"
+}
+
+#################################################
+# functions for test data preparation
+#################################################
+
+# get_default_vpc puts default and available vpc id to standard output
+# get_default_vpc returns status 1 if no default vpc is not found
+# Example usage:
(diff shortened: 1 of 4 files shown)


scripts/lib/aws_elasticache_testutil.sh
```diff
@@ -0,0 +1,297 @@
...
+#################################################
+# functions for tests
+#################################################
+
+# print_k8s_ack_controller_pod_logs prints kubernetes ack controller pod logs
+# this function depends upon testutil.sh
+print_k8s_ack_controller_pod_logs() {
+  local ack_ctrl_pod_id=$( controller_pod_id )
+  kubectl logs -n ack-system "$ack_ctrl_pod_id"
+}
```
jaypipes, line 12: ++

We should pull this into `scripts/lib/k8s.sh` (some future PR, doesn't need to be this one)

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -125,9 +125,9 @@ func (rm *resourceManager) diffReplicasNodeGroupConfiguration(
 			continue
 		}
 		latestReplicaCount := 0
-		for _, latestShardMember := range latestShard.NodeGroupMembers {
-			if latestShardMember.CurrentRole != nil && *latestShardMember.CurrentRole == "replica" {
-				latestReplicaCount++
+		if latestShard.NodeGroupMembers != nil {
+			if len(latestShard.NodeGroupMembers) > 0 {
+				latestReplicaCount = len(latestShard.NodeGroupMembers) - 1
```
jaypipes, line 130: good simplification with above changes. :+1:

jaypipes (approved):
I think this is good as-is. Definitely we can work on making these types of tests and utility library functions more readable and componentized, but we can do that in future PRs.

PR #388, opened 2020-10-08
`replacePkgName` handle cases of nested slices and nested maps

A-Hilaly:
Issue N/A

Description of changes:
- `replacePkgName` handle cases where `goType` is a nested slice or a nested map
- detach `replacePkgName` from `model.CRD` and change it signature
- Add unit tests for `replacePkgName`

### DIFF
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -1329,7 +1329,7 @@ func (r *CRD) goCodeVarEmptyConstructorSDKType(
 	indent := strings.Repeat("\t", indentLevel)
 	goType := shape.GoTypeWithPkgName()
 	keepPointer := (shape.Type == "list" || shape.Type == "map")
-	goType = r.replacePkgName(goType, "svcsdk", keepPointer)
+	goType = replacePkgName(goType, r.sdkAPI.API.PackageName(), "svcsdk", keepPointer)
 	switch shape.Type {
 	case "structure":
 		// f0 := &svcsdk.BookData{}
@@ -1355,7 +1355,7 @@ func (r *CRD) goCodeVarEmptyConstructorK8sType(
 	indent := strings.Repeat("\t", indentLevel)
 	goType := shape.GoTypeWithPkgName()
 	keepPointer := (shape.Type == "list" || shape.Type == "map")
-	goType = r.replacePkgName(goType, "svcapitypes", keepPointer)
+	goType = replacePkgName(goType, r.sdkAPI.API.PackageName(), "svcapitypes", keepPointer)
 	goTypeNoPkg := goType
 	goPkg := ""
 	hadPkg := false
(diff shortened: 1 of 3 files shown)


pkg/model/crd.go
```diff
@@ -2294,19 +2294,25 @@ func (r *CRD) goCodeSetOutputForScalar(
...
+	apiPkgName string,
 	replacePkgAlias string,
 	keepPointer bool,
 ) string {
 	memberType := subject
+	sliceDepth := 0 // Depth of the slice type
 	isSliceType := strings.HasPrefix(memberType, "[]")
 	if isSliceType {
-		memberType = memberType[2:]
+		sliceDepth = strings.LastIndex(subject, "[]")/2 + 1
```
jaypipes, line 2307: heh, OK :)

pkg/model/crd_test.go
```diff
@@ -0,0 +1,78 @@
...
+		},
+		{ // don't keep pointer
+			"*ecr.Repository",
+			"ecr",
+			"svcsdk",
+			false,
+			"svcsdk.Repository",
+		},
+		{ // non sdk type
+			"*anotherpackage.Repository",
```
jaypipes, line 33: let's use a specific real-world example here, yeah? How about `*time.Time`?

pkg/model/crd_test.go
```diff
@@ -0,0 +1,78 @@
...
+			"ecr",
+			"svcsdk",
+			true,
+			"*anotherpackage.Repository",
+		},
+		{ // map type
```
jaypipes, line 39: Do we happen to have a specific real-world example from one of our APIs?

pkg/model/crd_test.go
```diff
@@ -0,0 +1,78 @@
...
+			"map[string]*ecr.Repository",
+			"ecr",
+			"svcsdk",
+			true,
+			"map[string]*svcsdk.Repository",
+		},
+		{ // nested map type
```
jaypipes, line 46: Do we happen to have a specific real-world example from one of our APIs?

pkg/model/crd_test.go
```diff
@@ -0,0 +1,78 @@
...
+			"map[string]map[string]*svcsdk.Repository",
+		},
+		{ // slice type
+			"[]ecr.Repository",
+			"ecr",
+			"svcsdk",
+			true,
+			"[]svcsdk.Repository",
+		},
+		{ // nested slices type
```
jaypipes, line 60: Do we happen to have a specific real-world example from one of our APIs?

jaypipes (changes requested):
this is great, Amine. Just would love to put some real-world examples into the unit test scenarios to make it clearer which specific types this code is fixing up/covering.

pkg/model/crd_test.go
A-Hilaly: Done! with an example from codedeploy

pkg/model/crd_test.go
A-Hilaly: Yes, `map[string]*ecr.Repository` is a real example in the `ecr` service

jaypipes (approved):
:+1:

pkg/model/crd_test.go
A-Hilaly: I couldn't find a real example within the services i generated.. I'm using  `map[string]map[string]uint8` for now.. WDYT @jaypipes ?

pkg/model/crd_test.go
A-Hilaly: Yes, it's a good example!

Issue #378, opened 2020-10-06
NoOperationFault while scaling up ReplicationGroup


nmvk:
**Describe the bug**
`NoOperationFault` while scaling up ReplicationGroup which is cluster mode enabled.

```
"controller": "replicationgroup", "request": "default/test", "error": "NoOperationFault: Requested new replica count already matches current number of replicas.\n\tstatus code: 400
```

**Steps to reproduce**

```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: ReplicationGroup
metadata:
  name: test
spec:
    engine: redis 
    replicationGroupID: test
    replicationGroupDescription: test
    automaticFailover: true
    cacheNodeType: cache.t3.micro
    numNodeGroups: 2
    replicasPerNodeGroup: 1
```

Change `cacheNodeType` to `cache.t3.small`

**Expected outcome**
Scale up modification

**Environment**

* Kubernetes version 1.17
* Using EKS (yes/no), if so version? Yes
* AWS service targeted (S3, RDS, etc.) Elasticache

nmvk:
Duplicate https://github.com/aws/aws-controllers-k8s/issues/377

nmvk:
Probably similar issue with subnet 
```
2020-10-07T11:33:09.739-0700    ERROR   controller-runtime.controller   Reconciler error        {"controller": "cachesubnetgroup", "request": "default/demo-cache-subnet-group", "error": "InvalidParameterValue: No modifications were requested.\n\tstatus code: 400, request id: "}
```

PR #393, opened 2020-10-12
Update status after Read operation and Requeue sync based on Conditions

kumargauravsharma:
Issue #378 

Description of changes:
* Added reconcile logic to Requeue after default interval (based on Conditions).
* Updated reconcile logic to updateCRStatus using already 'read' status (using ReadOne). This helps provide updated status to user when an ongoing operation takes long on remote service side and rm.update() fails due ongoing operation.
  * Added logic to populate `OwnerAccountID` and `Conditions` inside sdkFind, as these is required in ko.Status
* Updated ElastiCache replicationgroup custom update method to check of available state in order to proceed.
* Updated ElastiCache replicationgroup custom output set method to populate `Conditions` with `ConditionTypeResourceSynced`
* Updated diff reporter to not include properties values. It lead to runtime errors when accessing properties using reflection

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -187,12 +187,39 @@ func (r *reconciler) sync(
 			"arn", latest.Identifiers().ARN(),
 			"is_adopted", isAdopted,
 		)
+		// Before we update the backend AWS service resources, let's first update
+		// the latest status of CR which was retrieved by ReadOne call.
+		// Else, latest read status is lost in-case Update call fails with error.
+		err = r.updateCRStatus(ctx, desired, latest)
+		if err != nil {
+			return err
+		}
 		latest, err = rm.Update(ctx, desired, latest, diffReporter)
 		if err != nil {
 			return err
 		}
 		r.log.V(0).Info("reconciler.sync updated resource")
 	}
+	err = r.updateCRStatus(ctx, desired, latest)
+	if err != nil {
+		return err
+	}
+	for _, condition := range latest.Conditions() {
+		if condition.Type == ackv1alpha1.ConditionTypeResourceSynced &&
+			condition.Status != corev1.ConditionTrue {
+			return requeue.NeededAfter(
+				errors.New("sync again"), requeue.DefaultRequeueAfterDuration)
+		}
+	}
+	return nil
+}
+
+// updateCRStatus updates status of CR using the supplied latest resource.
+func (r *reconciler) updateCRStatus(
+	ctx context.Context,
+	desired acktypes.AWSResource,
+	latest acktypes.AWSResource,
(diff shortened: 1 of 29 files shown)


pkg/runtime/reconciler.go
```diff
@@ -187,12 +187,39 @@ func (r *reconciler) sync(
...
 	}
+	err = r.updateCRStatus(ctx, desired, latest)
+	if err != nil {
+		return err
+	}
+	for _, condition := range latest.Conditions() {
+		if condition.Type == ackv1alpha1.ConditionTypeResourceSynced &&
+			condition.Status != corev1.ConditionTrue {
+			return requeue.NeededAfter(
+				errors.New("sync again"), requeue.DefaultRequeueAfterDuration)
```
jaypipes, line 211: Instead of using `errors.New()` here, please create a new error variable in the `pkg/errors` package:

https://github.com/aws/aws-controllers-k8s/blob/main/pkg/errors/error.go

How about we call this error `TemporaryOutOfSync` or something like that, to indicate the error isn't really an error but more of a marker that the status check will be performed after some wait time?

scripts/lib/aws_elasticache_testutil.sh
```diff
@@ -122,7 +122,6 @@ spec:
     engine: redis
     replicationGroupID: $rg_id
     replicationGroupDescription: $rg_description
-    automaticFailover: true
```
jaypipes, line 125: what's this for?

templates/pkg/crd_sdk.go.tpl
```diff
@@ -192,17 +213,17 @@ func (rm *resourceManager) sdkCreate(
...
 		ko.Status.ACKResourceMetadata = &ackv1alpha1.ResourceMetadata{}
 	}
 	if ko.Status.ACKResourceMetadata.OwnerAccountID == nil {
 		ko.Status.ACKResourceMetadata.OwnerAccountID = &rm.awsAccountID
 	}
 	ko.Status.Conditions = []*ackv1alpha1.Condition{}
+	{{ if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Create }}
+		// custom set output from response
+		rm.{{ $setOutputCustomMethodName }}(r, resp, ko)
+	{{ end }}
```
jaypipes, line 226: the above changes are fine for this PR, but in the future it would be nice to place the ACKResourceMetadata and Conditions setting into a helper function called `setStatusDefaults()` or something like that to reduce the repetition :)

jaypipes (changes requested):
@kumargauravsharma this is looking very nice, ++. Couple requests inline for ya, though, nothing too major.

Issue #355, opened 2020-10-01
Produce a set of standard Prometheus metrics


jaypipes:
We currently have no metrics story. The ACK service controllers neither gather nor publish any metrics, which means our observability and operational story is poor.

**What is needed?**

- [ ] Determine the set of metrics we want to track
- [ ] **Common metrics utility/runtime**: Create a `pkg/metrics` package that contains the set of standardized Prometheus metrics
- [ ] **Instrument the service controller**: Modify the `pkg/runtime.ServiceController` to instantiate the collection of Prometheus metrics and pass them to the important member structs (like the `reconciler`, which would need to pass the metrics to the `ResourceManagerFactory` structs that register with it, which would need to pass the metrics to the `ResourceManager` structs that manage resources for a particular AWS Account, etc)
- [ ] **Publish the collected metrics**: Expose the metric values using a `/metrics`  HTTP(S) endpoint within the ACK service controller binary
- [ ] (optional) Create a simple metrics-scraping helper that can call the various ACK service controller Pod's `/metrics` endpoint periodically and either aggregate or publish (to CloudWatch or some other service) the aggregated metrics values. We could optionally just recommend that the [`prometheus` server](https://prometheus.io/docs/prometheus/latest/configuration/configuration/) be configured to [scrape](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) the endpoints instead of writing a custom metrics helper just for ACK.

**What metrics do we want to track?**

I think the following metrics are required, at a minimum:

* `ack_outbound_api_request_total`: the total number of API calls made to the a backend AWS service API
  * *Type*: Counter
  * *Labels*
    * service=<SERVICE_ID> -- e.g. "sns"
    * op_type=["CREATE", "UPDATE", "READ_ONE", "READ_MANY", "GET_ATTRIBUTES", "SET_ATTRIBUTES", "DELETE"]
    * operation_id=<API OperationID> -- e.g. "CreateTopic"
* `ack_outbound_api_error_total`: The total number of API calls that returned a 4XX or 5XX HTTP status code
  * *Type*: Counter
  * *Labels*
    * service=<SERVICE_ID> -- e.g. "sns"
    * operation_id=<API OperationID> -- e.g. "CreateTopic"
    * status_code=<HTTP Status Code> -- e.g. 429
* `ack_reconcile_total`: The total number of times the service controller's reconciler loops were executed 
  * *Type*: Counter
  * *Labels*
[...]

PR #406, opened 2020-10-13
plumb log and metrics down into resource managers

jaypipes:
We want to use the service controller's logger and metrics struct from
within the lowest-level resource managers and SDK linkage. This patch
modifies the `pkg/types.AWSResourceManagerFactor:ManagerFor` interface
method to pass in a `logr.Logger` and `*ackmetrics.Metrics` that is then
placed into the `resourceManager` that is created by each resource
package's `NewResourceManager` method.

Also cleans up our mock generation to only generate mocks for the
`pkg/types` which is our package containing interfaces. There's no need
to generate mocks for any other package.

Issue #355


A-Hilaly (approved):
/lgtm

PR #408, opened 2020-10-13
Append an underscore to CRD names matching Go keywords

A-Hilaly:
Issue N/A

Description of changes:
- Append an underscore to CRD names matching Go keywords

This was originally detected during the process of generating `es` (elastic search) service controller. "elastic search'' uses a data type named "package" which causes `ack-generate controller` to generate a Go package called "package". 
This PR will help preventing `ack-generate` from using Go keywords as CRD names.

For more details see [Elastic Search API](https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-configuration-api.html#es-configuration-api-actions-createpackage) or try to run `make build-controller SERVICE=es`

### DIFF
--- a/pkg/names/names.go
+++ b/pkg/names/names.go
@@ -18,6 +18,8 @@ import (
 
 	regexp "github.com/dlclark/regexp2" // for negative lookahead support
 	"github.com/iancoleman/strcase"
+
+	"github.com/aws/aws-controllers-k8s/pkg/util"
 )
 
 type initialismTranslator struct {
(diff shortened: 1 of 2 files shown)


pkg/names/names.go
```diff
@@ -115,6 +117,34 @@ var (
 	}
 )
 
+var goKeyWords = []string{
```
jaypipes, line 120: femto-nit: should be `goKeywords` since "keyword" is a single word. :)

jaypipes (changes requested):
See inline for a concern...

pkg/names/names.go
A-Hilaly: Haaa TIL 👍

pkg/names/names.go
```diff
@@ -150,6 +180,9 @@ func goName(original string, lowerFirst bool, snake bool) (result string) {
 	if snake {
 		result = strcase.ToSnake(result)
 	}
+	if util.InStrings(result, goKeyWords) {
+		result = "_" + result
```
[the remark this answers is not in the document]
A-Hilaly: Nice catch, totally forgot about this https://golang.org/ref/spec#Identifiers

nithu0115 (approved):
👍

jaypipes (approved):
parfait.

Issue #409, opened 2020-10-14
Inject session handler user-agent info for monitoring


nithu0115:
**Is your feature request related to a problem?**

Today, there is no ability to check which user-agent has made the API calls to a respective service. And if we run into any problems, ACK users have to either filter via timestamp and search the entire Cloudtrail events/logs. This process is tedious. Along with the use-case, if service teams want to track the activity/rate of adoption of ACK, calling session handler user-agent will be helpful filter.

nithu0115:
Fixed in #410

PR #410, opened 2020-10-14
session handler info

nithu0115:
Issue #, if available: #409 

Description of changes:
With this PR, we will inject session handler information/user-agent information while creating the session. 

Tested with debug logging:
```
Host: api.ecr.us-west-2.amazonaws.com
**User-Agent: aws-controller-k8s/0c76963-dirty (GitCommit/0c76963771053ea4310b5ffd0aaa399af0f88e87; BuildDate/2020-10-14T00:33:32+0000; Kind/Repository; CRDVersion/v1alpha1) aws-sdk-go/1.34.32 (go1.14.1; linux; amd64)
Content-Length: 72**
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ ENV GO111MODULE=on
 ENV GOARCH=amd64
 ENV GOOS=linux
 ENV CGO_ENABLED=0
+ENV VERSION_PKG=github.com/aws/aws-controllers-k8s/pkg/version
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
@@ -19,10 +20,16 @@ RUN  go mod download
 # Copy the go source
 COPY . $work_dir/
 # Build
-RUN  go build -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
+RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
+    GIT_COMMIT=$(git rev-parse HEAD) && \
+    BUILD_DATE=$(date +%Y-%m-%dT%H:%M) && \
+    go build -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} \
+    -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} \
+    -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
+    -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
 
 FROM amazonlinux:2
 ARG work_dir=/github.com/aws/aws-controllers-k8s
 WORKDIR /
-COPY --from=builder $work_dir/bin/controller $work_dir/LICENSE $work_dir/ATTRIBUTION.md /bin/.
+COPY --from=builder $work_dir/bin/controller $work_dir/LICENSE $work_dir/ATTRIBUTION.md /bin/
 ENTRYPOINT ["/bin/controller"]
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -110,7 +110,7 @@ func (r *reconciler) reconcile(req ctrlrt.Request) error {
 	acctID := r.getOwnerAccountID(res)
 	region := r.getRegion(res)
 	roleARN := r.getRoleARN(acctID)
-	sess, err := NewSession(region, roleARN)
+	sess, err := NewSession(region, roleARN, res.RuntimeObject().GetObjectKind().GroupVersionKind())
 	if err != nil {
 		return err
 	}
--- a/pkg/runtime/session.go
+++ b/pkg/runtime/session.go
@@ -14,12 +14,19 @@
 package runtime
 
 import (
+	"fmt"
+
 	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
+	"github.com/aws/aws-controllers-k8s/pkg/version"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/request"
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -5,11 +5,8 @@ ARG service_alias
 ARG work_dir=/github.com/aws/aws-controllers-k8s
 WORKDIR $work_dir
 # For building Go Module required
-ENV GOPROXY=https://proxy.golang.org,direct
-ENV GO111MODULE=on
-ENV GOARCH=amd64
-ENV GOOS=linux
-ENV CGO_ENABLED=0
+ENV GOPROXY=direct
```
A-Hilaly, line 8: I'm curious, did you encounter any issue caused by the Go proxy ?
We made [this change a month ago](https://github.com/aws/aws-controllers-k8s/pull/314) because `gomodules.xyz` was down for more than 24h and was causing all the service controllers build to fail

Dockerfile
```diff
@@ -19,10 +16,14 @@ RUN  go mod download
 # Copy the go source
 COPY . $work_dir/
 # Build
-RUN  go build -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
+RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
+    GIT_COMMIT=$(git rev-parse HEAD) && \
+    BUILD_DATE=$(date +%Y-%m-%dT%H:%M:%S%z) && \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
+    -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
```
A-Hilaly, line 23: nit: i'm fine with this change but i find it easier to read when the environment variables are declared separately

A-Hilaly (commented):
LGTM overall, leaving few

Dockerfile
nithu0115: Ah, yes. I kept running into sink-hole issue while I running builds with VPN. I modified this during testing and forgot to revert it back. Thanks for catching :D

kumargauravsharma (commented):
As an AWS Service controller would have corresponding binary as container image; the current PR would have same user agent string for all service controller built against commit level.

How about having AWS service name also as part of user agent? Something like: `aws-controller-k8s/elasticache-0c76963-dirty`

nithu0115:
> As an AWS Service controller would have corresponding binary as container image; the current PR would have same user agent string for all service controller built against commit level.
> 
> How about having AWS service name also as part of user agent? Something like: `aws-controller-k8s/elasticache-0c76963-dirty`

Done, addressed this comment. 
```
User-Agent: aws-controller-k8s/ecr.services.k8s.aws-b216274-dirty (GitCommit/b2162744daa8d5cb631a535df492c8e12254887e; BuildDate/2020-10-14T18:44:02-0500; CRDKind/Repository; CRDVersion/v1alpha1) aws-sdk-go/1.34.32 (go1.15.1; darwin; amd64)
```

kumargauravsharma (approved):

A-Hilaly (approved):

Dockerfile
jaypipes: I'm actually not fine with this change... @nithu0115 is there a reason you changed this like so instead of adding another environment variable?

Dockerfile
```diff
@@ -19,10 +16,17 @@ RUN  go mod download
...
 # Build
-RUN  go build -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
+RUN GIT_VERSION=$(git describe --tags --dirty --always) && \
+    GIT_COMMIT=$(git rev-parse HEAD) && \
+    BUILD_DATE=$(date +%Y-%m-%dT%H:%M) && \
+    CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on \
+    go build -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} \
+    -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} \
+    -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
+    -a -o $work_dir/bin/controller $work_dir/services/$service_alias/cmd/controller/main.go
```
jaypipes, line 26: Is the reason you changed the above from using ENV variables to using CLI variables because you needed to execute shell commands to populate things like GIT_COMMIT? If so, why move the static env vars like CGO_ENABLED and GOOS to the command line invocation?

go.mod
```diff
@@ -9,11 +9,11 @@ require (
...
-	github.com/google/go-cmp v0.3.0
+	github.com/google/go-cmp v0.4.0
 	github.com/iancoleman/strcase v0.0.0-20191112232945-16388991a334
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_golang v1.0.0
+	github.com/prometheus/client_golang v1.7.1
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.5.1
```
jaypipes, line 19: please undo the changes to `go.mod` and `go.sum`. They are not germane to this pull request's content. If we want to update these packages, let's do that in a separate commit that is specific to the new code brought in with those new versions.

pkg/runtime/reconciler.go
```diff
@@ -15,7 +15,6 @@ package runtime
 
 import (
 	"context"
-
```
jaypipes, line 18: please restore this newline.

jaypipes (changes requested):
@nithu0115 biggest change I need you to make on this is to undo the changes to go.mod and go.sum please.

Dockerfile
nithu0115: I thought as these are just build an argument, so I went ahead and added them as CLI variables. I can revert this change and add all of these as env vars.

PR #411, opened 2020-10-14
renew credentials during e2e test

kumargauravsharma:
Issue #, if available:

Description of changes:
Renew credentials during e2e test.

`SERVICE=elasticache && make kind-test` ran the tests:
```
...
DEBUG: executing test: elasticache/smoke
...
DEBUG: Creating subnet group ack-test-smoke-elasticache-subnet-gp in elasticache
cachesubnetgroup.elasticache.services.k8s.aws/ack-test-smoke-elasticache-subnet-gp created
DEBUG: checking subnet group ack-test-smoke-elasticache-subnet-gp created in elasticache
DEBUG: Modifying subnet group ack-test-smoke-elasticache-subnet-gp in elasticache
cachesubnetgroup.elasticache.services.k8s.aws/ack-test-smoke-elasticache-subnet-gp configured
DEBUG: subnet group ack-test-smoke-elasticache-subnet-gp modified in elasticache
DEBUG: Deleting subnet group ack-test-smoke-elasticache-subnet-gp in elasticache
cachesubnetgroup.elasticache.services.k8s.aws "ack-test-smoke-elasticache-subnet-gp" deleted
smoke took 69 second(s)
DEBUG: executing test: elasticache/e2e
generating AWS temporary credentials and adding to env vars map ... DEBUG: Testing create replication group: ack-test-rg-1.
replicationgroup.elasticache.services.k8s.aws/ack-test-rg-1 created
generating AWS temporary credentials and adding to env vars map ... DEBUG: starting to wait for replication group: ack-test-rg-1 to be available.
generating AWS temporary credentials and adding to env vars map ... DEBUG: Testing modify replication group: ack-test-rg-1.
replicationgroup.elasticache.services.k8s.aws/ack-test-rg-1 configured
generating AWS temporary credentials and adding to env vars map ... DEBUG: starting to wait for replication group: ack-test-rg-1 to be available.
generating AWS temporary credentials and adding to env vars map ... DEBUG: Testing modify replication group: ack-test-rg-1 shards count to new value: 3.
replicationgroup.elasticache.services.k8s.aws/ack-test-rg-1 configured
generating AWS temporary credentials and adding to env vars map ... DEBUG: starting to wait for replication group: ack-test-rg-1 to be available.

Waiter ReplicationGroupAvailable failed: Max attempts exceeded
generating AWS temporary credentials and adding to env vars map ... DEBUG: starting to wait for replication group: ack-test-rg-1 to be available.
[...]

Issue #231, opened 2020-08-21
[Elasticache] Handling terminal conditions


nmvk:
**Is your feature request related to a problem?**
In Elasticache we may encounter asynchronous error when we hit few cases like deleting a master node in `DecreaseReplicaCount` or `InsufficientCacheClusterCapacity` in a particular AZ. Elasticache today provides notifications related to these errors in `DescribeEvents` and `SNS` notifications.

**Describe the solution you'd like**
Gracefully exit after few retries, and communicate about the error

**Describe alternatives you've considered**
Integrate these k8 events so one can get this information when they do `kubectl describe CacheCluster`

jaypipes:
let's break this into two steps:

1) for dev-preview let's do the static checks on Status field values (things like `IncompatibleNetwork` and `CreateFailed`) and add that logic into the `sdkCreate` and/or `sdkUpdate`

2) for beta or later, we can add some custom callback code that calls, for instance, the Elasticache DescribeEvents stream.

PR #412, opened 2020-10-14
add terminal condition during sync

kumargauravsharma:
Issue #231 

Description of changes:
Added terminal condition logic during reconciler sync.
Updated templates files for it and generated controllers using
```
SERVICE=apigatewayv2 && make build-controller && \
SERVICE=dynamodb && make build-controller && \
SERVICE=ecr && make build-controller && \
SERVICE=elasticache && make build-controller && \
SERVICE=s3 && make build-controller && \
SERVICE=sns && make build-controller;
```

The updates are independent of pr #411 however commit is based on it.

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -151,6 +151,14 @@ func (r *reconciler) sync(
 
 	latest, err := rm.ReadOne(ctx, desired)
 	if err != nil {
+		if latest != nil {
+			// this indicates, that even though ReadOne failed
+			// there is some changes available in the latest.RuntimeObject()
+			// (example: ko.Status.Conditions) which have been
+			// updated in the resource
+			// Thus, updateCRStatus() call should be made here
+			_ = r.updateCRStatus(ctx, desired, latest)
+		}
 		if err != ackerr.NotFound {
 			return err
 		}
@@ -168,6 +176,14 @@ func (r *reconciler) sync(
 
 		latest, err = rm.Create(ctx, desired)
 		if err != nil {
+			if latest != nil {
+				// this indicates, that even though Create failed
+				// there is some changes available in the latest.RuntimeObject()
+				// (example: ko.Status.Conditions) which have been
+				// updated in the resource
+				// Thus, updateCRStatus() call should be made here
+				_ = r.updateCRStatus(ctx, desired, latest)
+			}
 			return err
 		}
 		r.log.V(0).Info(
@@ -196,6 +212,14 @@ func (r *reconciler) sync(
 		}
 		latest, err = rm.Update(ctx, desired, latest, diffReporter)
 		if err != nil {
+			if latest != nil {
+				// this indicates, that even though update failed
+				// there is some changes available in the latest.RuntimeObject()
+				// (example: ko.Status.Conditions) which have been
+				// updated in the resource
+				// Thus, updateCRStatus() call should be made here
+				_ = r.updateCRStatus(ctx, desired, latest)
+			}
 			return err
 		}
 		r.log.V(0).Info("reconciler.sync updated resource")
@@ -332,7 +356,7 @@ func (r *reconciler) getAWSResource(
 // handleReconcileError will handle errors from reconcile handlers, which
 // respects runtime errors.
 func (r *reconciler) handleReconcileError(err error) (ctrlrt.Result, error) {
-	if err == nil {
+	if err == nil || err == ackerr.Terminal{
 		return ctrlrt.Result{}, nil
 	}
 
@@ -350,7 +374,7 @@ func (r *reconciler) handleReconcileError(err error) (ctrlrt.Result, error) {
 	var requeueNeeded *requeue.RequeueNeeded
 	if errors.As(err, &requeueNeeded) {
 		r.log.V(1).Info(
-			"requeue needed after error",
+			"requeue needed error",
 			"error", requeueNeeded.Unwrap(),
 		)
 		return ctrlrt.Result{Requeue: true}, nil
--- a/services/elasticache/pkg/resource/replication_group/custom_set_conditions.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_set_conditions.go
@@ -1,0 +1,60 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package replication_group
+
+import (
+	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
+	svcapitypes "github.com/aws/aws-controllers-k8s/services/elasticache/apis/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// CustomUpdateConditions sets conditions (terminal) on supplied replication group
+// it examines supplied resource to determine conditions.
+// It returns true if conditions are updated
+func (rm *resourceManager) CustomUpdateConditions(
+	ko *svcapitypes.ReplicationGroup,
+	r *resource,
+	err error,
+) bool {
+	rgStatus := r.ko.Status.Status
+	if rgStatus == nil || *rgStatus != "create-failed" {
+		return false
+	}
+	// Terminal condition
+	var terminalCondition *ackv1alpha1.Condition = nil
+	if ko.Status.Conditions == nil {
+		ko.Status.Conditions = []*ackv1alpha1.Condition{}
+	} else {
+		for _, condition := range ko.Status.Conditions {
+			if condition.Type == ackv1alpha1.ConditionTypeTerminal {
+				terminalCondition = condition
+				break
+			}
+		}
+		if terminalCondition != nil && terminalCondition.Status == corev1.ConditionTrue {
+			// some other exception already put the resource in terminal condition
+			return false
+		}
+	}
+	if terminalCondition == nil {
+		terminalCondition = &ackv1alpha1.Condition{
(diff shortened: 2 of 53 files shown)


scripts/lib/aws/elasticache.sh
```diff
@@ -12,6 +12,29 @@ SCRIPTS_DIR="$ROOT_DIR/scripts"
...
 #################################################
 
+# k8s_controller_apply_temp_creds adds AWS temporary credentials
+# to service controller running on kubernetes cluster.
+# it depends upon:
+#   $service_name
+#   $AWS_ACCESS_KEY_ID
+#   $AWS_SECRET_ACCESS_KEY
+#   $AWS_SESSION_TOKEN
+k8s_controller_apply_temp_creds() {
```
A-Hilaly, line 22: You already opened a PR for this commit cad9d79bf5e2d2d58ec2f726a06cc3c8daf8bbaf :)

kumargauravsharma: That PR was pending merge. To keep changes separate, I kept the commits as is in this PR and did not squash them.

A-Hilaly: Ohh,  didn't know Github can do a strategic merge :O

nmvk (approved):
LGTM

scripts/lib/aws/elasticache.sh
jaypipes: > Ohh, didn't know Github can do a strategic merge :O

it can't. basically, when the base PR is merged, then refreshing this page will remove the diff of the base patch's changes.

A-Hilaly (approved):
/lgtm

scripts/lib/aws/elasticache.sh
kumargauravsharma: As I updated PR #411, I have removed corresponding commit from this PR. Now, It contains only commit relevant to it.

pkg/runtime/reconciler.go
```diff
@@ -359,6 +363,55 @@ func (r *reconciler) handleReconcileError(err error) (ctrlrt.Result, error) {
 	return ctrlrt.Result{}, err
 }
 
+// updateCRTerminalConditions updates status of CR for terminal conditions
```
jaypipes, line 366: I don't believe this should be in the reconciler. The resourceManager should modify the `*resource` that is returned from one of its methods (like Update or ReadOne, etc) with a Condition indicating a terminal state... the reconciler's `updateCRStatus()` call, which should patch the CR's Status struct, should handle the call to k8sclient.Patch.

jaypipes (changes requested):
I think this is introducing a layering violation. I'd like to have the resourceManager be the thing that appends the Condition into the Status.Conditions collection, not the reconciler. The reconciler's already-existing call to k8sclient Status.Patch() should be all that is needed...

templates/pkg/crd_manager.go.tpl
```diff
@@ -168,3 +168,16 @@ func newResourceManager(
 		sdkapi:	   svcsdk.New(sess),
 	}, nil
 }
+
+// EnsureStatusDefaults ensures that the default status properties are
+// set to defaults if found nil.
+func (rm *resourceManager) EnsureStatusDefaults(
```
jaypipes, line 174: If you move the responsibility of setting/adding a new Condition on the CR to the `resourceManager`, I don't believe this method will be necessary, right? Since `setStatusDefaults` will be callable by the `resourceManager` and it does all the meat of this method, which I think you only created in order to have the `reconciler` be able to call it?

templates/pkg/crd_descriptor.go.tpl
```diff
@@ -154,3 +155,34 @@ func (d *resourceDescriptor) MarkUnmanaged(
...
+	}
+	if existingCondition == nil {
+		c.ko.Status.Conditions = append(c.ko.Status.Conditions, condition)
+	} else {
+		existingCondition.Status = condition.Status
+		existingCondition.LastTransitionTime = condition.LastTransitionTime
+		existingCondition.Reason = condition.Reason
+		existingCondition.Message = condition.Message
+	}
+}
```
jaypipes, line 188: ++ the above method is great.

pkg/types/aws_resource_manager.go
```diff
@@ -60,6 +60,10 @@ type AWSResourceManager interface {
 	// GetAttributes operations but all we have (for new CRs at least) is a
 	// name for the resource
 	ARNFromName(string) string
+
+	// EnsureStatusDefaults ensures that the default status properties are
+	// set to defaults if found nil.
+	EnsureStatusDefaults(AWSResource)
```
jaypipes, line 66: see comment from below... I think you can remove this change.

jaypipes (changes requested):
Added a couple more notes inline, @kumargauravsharma. Ping me when you're up and we can chat :)

templates/pkg/crd_sdk.go.tpl
```diff
@@ -369,3 +370,66 @@ func (rm *resourceManager) setStatusDefaults (
...
+{{- $terminalCodesMaxIndex := minus $terminalCodesCount 1 }}
+	if {{ range $index, $terminalCode := .CRD.TerminalExceptionCodes -}}
+		awsErr.Code() == "{{ $terminalCode }}"
+		{{- if eq ($index) ($terminalCodesMaxIndex) }} { {{ else }} || {{ end }}
+		{{ end -}}
+		return awsErr, true
+	}
+{{ end }}
+	return nil, false
+}
```
jaypipes, line 435: I think we can clean up this `terminalAWSError` method in a few ways:

1) Only have it return a `bool` and call it `isTerminalError()`

2) You can remove the need for the `{{ minus }}` template function like so:

```
if err == nil {
    return false
}
awsErr, ok := ackerr.AWSError(err)
if !ok {
    return false
}
{{ if .CRD.TerminalExceptionCodes }}
switch awsErr.Code() {
    case {{- range $x, $terminalCode := .CRD.TerminalExceptionCodes -}}{{ if ne ( $x 0 ) }}  || "{{ end }}{{ $terminalCode }}"{{ end }}:
        return true
    default:
        return false
    }
{{- else }}
    return false
{{- end }}
}
```

templates/pkg/crd_sdk.go.tpl
```diff
@@ -369,3 +370,66 @@ func (rm *resourceManager) setStatusDefaults (
...
+			break
+		}
+	}
+	if terminalCondition == nil {
+		terminalCondition = &ackv1alpha1.Condition{
+			Type:   ackv1alpha1.ConditionTypeTerminal,
+		}
+		ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
+	}
+	if terminalErr, ok := rm.terminalAWSError(err); ok {
```
jaypipes, line 397: Given my advice below about making the `terminalAWSError()` function only return a `bool`, you would change the above to this instead, which I think is a bit easier to read:

```go
if rm.isTerminalError(err) {
    terminalCondition.Status = corev1.ConditionTrue
    awsErr, _ := ackerr.AWSError(err)
    errorMessage := awsErr.Message()
    terminalCondition.Message = &errorMessage
}
```

templates/pkg/crd_sdk.go.tpl
```diff
@@ -369,3 +370,66 @@ func (rm *resourceManager) setStatusDefaults (
...
+	// Terminal condition
+	var terminalCondition *ackv1alpha1.Condition = nil
+	for _, condition := range ko.Status.Conditions {
+		if condition.Type == ackv1alpha1.ConditionTypeTerminal {
+			terminalCondition = condition
+			break
+		}
+	}
+	if terminalCondition == nil {
+		terminalCondition = &ackv1alpha1.Condition{
```
jaypipes, line 392: Why are we creating a terminal condition if there isn't one?

services/elasticache/pkg/resource/replication_group/custom_set_conditions.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if terminalCondition == nil {
+		terminalCondition = &ackv1alpha1.Condition{
+			Type: ackv1alpha1.ConditionTypeTerminal,
+		}
+		ko.Status.Conditions = append(ko.Status.Conditions, terminalCondition)
+	}
+
+	terminalCondition.Status = corev1.ConditionTrue
+	errorMessage := "Replication group status: create-failed"
```
jaypipes, line 58: If all we're really needing a custom update condition method for is to set this string, then it might be easier to just change the `ExceptionsConfig.TerminalCodes` Go type from `[]string` to instead be a struct with some fields that can accomodate this custom error message. Perhaps even place this struct on the `OperationConfig` struct so you can vary the message by API call, too. For instance, above is hard-coding "create-failed" when you might want to have a message customized to the update failed path...

templates/pkg/crd_manager.go.tpl
```diff
@@ -5,6 +5,8 @@ package {{ .CRD.Names.Snake }}
 import (
 	"context"
 	"fmt"
+	ackerr "github.com/aws/aws-controllers-k8s/pkg/errors"
+    corev1 "k8s.io/api/core/v1"
```
jaypipes, line 9: femto-nit, BTW: this is indented wrong (make sure you replace four spaces with a TAB character in templates :) )

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Updated the template.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Updated code template.

templates/pkg/crd_sdk.go.tpl
kumargauravsharma: Updated code template to not add condition when not required.

services/elasticache/pkg/resource/replication_group/custom_set_conditions.go
kumargauravsharma: This scenario is when a successful sdk method call response includes a property (ko.Status.Status), which help determine if the resource is in terminal state. Thus, needed to have it as a custom method.

templates/pkg/crd_manager.go.tpl
kumargauravsharma: done

jaypipes (approved):
++ good to go.

Issue #413, opened 2020-10-14
replication group custom modify logic


kumargauravsharma:
**Describe the bug**
    1. Updated logic to handle order of operations when
       failover, multi az are present in desired Spec
    2. Updated logic for Increase, Decrease replicas
       payload to include only those Node Groups
       from desired Spec that have matching nodeGroupId
       present on server
**Steps to reproduce**
   1. Apply replication group yaml with failover false and node group configurations
   2. Apply replication group yaml with Increase/Decrease replicas scenario
       and add node group id which is not present yet on server.
**Expected outcome**
A concise description of what you expected to happen.

**Environment**

* Kubernetes version
* Using EKS (yes/no), if so version? 
* AWS service targeted (S3, RDS, etc.) Elasticache

jaypipes:
@kumargauravsharma hey, quick request... could you update the bug description here to indicate better what the bug is? From what I can understand from the PR, the bug is that if either the automatic failover or multAZ fields are set to false, then we are improperly calling the update replica count or update shard configuration APIs before calling the "normal" ModifyReplicationGroup API. Is that correct? If so, please update the description here :)

kumargauravsharma:
@jaypipes I have updated the defect description.

PR #414, opened 2020-10-14
replication group custom modify logic fixes

kumargauravsharma:
1. Updated logic to handle order of operations when
   failover, multi az are present in desired Spec
2. Updated logic for Increase, Decrease replicas
   payload to include only those Node Groups
   from desired Spec that have matching nodeGroupId
   present on server

Issue #413 

Description of changes:

### DIFF
--- a/services/elasticache/pkg/resource/replication_group/custom_update_api.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_update_api.go
@@ -42,16 +42,34 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 	}
 
 	// Order of operations when diffs map to multiple updates APIs:
-	// 1. updateReplicaCount() is invoked Before updateShardConfiguration()
-	//	  because both accept availability zones, however the number of
-	//	  values depend on replica count.
+	// 1. When automaticFailoverEnabled differs:
+	//		if automaticFailoverEnabled == false; do nothing in this custom logic, let the modify execute first.
+	// 		else if automaticFailoverEnabled == true then following logic should execute first.
+	// 2. When multiAZ differs
+	// 		if multiAZ = true  then below is fine.
+	// 		else if multiAZ = false ; do nothing in custom logic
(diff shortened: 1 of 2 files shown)


nmvk (approved):
LGTM

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -300,13 +344,35 @@ func (rm *resourceManager) newDecreaseReplicaCountRequestPayload(
...
+			continue
+		}
+		latestReplicaCount := 0
+		if latestShard.NodeGroupMembers != nil {
+			if len(latestShard.NodeGroupMembers) > 0 {
+				latestReplicaCount = len(latestShard.NodeGroupMembers) - 1
+			}
+		}
+		latestReplicaCounts[*latestShard.NodeGroupID] = latestReplicaCount
+	}
```
jaypipes, line 361: suggestion for a future cleanup patch... you could make a little helper function that has the above code so you don't need to repeat it from `newIncreaseReplicaCountRequestPayload`.

services/elasticache/pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -410,7 +410,8 @@ func TestNewIncreaseReplicaCountRequestPayload(t *testing.T) {
 	// Tests
 	t.Run("EmptyPayload=NoSpec", func(t *testing.T) {
 		desired := provideResource()
-		payload, err := rm.newIncreaseReplicaCountRequestPayload(desired)
+		latest := provideResource()
```
jaypipes, line 413: I was initially a little confused by this because the name of the function `provideResource` seems to indicate that some filled-in `resource` is returned, but actually it's an *empty* `resource`.

So, might be good to rename the `providerResource` function to `emptyResource` or similar :)

jaypipes (approved):
Code looks good. Added a few suggestions you can address in a future PR if you'd like.

PR #416, opened 2020-10-16
Ignore CRs `owner-account-id` annotation

A-Hilaly:
Issue N/A

Description of changes:
- Currently service controllers allow overriding the AWS Account ID by specifying a the `owner-account-id` annotation in the CR object. This commit disable this feature.  AWS Account ID should be specified in the annotations of the namespace in which the CR is created.


jaypipes (approved):
:+1: never more shall this be spoken of. :P

Issue #205, opened 2020-08-18
SQS service controller


tabern:
## New ACK service controller

Support for Amazon SQS.

### List of API resources

List the API resources in order of importance to you:

* [Queue](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html)

sergey-telpuk:
Hello, any updates?

jaypipes:
Apologies @sergey-telpuk, but SQS has lagged behind :( I hope to get back to working on it this quarter.

sergey-telpuk:
@jaypipes Hello, thanks for answer. I really want to start using SQS with help of this operator. I'll be waiting for work done

paulermo:
@jaypipes we will wait too, hope things will go well and coding wil be fast:)

satdeveloping:
@jaypipes Do you have a roadmap/list of issues that the community can work on? A lot of us are looking for a well supported SQS operator.

RedbackThomson:
> @jaypipes Do you have a roadmap/list of issues that the community can work on? A lot of us are looking for a well supported SQS operator.

The SQS controller needs to be updated in the form of:
- Updating the runtime from a very old version ([`v0.0.6`](https://github.com/aws-controllers-k8s/sqs-controller/blob/main/go.mod#L6)) to the latest version, and re-generating to fit that runtime
- Updating to Go v1.17
- Adding tests for the `Queue` resource

jaypipes:
Unfortunately, the code generator needs to be refreshed to work with attribute-based field definitions like SNS and SQS. This refresh work is not easy and I need to find time to get it done in the next couple sprint cycles as Bryan works on refactoring other parts of the code generator.

dlgordon:
Hello

I spent some time this weekend having a look at the SQS controller and made some progress - interested if a) I'm on the right track and b) this is useful effort. I know SQS is in progress as per the roadmap.

Anyway, I was able to upgrade the SQS Controller to the latest packages, update the generator yaml (https://github.com/dlgordon/sqs-controller/blob/update-sqs-controller/generator.yaml) and generate a new controller which works in the Create and Delete cases. Far as I can see, my branch mostly works with the following exceptions

[...]

PR #418, opened 2020-10-18
allow override of GetAttributes input shape

jaypipes:
The SQS GetQueueAttributes API call is entirely different from the SNS
GetTopicAttributes call. Whereas the latter is called and returns the
Topic's attributes, the former has a required `[]*string` field on the
Input shape called `AttributeNames` that has to have a single "All"
element in it in order for the API call not to return an error about
malformed Input data :(

This patch introduces new structures in the generator configuration that
allows the code generator to write override values for the Input shape
in order to populate this properly.

Issue #205


A-Hilaly (approved):
/lgtm 👍

PR #426, opened 2020-10-21
add --log-level to controller config options

jaypipes:
The zap logger configuration doesn't apparently work when
--enable-development-logging is true and --log-level is info. In order
to get debug logging, we need to set --log-level = debug. This patch
changes the deployment manifest YAML template along with the
kind-build-test.sh script to pass in the log level environment variable
appropriately, which gets the debug logging output working properly.

I also modified the assert_controller_not_restarted Bash helper function
to output the kubectl logs from previously-running controller pods.


nithu0115 (approved):
👻

scripts/lib/testutil.sh
```diff
@@ -66,6 +66,8 @@ assert_pod_not_restarted() {
     restartCount=$( kubectl get pods -n "$__ns" "$__pod_id" --output jsonpath='{.status.containerStatuses[0].restartCount}' )
     if [ "$restartCount" != "0" ]; then
         echo "FAIL: Expected pod $__pod_id to not have been restarted but it has been restarted $restartCount times."
+        echo "****************************** logs from previous controller pod ************************************"
+        kubectl logs -n ack-system --previous "$__pod_id"
```
A-Hilaly, line 70: 👍 👍

A-Hilaly (approved):

Issue #427, opened 2020-10-22
Replication group create leads to invalid memory address or nil pointer dereference error


kumargauravsharma:
**Describe the bug**

Replication group create leads to invalid memory address or nil pointer dereference error

**Steps to reproduce**

Create a replication group using sample yaml as:
```
apiVersion: elasticache.services.k8s.aws/v1alpha1
kind: ReplicationGroup
metadata:
  name: test
spec:
    engine: redis 
    replicationGroupID: test
    replicationGroupDescription: test
    multiAZEnabled: false
    cacheNodeType: cache.t3.micro
    numNodeGroups: 3
    atRestEncryptionEnabled: true
    transitEncryptionEnabled: true
    cacheSubnetGroupName: default
    nodeGroupConfiguration:
      - 
        nodeGroupID: "0020"
        primaryAvailabilityZone: us-east-1a
        replicaAvailabilityZones:
          - us-east-1b
          - us-east-1c
        replicaCount: 2
      - 
        nodeGroupID: "0090"
        primaryAvailabilityZone: us-east-1b
        replicaAvailabilityZones:
          - us-east-1a
        replicaCount: 1
      - 
        nodeGroupID: "0290"
        primaryAvailabilityZone: us-east-1b
        replicaAvailabilityZones:
          - us-east-1a
          - us-east-1b
        replicaCount: 2
```

Observe logs with error:
```
runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 364 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic (http://k8s.io/apimachinery/pkg/util/runtime.logPanic)(0x223aee0, 0x2f91f60)
...
...
github.com/aws/aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/resource.go:50 (http://github.com/aws/aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/resource.go:50) +0x5
github.com/aws/aws-controllers-k8s/pkg/runtime.(reconciler).updateCRStatus(0xc0005f1290, 0x2643200, 0xc000042050, 0x264b660, 0xc0005a0128, 0x264b660, 0x0, 0xc000587410, 0x264b660)
```

**Expected outcome**
Create replication group to succeed with no nil pointer error.

**Environment**

* Kubernetes version
* Using EKS (yes/no), if so version? no
[...]

PR #428, opened 2020-10-22
Fix nil pointer dereference inside reconciler when manager onError returns nil

kumargauravsharma:
Issue #427 

Description of changes:
Changed the return type of onError and onSuccess methods from:
`(*resource, error)` to `(acktypes.AWSResource, error)` for nil check to work inside reconciler.

Inside reconciler, the value is received into `var latest acktypes.AWSResource` 
and it failed to have the expected execution of 
```
if latest != nil {
```
condition when manager.go `onError()` method return `nil`.
Changing the return type to `(acktypes.AWSResource, error)` resolves the issue.

### DIFF
--- a/templates/pkg/crd_manager.go.tpl
+++ b/templates/pkg/crd_manager.go.tpl
@@ -176,7 +176,7 @@ func newResourceManager(
 func (rm *resourceManager) onError(
 	r *resource,
 	err error,
-) (*resource, error) {
+) (acktypes.AWSResource, error) {
 	r1, updated := rm.updateConditions(r, err)
 	if !updated {
 		return nil, err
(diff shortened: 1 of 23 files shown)


templates/pkg/crd_manager.go.tpl
```diff
@@ -176,7 +176,7 @@ func newResourceManager(
 func (rm *resourceManager) onError(
 	r *resource,
 	err error,
-) (*resource, error) {
+) (acktypes.AWSResource, error) {
 	r1, updated := rm.updateConditions(r, err)
 	if !updated {
 		return nil, err
```
jaypipes, line 182: actually the problem is here. This needs to return `r, err` not `nil, err`

jaypipes (changes requested):
pretty sure this isn't going to fix the source of the error...

jaypipes (approved):
Kumar walked me through this in the debugger... absolutely correct. the issue is that the *resource is an unexported type and the interface AWSResource can actually be evaluated by the if nil guard..

PR #429, opened 2020-10-22
Replication Group multi az and auto failover status values handling during update

kumargauravsharma:
Issue #413 

Description of changes:

Code updates to compare the multi az enabled, auto failover enabled from Spec with the values from Status.
Also the values from Status are not boolean, thus determined true/false values based on string comparison. 


nmvk (approved):
LGTM

jaypipes (approved):
I think eventually I'd like to get to the point where we use the generatorConfig to identify "field pairs" like this where the Spec field and the Status field are essentially the same thing...

PR #437, opened 2020-10-23
simplify SDK templates

jaypipes:
The primary `templates/pkg/crd_sdk.go.tpl` was getting a little long in the
tooth and barely readable. This patch cleans up this  top-level template by
separating sdkFind implementations into their own sub-templates.

In addition, we rename the templates that began with `templates/pkg/crd_`
by moving them into a new `templates/pkg/resource/` directory, aligning
the template name with the actual output Go file that is produced from the
template.


A-Hilaly (approved):
/lgtm 👍

nithu0115 (approved):
🎃

PR #441, opened 2020-10-25
Elasticache Snapshot Support

nmvk:
Description of changes:
* Add addition field to spec from CopySnapShot
* Snapshot controller auto generated code
* Terminal conditions and custom create/update implementations

Will add end to end tests in separate PR.


services/elasticache/apis/v1alpha1/snapshot.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// SnapshotSpec defines the desired state of Snapshot
+type SnapshotSpec struct {
+	CacheClusterID     *string `json:"cacheClusterID,omitempty"`
+	KMSKeyID           *string `json:"kmsKeyID,omitempty"`
+	ReplicationGroupID *string `json:"replicationGroupID,omitempty"`
+	// +kubebuilder:validation:Required
+	SnapshotName       *string `json:"snapshotName"`
+	SourceSnapshotName *string `json:"sourceSnapshotName,omitempty"`
```
jaypipes, line 30: :+1: 

I think eventually I'd like to inject some code comments here for "additional spec fields" to indicate that these are fields that come from non-standard-create operations. I'd also like the ability to inject some docstring here to help inform the Kubernetes user of these CRDs how to use these fields.

pkg/generate/config/config.go
```diff
@@ -105,6 +105,9 @@ type ResourceConfig struct {
 	// `resourceManager` struct that will set Conditions on a `resource` struct
 	// depending on the status of the resource.
 	UpdateConditionsCustomMethodName string `json:"update_conditions_custom_method_name,omitempty"`
+	// Additional spec fields that are required for the resources. These can be other operation
+	// and the member name of those operations.
+	AdditionalSpec map[string]*[]string `json:"additional_spec"`
```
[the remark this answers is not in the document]
jaypipes: After reading through this patch, I actually think a different structure should be used to contain information about these override spec fields. What do you think about this instead?

```go

type ResourceConfig struct {
...
    // SpecFields is a lsit of instructions about additional Spec fields
    // on this Resource
    SpecFields []*SpecFieldConfig `json:"spec_fields"`
}

// SpecFieldConfig instructs the code generator how to handle an additional
// field in the Resource's SpecFields collection. This additional field can source
// its value from a shape in a different API Operation.
type SpecFieldConfig struct {
    // OperationID refers to the ID of the API Operation where we will
    // determine the field's Go type.
    OperationID string `json:"operation_id,omitempty"`
    // MemberName refers to the name of the member of the
    // Input shape in the Operation identified by OperaitonID that
    // we will take as our additional spec field.
    MemberName string `json:"member_name"`
}
```

I think the above structure would be more flexible and allow us to adapt to future wrinkles in the various APIs. Thoughts?

services/elasticache/pkg/resource/snapshot/custom_create_api.go
```diff
@@ -0,0 +1,197 @@
...
+		}
+		if resp.Snapshot.VpcId != nil {
+			ko.Status.VPCID = resp.Snapshot.VpcId
+		}
+
+		rm.setStatusDefaults(ko)
+		// custom set output from response
+		rm.CustomCopySnapshotSetOutput(r, resp, ko)
+		return &resource{ko}, nil
+	}
```
jaypipes, line 173: where is the implementation when SourceSnapshotName is nil?

services/elasticache/pkg/resource/snapshot/custom_update_api.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Snapshot API has no update
+func (rm *resourceManager) customUpdateSnapshot(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	diffReporter *ackcompare.Reporter,
+) (*resource, error) {
+	return desired, nil
+}
```
jaypipes, line 16: Is the above necessary? Is it not possible to stick with the default implementation fallthrough that returns `nil, ackerr.NotImplemented` when there is no Update operation for a resource?

jaypipes (changes requested):
Thanks @nmvk! Please see inline for a number of suggestions and questions :)

pkg/generate/config/config.go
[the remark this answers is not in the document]
nmvk: Agreed updated

pkg/generate/elasticache_test.go
```diff
@@ -738,3 +738,18 @@ func TestElasticache_Override_Values(t *testing.T) {
...
+	g := testutil.NewGeneratorForService(t, "elasticache")
+	crds, err := g.GetCRDs()
+
+	require.Nil(err)
+
+	crd := getCRDByName("Snapshot", crds)
+	require.NotNil(crd)
+
+	assert := assert.New(t)
+	assert.NotNil(crd.SpecFields["SourceSnapshotName"])
```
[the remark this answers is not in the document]
nmvk: Done

pkg/generate/generator.go
```diff
@@ -102,6 +102,20 @@ func (g *Generator) GetCRDs() ([]*ackmodel.CRD, error) {
 			crd.AddSpecField(memberNames, memberShapeRef)
 		}
 
+		// Now add the additional Spec fields that are required from other API
```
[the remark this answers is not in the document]
nmvk: Thanks, fixed.

pkg/model/sdk_helper.go
```diff
@@ -164,6 +164,21 @@ func (a *SDKAPI) GetOperationMap() *OperationMap {
 	return &opMap
 }
 
+// Given an API operation and member of API operation, return shape reference associated with the member
+func (a *SDKAPI) GetMemberShapeRef(operation string, memberName string) (bool, *awssdkmodel.ShapeRef) {
```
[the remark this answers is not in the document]
nmvk: Fixed.

services/elasticache/pkg/resource/snapshot/custom_create_api.go
nmvk: We would use the auto generated implementation if SourceSnapshotName is nil.

services/elasticache/pkg/resource/snapshot/custom_update_api.go
nmvk: After creating snapshot from source snapshot we would populate `ReplicationGroupId` thus ACK would keep on invoking `sdkUpdate` and this keeps throwing `ackerr.NotImplemented`. Since there is no update snapshot API I think it should not be `ackerr.NotImplemented`?

services/elasticache/apis/v1alpha1/snapshot.go
nmvk: Agreed on this I will add this as part of https://github.com/aws/aws-controllers-k8s/issues/447

services/elasticache/pkg/resource/snapshot/custom_update_api.go
jaypipes: I don't understand why ACK would keep on invoking `sdkUpdate`. If `ReplicationGroupId` is returned from the call to `CreateSnapshot`, then the `Status.ReplicationGroupID` field of the CR should be set in the `GoCodeSetOutput` returned Go code, right? And that CR is then saved to etcd and there would not be a new reconcile loop for the snapshot CR.

services/elasticache/pkg/resource/cache_subnet_group/sdk.go
```diff
@@ -133,6 +133,7 @@ func (rm *resourceManager) sdkCreate(
 	ctx context.Context,
 	r *resource,
 ) (*resource, error) {
+
```
jaypipes, line 136: Can you remove these two code changes please? I think you may have just added a newline into the sdk template?

services/elasticache/pkg/resource/snapshot/custom_create_api.go
jaypipes: Yes, you are correct... I forgot there are two ways of overriding custom stuff. There is the complete override of the custom update method name and then there is the technique of overriding the "implementation" that occurs before the default code paths are taken. We definitely need to clean this up. It's way too confusing. :(

services/elasticache/pkg/resource/snapshot/custom_set_output.go
```diff
@@ -0,0 +1,89 @@
...
+	return ko
+}
+
+func (rm *resourceManager) CustomCreateSnapshotSetOutput(
+	r *resource,
+	resp *elasticache.CreateSnapshotOutput,
+	ko *svcapitypes.Snapshot,
+) *svcapitypes.Snapshot {
+	rm.customSetOutput(r, resp.Snapshot, ko)
+	return ko
```
jaypipes, line 42: Here, why don't we set `Spec.ReplicationGroupID`?

templates/pkg/resource/sdk.go.tpl
```diff
@@ -49,6 +49,13 @@ func (rm *resourceManager) sdkCreate(
 	ctx context.Context,
 	r *resource,
 ) (*resource, error) {
+{{ $customMethod := .CRD.GetCustomImplementation .CRD.Ops.Create }}
```
jaypipes, line 52: Use `{{- ... -}}` here to remove the empty newline being generated

templates/pkg/resource/sdk.go.tpl
```diff
@@ -49,6 +49,13 @@ func (rm *resourceManager) sdkCreate(
...
+{{ if $customMethod }}
+	customResp, customRespErr := rm.{{ $customMethod }}(ctx, r)
+	if customResp != nil || customRespErr != nil {
+		return customResp, customRespErr
+	}
+{{ end }}
```
jaypipes, line 58: And here, use:

```
{{- end }}
```

templates/pkg/resource/sdk.go.tpl
```diff
@@ -49,6 +49,13 @@ func (rm *resourceManager) sdkCreate(
 	ctx context.Context,
 	r *resource,
 ) (*resource, error) {
+{{ $customMethod := .CRD.GetCustomImplementation .CRD.Ops.Create }}
+{{ if $customMethod }}
```
jaypipes, line 53: And here, use:

```
{{- if $customMethod }}
```

kumargauravsharma:
Related enhancement request issue #455

services/elasticache/pkg/resource/snapshot/custom_update_api.go
nmvk: `Spec` is not patched so `desired` will not have `ReplicationGroupId` updated thus we see `sdkUpdate` invoked continuously.

services/elasticache/pkg/resource/snapshot/custom_set_output.go
nmvk: Thanks! updated to set `Spec.ReplicationGroupID`

templates/pkg/resource/sdk.go.tpl
nmvk: Done

templates/pkg/resource/sdk.go.tpl
nmvk: Done

templates/pkg/resource/sdk.go.tpl
nmvk: Done

pkg/generate/config/config.go
```diff
@@ -111,6 +111,23 @@ type ResourceConfig struct {
...
+
+// SpecFieldConfig instructs the code generator how to handle an additional
+// field in the Resource's SpecFields collection. This additional field can source
+// its value from a shape in a different API Operation.
+type SpecFieldConfig struct {
+	// OperationID refers to the ID of the API Operation where we will
+	// determine the field's Go type.
+	OperationID string `json:"operation_id,omitempty"`
+	// MemberName refers to the name of the member of the
+	// Input shape in the Operation identified by OperaitonID that
```
jaypipes, line 128: s/OperaitonID/OperationID/

pkg/generate/config/config.go
```diff
@@ -321,6 +338,19 @@ func (c *Config) OverrideValues(operationName string) (map[string]string, bool)
 	return oConfig.OverrideValues, ok
 }
 
+// AdditionSpec gives map of operation and their MemberFields to
```
jaypipes, line 341: s/AdditionSpec/SpecFieldConfig/

(plus, the description of what this method does is incorrect :) )

pkg/generate/testdata/models/apis/elasticache/0000-00-00/generator.yaml
```diff
@@ -1,4 +1,9 @@
 resources:
+  Snapshot:
+    spec_fields:
+        -
```
jaypipes, line 4: FYI, no need for the extra line here. You could just do this:

```yaml
    - operation_id: CopySnapshot
      member_name: SourceSnapshotName
```

services/elasticache/generator.yaml
```diff
@@ -10,6 +10,19 @@ resources:
...
         - InvalidParameterCombination
+  Snapshot:
+    update_conditions_custom_method_name: CustomUpdateConditions
+    exceptions:
+      terminal_codes:
+        - InvalidParameter
+        - InvalidParameterValue
+        - InvalidParameterCombination
+    spec_fields:
+          -
```
jaypipes, line 21: similar comment from above about no need for newline :)

jaypipes (approved):
++ thanks @nmvk :)

PR #443, opened 2020-10-26
Additional fields with defaults to generate.Config object

muvaf:
Description of changes: As part of the ongoing [effort](https://github.com/crossplane/crossplane/issues/1845) about generating Crossplane AWS Provider, there are some small changes where ACK and Crossplane patterns diverge. In CP, we have all the provider-related fields under `spec.forProvider` so that it's separate from CP machinery fields. Same with status, we have `status.atProvider` to differentiate them from other fields. For details, you can see [Managed Resources API Patterns Design Doc](https://github.com/crossplane/crossplane/blob/37013c6/design/one-pager-managed-resource-api-design.md).

I tried to make the change as least intrusive as possible to the existing mechanisms. Please let me know if you think it could be done in a better way.

### DIFF
--- a/pkg/generate/generator.go
+++ b/pkg/generate/generator.go
@@ -361,14 +361,11 @@ func New(
 	apiVersion string,
 	configPath string,
 	templateBasePath string,
+	defaultConfig ackgenconfig.Config,
 ) (*Generator, error) {
-	var gc *ackgenconfig.Config
-	var err error
-	if configPath != "" {
-		gc, err = ackgenconfig.New(configPath)
-		if err != nil {
-			return nil, err
-		}
+	gc, err := ackgenconfig.New(configPath, defaultConfig)
+	if err != nil {
+		return nil, err
 	}
 
 	return &Generator{
@@ -378,6 +375,6 @@ func New(
 		serviceAlias:     SDKAPI.ServiceID(),
 		apiVersion:       apiVersion,
 		templateBasePath: templateBasePath,
-		cfg:              gc,
+		cfg:              &gc,
 	}, nil
 }
--- a/pkg/model/crd.go
+++ b/pkg/model/crd.go
@@ -315,6 +315,9 @@ func (r *CRD) UnpackAttributes() {
 // IsPrimaryARNField returns true if the supplied field name is likely the resource's
 // ARN identifier field.
 func (r *CRD) IsPrimaryARNField(fieldName string) bool {
+	if r.genCfg != nil && !r.genCfg.IncludeACKMetadata {
+		return false
+	}
 	return strings.EqualFold(fieldName, "arn") ||
 		strings.EqualFold(fieldName, r.Names.Original+"arn")
 }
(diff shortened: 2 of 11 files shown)


pkg/model/crd.go
```diff
@@ -732,14 +734,14 @@ func (r *CRD) GoCodeSetInput(
 		sourceAdaptedVarName := sourceVarName
 		crdField, found = r.SpecFields[renamedName]
 		if found {
-			sourceAdaptedVarName += ".Spec"
+			sourceAdaptedVarName += r.StringStore[SpecFieldPath]
```
jaypipes, line 737: Instead of the `map[string]string` here, if you went with the `PrefixConfig` struct I suggest above, you could have little helper methods on `CRD` like so:

```go

func (crd *CRD) specFieldPrefix() string {
    if crd.cfg == nil || crd.cfg.Prefix == nil {
        return ".Spec"
    }
    return c.Prefix.SpecField
}

func (crd *CRD) statusFieldPrefix() string {
    if crd.cfg == nil || crd.cfg.Prefix == nil {
        return ".Status"
    }
    return c.Prefix.StatusField
}
```

jaypipes (changes requested):
Hi @muvaf! :) Left a note inline about using the `pkg/generate/config.Config` struct for this kind of information...

pkg/generate/generator.go
```diff
@@ -38,7 +38,8 @@ type Generator struct {
 	typeRenames      map[string]string
 	// Instructions to the code generator how to handle the API and its
 	// resources
-	cfg *ackgenconfig.Config
+	cfg         *ackgenconfig.Config
+	stringStore map[ackmodel.StringIdentifier]string
```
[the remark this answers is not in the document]
muvaf: Hmm. As it currently stands, the list of strings is here https://github.com/aws/aws-controllers-k8s/pull/449/files#diff-bbfadd4328f26a3a50306ade20809a5d15a16a0803dc14d85463e0cce64cb4eaR27

I was kind of hesitant about whether it'd scale to have all custom strings strongly typed in the structs because the list seems to be growing and having them hard-coded works for both ack and crossplane. But making them part of `Config` makes sense, too, in order to have some per-resource flexibility. I think we can have the current set of strings as hard-coded default for `Config` object and overwrite what we read from `generator.yaml` (if any) onto that `Config` object. Let me know what you think after looking at the current list and I'll make the change.

pkg/model/crd.go
muvaf: I was thinking that for both ack and crossplane, we'd have a hard-coded default `Config` object and use the `generator.yaml` as an overlay on top of that `Config` so we'd make sure there is always a value in those fields.

pkg/model/crd.go
```diff
@@ -1855,16 +1858,6 @@ func (r *CRD) goCodeSetOutputReadMany(
 	// operation by checking for matching values in these fields.
 	matchFieldNames := r.listOpMatchFieldNames()
 
-	//  if len(resp.CacheClusters) == 0 {
```
muvaf, line 1858: If `len(resp.CacheClusters) == 0`, the `for` loop won't run at all, meaning `found` will be `false` and there is a return statement after the `for` loop that returns the same answer if `found == false`. So, behavior shouldn't change when we remove this if block.

pkg/model/crd.go
```diff
@@ -315,6 +315,9 @@ func (r *CRD) UnpackAttributes() {
 // IsPrimaryARNField returns true if the supplied field name is likely the resource's
 // ARN identifier field.
 func (r *CRD) IsPrimaryARNField(fieldName string) bool {
+	if r.genCfg != nil && !r.genCfg.IncludeACKMetadata {
```
muvaf, line 318: @jaypipes since we know that it will always have some value, what would you think about making `r.genCfg` a value variable instead of pointer? We wouldn't need to make if `r.genCfg != nil` check everywhere and run into risk of panicing in places we don't.

muvaf (commented):
@jaypipes Now they are part of the config object except for the `APIGroupSuffix` because that one is processed in `SDKAPI` object. I plan to change its value in `cmd/crossplane.go` after it's initialized. For ACK, nothing really changes.

jaypipes (approved):
Awesomeness. You rock @muvaf :)

PR #446, opened 2020-10-27
Allow override of operation types and resource names for API operations

### LINKED ISSUE #391: Code generator does not identify top-level resources (CRDs) for ApplicationAutoscaling: API Operations begin with "Register" and "Put"
**Describe the bug**
ACK for the application autoscaling service does not generate resources (config/crd/bases) since the code generator identifies top-level resources (CRDs) by looking for API Operations that begin with "Create". The Application Autoscaling API uses `RegisterScalableTarget` and `PutScalingPolicy` instead. 

**Steps to reproduce**
1. `make build-controller SERVICE=applicationautoscaling`
2. `./scripts/kind-build-test.sh -s applicationautoscaling`
The second command fails with the error `applicationautoscaling/config/crd/bases does not exist`

**Expected outcome**
The controller manager docker image be build to install the service operator. 

**Environment**

* Kubernetes version
* Using EKS (yes/no), if so version? No, using the existing scripts on Kind. 
* AWS service targeted (S3, RDS, etc.) - ApplicationAutoscaling (For SageMaker)

zacharya:
Fixes #391 

Description of changes: This change allows for the ability to provide operation type and resource name overrides for API operations that are not correctly classified based on the heuristics logic.  For example, an API operation name that begins with `Put..` is not currently classified as a `Create` operation.

Example config:
```
ignore:
  ...
operations:
  PutScalingPolicy:
    operation_type: Create
    resource_name: ScalingPolicy
resources:
  ...
```

This PR also subpackaged `operations` in the model package to avoid cyclic dependency issues.


pkg/model/operations/op.go
```diff
@@ -20,23 +20,23 @@ import (
...
+	OpTypeReplace = "Replace"
+	OpTypeUpdate = "Update"
+	OpTypeAddChild = "AddChild"
+	OpTypeAddChildren = "AddChildren"
+	OpTypeRemoveChild = "RemoveChild"
+	OpTypeRemoveChildren = "RemoveChildren"
+	OpTypeGet = "Get"
+	OpTypeList = "List"
+	OpTypeGetAttributes = "GetAttributes"
+	OpTypeSetAttributes = "SetAttributes"
```
jaypipes, line 39: I'm a little confused about this... any reason for it?

jaypipes (changes requested):
Thank you @zacharya! Great to get contribution on this :)

Can I ask that you undo the changes that add a `pkg/model/operations` package, though?

The meat of your patch looks good; it's just a bit buried by the addition of the new operations package. Is there a specific reason you added that new package?

Also, helpful hint: enable the plugin in IntelliJ IDE that automatically runs `go fmt` on Go source files for you (it would automatically catch a number of formatting things...)

pkg/generate/config/config.go
```diff
@@ -43,6 +44,12 @@ type OperationConfig struct {
 	// `resourceManager` struct that will set fields on a `resource` struct
 	// depending on the output of the operation.
 	SetOutputCustomMethodName string `json:"set_output_custom_method_name,omitempty"`
+	// Override for resource name in case of heuristic failure
+	// An example of this is correcting stutter when the resource logic doesn't properly determine the resource name
+	ResourceName  string            `json:"resource_name"`
+	// Override for operation type in case of heuristic failure
+	// An example of this is `Put...` or `Register...` API operations not being correctly classified as `Create` op type
+	OperationType operations.OpType `json:"operation_type"`
```
jaypipes, line 52: > This PR also subpackaged `operations` in the model package to avoid cyclic dependency issues.

Hmm. OK. I think there's probably a way to structure the generatorConfig fields without having to create a new package for the operations codes.

You can just use a `string` data type here and do the translation to a `pkg/model.OpType` within the `pkg/model/crd.go` file.

pkg/model/operations/op.go
zacharya: This was originally to allow us to go straight from config JSON to an `OpType` but as discussed below this has been reverted to an `int`.

pkg/generate/config/config.go
zacharya: Switched back to the original package and using a `string` in the config instead.

To make it a bit more reusable, I instead added an `OpTypeFromString(s string)` method to `op.go`.

jaypipes (approved):
Very nice, thank you @zacharya and welcome to the ACK contributors club :)

Issue #239, opened 2020-08-25
SFN service controller


tabern:
## New ACK Service Controller

Support for AWS Step Functions

### List of API resources

List the API resources in order of importance to you:

* [StateMachine](https://docs.aws.amazon.com/step-functions/latest/apireference/API_CreateStateMachine.html)
* [Activity](https://docs.aws.amazon.com/step-functions/latest/apireference/API_CreateActivity.html)

russmackenzie:
Now that the 'sfn' controller and end-to-end test are merged (https://github.com/aws/aws-controllers-k8s/pull/450), should I submit a PR to add Step Functions as 'developer preview' to the Testing and Services pages or are there any other prerequisites?

jaypipes:
@russmackenzie yep. :)

russmackenzie:
I believe this issue can now be closed.

jaypipes:
@russmackenzie we're going to keep these service controller issues open since they are on the Service Controller Roadmap project and provide a good top-level landing page...

mikestef9:
Closing as this service controller has graduated to GA. Separate issues can be opened to discuss specific follow on topics on the controller.

PR #450, opened 2020-10-30
Add sfn controller and smoke test

russmackenzie:
Issue #, if available: https://github.com/aws/aws-controllers-k8s/issues/239

Description of changes: 
* Added `services/sfn/generator.yaml`, which defines the names of 404 exceptions when getting state machines or activities.
* Added `test/e2e/sfn/smoke.sh`. This creates an IAM role (with the `AWSDenyAll` policy), creates a state machine, deletes the state machine, deletes the IAM role, creates an activity, and deletes the activity.
* Added `scripts/lib/aws/sfn.sh`, which adds helper functions used by `smoke.sh`.
* Built Step Functions controller following instructions from https://aws.github.io/aws-controllers-k8s/dev-docs/testing/
```
make build-ack-generate
export SERVICE=sfn
make build-controller SERVICE=$SERVICE
... <set up IAM role>
make kind-test SERVICE=$SERVICE
```
* The smoke test succeeds.


services/sfn/config/iam/recommended-policy-arn
```diff
@@ -0,0 +1 @@
+arn:aws:iam::aws:policy/AWSStepFunctionsFullAccess
```
jaypipes, line 1: If I attach the above managed policy to my IAM role that I'm executing the service controller under, I'm getting the following error in my service controller logs when running the SFN smoke test:

```
AccessDeniedException: User: arn:aws:sts::750630568209:assumed-role/elasticache.services.k8s.aws.full/tmp-role-776639d8 is not authorized to perform: iam:PassRole on resource: arn:aws:iam::750630568209:role/ack-sfn-execution-role
```

Is there a need to also allow `iam:PassRole` permission?

Note that `arn:aws:sts::750630568209:assumed-role/elasticache.services.k8s.aws.full` is the IAM role I'm using for my service controller. Even though it's called "elasticache.services.k8s.aws.full", it has the `AWSStepFunctionsFullAccess` IAM managed policy associated with it.

jaypipes (commented):
Awesome work so far on this @russmackenzie! Question inline about required managed IAM policy for the IAM role executing the SFN service controller, though...

services/sfn/config/iam/recommended-policy-arn
russmackenzie: > Is there a need to also allow iam:PassRole permission?
Yes. Yes there is. 

Any advice on this? The only AWS managed policy that I'm aware of which contains sufficient permissions is `arn:aws:iam::aws:policy/IAMFullAccess`. I'd prefer that we don't suggest users sling that policy around too cavalierly. 

Maybe we could add a second file such as `services/sfn/config/iam/additional-policy` with a script/instructions to attach a policy with the iam:PassRole permissions?

jaypipes: > > Is there a need to also allow iam:PassRole permission?
> > Yes. Yes there is.
> 
> Any advice on this? The only AWS managed policy that I'm aware of which contains sufficient permissions is `arn:aws:iam::aws:policy/IAMFullAccess`. I'd prefer that we don't suggest users sling that policy around too cavalierly.
> 
> Maybe we could add a second file such as `services/sfn/config/iam/additional-policy` with a script/instructions to attach a policy with the iam:PassRole permissions?

Hmm, yeah, that's a bit of a conundrum. I think adding a second file called `additional-policy` would be fine, yeah. If you wouldn't mind adding a quick note to the docs here:

https://github.com/aws/aws-controllers-k8s/blob/main/docs/contents/user-docs/authorization.md#configuring-aws-iam

stating that there may be an `additional-policy` file for the service, that would be great.

Thanks!

russmackenzie: I have added `services/sfn/config/iam/additional-policy` and mentioned it in `docs/contents/user-docs/authorization.md `. Give it another go, please.

jaypipes (approved):
:+1:

jaypipes:
Awesome, thank you @russmackenzie! Welcome StepFunctions to the ACK project :)

Issue #451, opened 2020-10-30
Incorrect error returned from sdkFind()


russmackenzie:
**Describe the bug**
There's an error handling block after calling &lt;ReadOne&gt;WithContext() in sdkFind(). This block checks the variable `respErr`, but returns an unrelated variable `err` instead.

This comes from `templates/pkg/resource/sdk_find_read_one.go.tpl`, line 24.

**Steps to reproduce**
Build a controller, e.g. dynamodb. Look at `services/dynamodb/pkg/resource/table/sdk.go`'s sdkFind() method.
Observe that it returns `err` when it should return `respErr`:
```
if respErr != nil {
	if awsErr, ok := ackerr.AWSError(respErr); ok && awsErr.Code() == "ResourceNotFoundException" {
		return nil, ackerr.NotFound
	}
	return nil, err
}
``` 

The consequence is that if you encounter an error other the one handled by e.g. `awsErr.Code() == "ResourceNotFoundException"`, you'll return what is most likely a nil and it will cause a seg fault later on. I encountered this while building the `sfn` controller, before I defined the name of the 404 exception in my generator.yaml.

**Expected outcome**
It should return `respErr`.

**Environment**
* kind v0.9.0 go1.15.2 darwin/amd64
* AWS service targeted: Step Functions

PR #452, opened 2020-10-30
Fix incorrect error returned from sdkFind()

russmackenzie:
Issue #, if available: https://github.com/aws/aws-controllers-k8s/issues/451

Description of changes: Fix a somewhat minor bug. This error handling block is checking one error variable, but returning an unrelated one. The unrelated one is nil and will cause seg faults elsewhere.


jaypipes (approved):
Nice catch, @russmackenzie thank you! :)

jaypipes (commented):
We will want to regenerate the service controllers that use a ReadOne finder method.

Issue #457, opened 2020-11-02
ACK ElastiCache controller CacheParameterGroup support


kumargauravsharma:
**Is your feature request related to a problem?**
Enable CacheParameterGroup support from ACK service controller for Amazon ElastiCache

**Describe the solution you'd like**
A description of what you want to happen.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

Issue #262, opened 2020-09-01
Enable ResetCacheParameterGroup api on cache parameter group custom resource


kumargauravsharma:
**Is your feature request related to a problem?**
ElastiCache Cache parameter group api: ResetCacheParameterGroup modifies the parameters of a cache parameter group to the engine or system default value. 
Reset specific parameters by submitting a list of parameter names. To reset the entire cache parameter group, specify the ResetAllParameters and CacheParameterGroupName parameters.

Reference: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/elasticache/reset-cache-parameter-group.html

This issue to to enable support for this API on cache parameter group custom resource.

**Describe the solution you'd like**
A description of what you want to happen.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

PR #464, opened 2020-11-05
Elasticache CacheParameterGroup support

kumargauravsharma:
Issue #457
Issue #262

Description of changes:
* Enables `CacheParameterGroup` API for ElastiCache controller.
* Adds support for `status_fields` in generator config (on the similar lines of `spec_fields`)


pkg/generate/config/config.go
```diff
@@ -133,19 +133,22 @@ type ResourceConfig struct {
...
 }
 
-// SpecFieldConfig instructs the code generator how to handle an additional
-// field in the Resource's SpecFields collection. This additional field can source
-// its value from a shape in a different API Operation.
-type SpecFieldConfig struct {
+// AdditionalFieldConfig instructs the code generator how to handle an additional
+// field in the Resource's SpecFields/StatusFields collection. This additional field
+// can source its value from a shape in a different API Operation.
+type AdditionalFieldConfig struct {
```
jaypipes, line 145: eventually, we'll probably want to consolidate the existing `FieldConfig` with this, but this is cool for now :)

pkg/generate/testdata/models/apis/elasticache/0000-00-00/generator.yaml
```diff
@@ -1,9 +1,25 @@
...
+        - InvalidCacheParameterGroupState
+        - InvalidGlobalReplicationGroupState
+        - InvalidParameterCombination
+        - InvalidParameterValue
+    spec_fields:
+      - operation_id: ModifyCacheParameterGroup
+        member_name: ParameterNameValues
+    status_fields:
+      - operation_id: ModifyCacheParameterGroup
+        member_name: ParameterNameValues
```
jaypipes, line 20: We should not have the same-named field in both the Spec and the Status...

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -0,0 +1,225 @@
...
+) ([]*svcapitypes.ParameterNameValue, error) {
+	parameterNameValues := []*svcapitypes.ParameterNameValue{}
+	var paginationMarker *string = nil
+	for {
+		input, err := rm.newDescribeCacheParametersRequestPayload(cacheParameterGroupName, source, paginationMarker)
+		if err != nil {
+			return nil, err
+		}
+		response, respErr := rm.sdkapi.DescribeCacheParametersWithContext(ctx, input)
+		rm.metrics.RecordAPICall("READ_ONE", "DescribeCacheParameters", respErr)
```
jaypipes, line 39: technically this is a READ_MANY call, not a READ_ONE. The resource being read is the cache parameters, not the cache parameter group.

services/elasticache/pkg/resource/cache_parameter_group/custom_update_api.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Implements specialized logic for update CacheParameterGroup.
+func (rm *resourceManager) customUpdateCacheParameterGroup(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	diffReporter *ackcompare.Reporter,
+) (*resource, error) {
+	desiredParameters := desired.ko.Spec.ParameterNameValues
+	latestParameters := latest.ko.Status.ParameterNameValues
```
jaypipes, line 30: This should be `latest.ko.Spec.ParameterNameValues`. There's no need to have ParameterNameValues in both the Spec and Status because `latest` is the CR constructed from the ReadOne operation...

jaypipes (commented):
I think we need to have a conversation about getting a more structured approach to setting `Spec` field values to some default value returned from the AWS service side when the client either nulls out an existing non-nil field (like is the case when the cache parameter values get "reset" here) or when the client sends a nil value for an optional field in the Spec and the server sets it some default value and that value is exposed either in the return from the Create operation or shows up in the next call to the ReadOne operation.

/cc @vijtrip2 since this is something we've previously discussed.

pkg/generate/testdata/models/apis/elasticache/0000-00-00/generator.yaml
kumargauravsharma: removed the field from Status, however kept the changes in generator logic for status field as it is needed for upcoming changes.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Updated the code.

services/elasticache/pkg/resource/cache_parameter_group/custom_update_api.go
kumargauravsharma: Removed the field from status

templates/pkg/resource/sdk.go.tpl
```diff
@@ -73,7 +73,10 @@ func (rm *resourceManager) sdkCreate(
 	rm.setStatusDefaults(ko)
 	{{ if $setOutputCustomMethodName := .CRD.SetOutputCustomMethodName .CRD.Ops.Create }}
 		// custom set output from response
-		rm.{{ $setOutputCustomMethodName }}(r, resp, ko)
+		ko, err = rm.{{ $setOutputCustomMethodName }}(ctx, r, resp, ko)
```
jaypipes, line 76: :+1:

services/elasticache/pkg/resource/cache_parameter_group/custom_update_api.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	if updated {
+		rm.setStatusDefaults(latest.ko)
+		// Populate ko.Spec.ParameterNameValues with latest parameter values
+		source := "user"
+		parameterNameValues, err := rm.describeCacheParameters(ctx, desired.ko.Spec.CacheParameterGroupName, &source)
+		if err != nil {
+			return nil, err
+		}
+		latest.ko.Spec.ParameterNameValues = parameterNameValues
```
jaypipes, line 70: ++

jaypipes (approved):
Great stuff, @kumargauravsharma :)

PR #465, opened 2020-11-06
Add `compare` configuration to support ignoring specific fields when comparing CRs

A-Hilaly:
Issue N/A

Description of changes:
- Adding `Compare` field to the generator configuration, to support custom comparison logic for CR objects.
- Adding `Ignore` field to the Compare configuration, to instruct the generator to ignore certain fields when comparing two different CRs

Configuration example:
```yaml
# lambda/generator.yaml
resources:
  Function:
    exceptions:
      codes:
        404: ResourceNotFoundException
    list_operation:
      match_fields:
        - FunctionName
    compare:
      ignore:
        - 'Spec.Code.s3Bucket'
        - 'Spec.Code.s3Key'
        - 'Status.Code.Location'
        - 'Status.Code.Type'

```

### DIFF
--- a/templates/pkg/resource/descriptor.go.tpl
+++ b/templates/pkg/resource/descriptor.go.tpl
@@ -62,8 +62,15 @@ func (d *resourceDescriptor) Equal(
 ) bool {
 	ac := a.(*resource)
 	bc := b.(*resource)
-	opts := cmpopts.EquateEmpty()
-	return cmp.Equal(ac.ko, bc.ko, opts)
+	opts := []cmp.Option{cmpopts.EquateEmpty()}
+	{{- if .CRD.CompareIgnoredFields }}
+	opts = append(opts, cmpopts.IgnoreFields(ac,
+		{{- range $fieldPath := .CRD.CompareIgnoredFields }}
+		{{ printf "%q" $fieldPath }},
+		{{- end }}
+	))
+	{{- end }}
+	return cmp.Equal(ac.ko, bc.ko, opts...)
 }
 
 // Diff returns a Reporter which provides the difference between two supplied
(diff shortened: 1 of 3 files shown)


pkg/generate/template.go
```diff
@@ -150,7 +150,8 @@ type templateAPIVars struct {
 // code for a single top-level resource's API definition
 type templateCRDVars struct {
 	templateMetaVars
-	CRD *ackmodel.CRD
+	CRD                  *ackmodel.CRD
+	CompareIgnoredFields []string
```
A-Hilaly, line 154: This is working fine, but i'm not sure if it's the best we can do here. Maybe i should hide this logic within the CRD object and use custom template functions to generate the compare code ? /cc @jaypipes

jaypipes: Well, you could add a method to CRD like this:

https://github.com/aws/aws-controllers-k8s/blob/18f42be3be14b2d91363249d198dbca7616e45c2/pkg/model/crd.go#L275-L281

That way you wouldn't need to add any fields to the templateCRDVars.

templates/pkg/resource/descriptor.go.tpl
```diff
@@ -62,8 +62,15 @@ func (d *resourceDescriptor) Equal(
 ) bool {
 	ac := a.(*resource)
 	bc := b.(*resource)
-	opts := cmpopts.EquateEmpty()
-	return cmp.Equal(ac.ko, bc.ko, opts)
+	opts := []cmp.Option{cmpopts.EquateEmpty()}
+	{{- if .CompareIgnoredFields }}
```
jaypipes, line 66: And this would just be:

```
{{- if .CRD.CompareIgnoredFields }}
```

A-Hilaly: done

pkg/generate/template.go
A-Hilaly: Aaaah, TIL - Didn't know functions with `func() bool` signature can be evaluated with the templating library.  Thank you!

A-Hilaly:
@jaypipes Addressed your comments, please take a look

jaypipes (approved):
Merci beaucoup, @A-Hilaly!

PR #473, opened 2020-11-10
Pass a struct value sample to `cmpopts.IgnoreFields` instead of a value pointer

A-Hilaly:
Issue N/A, followup to: #465 

While debugging the lambda controller i found an issue regarding `cmpopts.IgnoreFields` usage (oops 😄 )
 
Description of changes:
Fix `cmpopts` usage by passing a struct value sample to `cmpopts.IgnoreFields` instead of a value pointer


jaypipes (approved):
:+1:

Issue #366, opened 2020-10-02
generate Helm charts for each service controller


jaypipes:
In addition to the `kustomize`-based manual install method currently used by the KinD-based e2e tests, we need to provide Helm charts for easily installing ACK service controllers.

- [x] Push Docker images for built ACK service controllers to the newly-created DockerHub [`amazon/aws-controllers-k8s` Docker repository](https://hub.docker.com/r/amazon/aws-controllers-k8s)
- [x] Generate standardized Helm chart and values.yaml files for each ACK service controller in a new `ack-generate helm` CLI command
- [x] Create an e2e test that verifies the installation of the ACK service controller via Helm into a KinD cluster
- [x] Update the [install and setup documentation](https://github.com/aws/aws-controllers-k8s/blob/main/docs/contents/user-docs/install.md)

red8888:
Is there documentation for this? I want to install aws-controllers-k8s with helm but having trouble finding docs on it

PR #478, opened 2020-11-10
Fix ClusterRole ref name in cluster-role-binding.yml

A-Hilaly:
- Add service account name for controller deployment pod specs
- Use `app.name` instead of `app.fullname` for ClusterRole refName

Issue #366 


jaypipes (approved):
:+1:

Issue #81, opened 2020-07-13
Allow overriding target AWS Region for CRs


jaypipes:
AWS resources are created in a specific region. When calling an AWS service API, a regional API endpoint/URI is used. Similarly, when instantiating an `aws-sdk-go`  `Session`, a regional endpoint is determined by looking at the `AWS_REGION` environs variable.

Currently, ACK CRs are created in the AWS Region associated with the `AWS_REGION` environs variable for the Pod running the ACK service controller. 

We want to support a Kubernetes user being able to specify an AWS Region in which to create a particular resource. I'm thinking about support for a standardized `services.k8s.aws/region` annotation that can be associated with a `Namespace` (providing a default AWS region for CRs within that `Namespace` to use to override the default one used by the service controller) as well as the ability for a Kubernetes user to specify a `service.k8s.aws/region` annotation directly on the CR itself to override the `Namespace` default.

jaypipes:
@A-Hilaly I believe this is already done but we just need a test case demonstrating the functionality. Can you whip one up? The test should be the same AWS account but two CRs targeting two different AWS regions .

PR #499, opened 2020-11-13
Use "accountID/region" as cache key to in the resourceManagerFactory

A-Hilaly:
Issue #81 

Fix an issue that was preventing service controllers from creating multiple resources in different AWS regions (all the resources were created in the first used region) . Essentially because of `resourceManagerFactory` internal cache that was using `accountID` as a key. Changing the key format to `$accountID/$region` guarantees that we will have one unique resourceManager per account per region.

Description of changes:
- `resourceManagerFactory`: Use `$accountID/$region` as a cache key id
- Regenerate all `manager_factory.go` files


templates/pkg/resource/manager_factory.go.tpl
```diff
@@ -38,8 +39,9 @@ func (f *resourceManagerFactory) ManagerFor(
 	id ackv1alpha1.AWSAccountID,
 	region ackv1alpha1.AWSRegion,
 ) (acktypes.AWSResourceManager, error) {
+	rmId := fmt.Sprintf("%s/%s", id, region)
 	f.RLock()
-	rm, found := f.rmCache[id]
+	rm, found := f.rmCache[rmId]
 	f.RUnlock()
 
 	if found {
```
A-Hilaly, line 47: One other "working" solution to this bug could have been to simply check for the region after finding a resourceManager.. But this solution means that we will recreate and override the resourceManager everytime we create a resource in a different region
```golang
if found && region == rm.awsRegion {
    return rm, nil
}
```

jaypipes (approved):
:+1:

Issue #504, opened 2020-11-13
Standardize structure of Bash scripts


jaypipes:
We currently have quite a few Bash scripts in the [`scripts/`](https://github.com/aws/aws-controllers-k8s/tree/main/scripts) directory.

I've tried to keep the calling format and structure of these Bash scripts relatively consistent. However, because some of the scripts have roots in other repositories (aws/aws-node-termination-handler, aws/amazon-vpc-cni-k8s, etc), there are inconsistencies in:

- [ ] whether and how getopts/CLI options vs environment variables to control the script
- [ ] how the `$ROOT_DIR` is determined
- [ ] how variables are defaulted
- [ ] how errors and Bash options are set
- [ ] documentation about scripts is inconsistent and we should have a Bash script writing guide

It would be awesome to clean all these inconsistencies up and have a `scripts/` directory that we can be proud of :)

## `getopts` vs environment variables

Some scripts use `getopts` to process CLI arguments. Here's an example from the `scripts/kind-build-test.sh` script:

https://github.com/aws/aws-controllers-k8s/blob/079fa26fe98aa289e5770e8226129917971473f8/scripts/kind-build-test.sh#L84-L113

Other scripts don't use `getopts` and prefer instead to simply have a fixed number of arguments and rely on environment variables for optional behaviour switches. An example would be the `scripts/helm-package-controller.sh` script:

https://github.com/aws/aws-controllers-k8s/blob/079fa26fe98aa289e5770e8226129917971473f8/scripts/helm-package-controller.sh#L16-L34

My **personal** preference is for the latter environment variable approach. However, if contributors feel strongly about the former, I will bias more towards having a consistent calling structure than my personal preference.

## how the `$ROOT_DIR` is determined

Some scripts use this:

```bash
SCRIPTS_DIR=$(cd "$(dirname "$0")"; pwd)
ROOT_DIR="$SCRIPTS_DIR/.."
```

other scripts use this:

```bash
SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
```

others use this:

```bash
THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```

(that last one is from `test/e2e/run-tests.sh`)

There's also the potential of using this technique as our standard:

https://github.com/aws/eks-charts/blob/d721c97c339c8b1060ac8b76a247ce67a470e90c/scripts/publish-charts.sh#L4

[...]

PR #518, opened 2020-11-20
Set env variables usage instead of getopts

niconosenzo:
Efforts part of Issue #504
Refactoring stage: Argument variable values as env variables instead of getops strategy
Script: **build-controller-image.sh**

Description of changes:
Removed getops strategy in favor of a fixed number of arguments (aws service name in this case) and env variables for the previous values passed through arguments


scripts/build-controller-image.sh
```diff
@@ -17,43 +16,27 @@ check_is_installed docker
...
-      ;;
-    \? )
-        echo "${USAGE}" 1>&2
-        exit
-      ;;
-  esac
-done
+Environment variables:
+  QUIET:                    Build controller container image quietly (<true|false> Default: false)
+  AWS_SERVICE_DOCKER_IMG:   Controller container image tag (Default: ack-<service>-controller:$VERSION)
```
jaypipes, line 29: Let's go ahead and place the `Default: XXXX` part on a separate line, similar to how these are done:

https://github.com/aws/aws-controllers-k8s/blob/1ca9215edc0b5883075712e0dbeb7fd5374a8bd0/scripts/helm-package-controller.sh#L24-L27

jaypipes (changes requested):
teeny little request, see inline :)

scripts/build-controller-image.sh
```diff
@@ -17,43 +16,27 @@ check_is_installed docker
 
 USAGE="
 Usage:
```
A-Hilaly, line 18: One more thing we can take care of in this PR, is changing how this script is called in other files. PTAL at [./kind-build-test.sh](https://github.com/aws/aws-controllers-k8s/blob/main/scripts/kind-build-test.sh#L140-L145) and [./publish-controller-image.sh](https://github.com/aws/aws-controllers-k8s/blob/main/scripts/publish-controller-image.sh#L57-L60)

A-Hilaly (commented):
LGTM overall 👍 . I left few comments :)

scripts/build-controller-image.sh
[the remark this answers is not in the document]
niconosenzo: Understood, then changing DEFAULT_AWS_SERVICE_DOCKER_IMG to "$AWS_SERVICE-$VERSION" as well as the Usage string. Will take care of the default var assignation in a separate PR.

scripts/build-controller-image.sh
niconosenzo: @A-Hilaly  public-controller-image.sh calls this script setting the svc docker image tag to AWS_SERVICE_DOCKER_IMG_TAG, this script expects var name AWS_SERVICE_DOCKER_IMG instead, there seems to be some confusion and the script (public-controller-image.sh) uses both variable names indistinctly, is that correct ?

A-Hilaly: Nice catch. This is correct - `AWS_SERVICE_DOCKER_IMG` is the variable that should be used to call `./build-controller-image.sh`. the other variable `AWS_SERVICE_DOCKER_IMG_TAG` should only be used to construct the variable `AWS_SERVICE_DOCKER_IMG`

A-Hilaly: Something like this makes more sense
```bash
: "${AWS_SERVICE_DOCKER_IMG:="$DOCKER_REPOSITORY:$DEFAULT_AWS_SERVICE_DOCKER_IMG_TAG"}"

$SCRIPTS_DIR/build-controller-image.sh -s "$AWS_SERVICE" -i "$AWS_SERVICE_DOCKER_IMG"
```

A-Hilaly (approved):
/lgtm thanks @niconosenzo!

jaypipes (approved):
:+1: thanks @niconosenzo! Welcome to the ACK contributor club :)

PR #522, opened 2020-11-20
Set env variables usage instead of getopts

niconosenzo:
Efforts part of Issue #504
Refactoring stage: Argument variable values as env variables instead of getops strategy
Script: **build-controllers.sh**

Description of changes:
Removed getops strategy in favor of a fixed number of arguments (list of services in this case) and env variables for the previous values passed through arguments

Note: there is no reference to this script among others in the same folder.


jaypipes (commented):
Actually, I think I'd rather just remove the `scripts/build-controllers.sh` script entirely.

It can be replaced entirely with:

```bash
for SERVICE in s3 ecr elasticache; do ./scripts/build-controller.sh $SERVICE; done
```

@A-Hilaly what do you think? Get rid of this script?

A-Hilaly:
@jaypipes Yep agreed! We don't use this script at all. Let's get rid of it.

niconosenzo:
@jaypipes @A-Hilaly  deleted the file before continuing with the refact tasks.

A-Hilaly (approved):

Issue #529, opened 2020-11-25
Defer modify replication group if the replication group or any of its node groups is not in available state


kumargauravsharma:
**Describe the bug**
Defer modify replication group API call if the replication group or any of its node groups is not in available state.

The code currently does not check node group status.

**Steps to reproduce**
Added unit test inside `aws-controllers-k8s/services/elasticache/pkg/resource/replication_group/custom_update_api_test.go` 
```
func TestCustomModifyReplicationGroup_NodeGroup_Unvailable(t *testing.T) {
	assert := assert.New(t)
	// Setup
	rm := provideResourceManager()
	// Tests
	t.Run("UnavailableNodeGroup=Requeue", func(t *testing.T) {
		desired := provideResource()
		latest := provideResource()
		latest.ko.Status.NodeGroups = provideNodeGroups("1001")
		unavailableStatus := "modifying"
		for _, nodeGroup := range latest.ko.Status.NodeGroups {
			nodeGroup.Status = &unavailableStatus
		}
		var diffReporter ackcompare.Reporter
		var ctx context.Context
		res, err := rm.CustomModifyReplicationGroup(ctx, desired, latest, &diffReporter)
		assert.Nil(res)
		assert.NotNil(err)
		var requeueNeededAfter *requeue.RequeueNeededAfter
		assert.True(errors.As(err, &requeueNeededAfter))
	})
}
```
and it fails.

**Expected outcome**
The test should pass.

**Environment**

* Kubernetes version
* Using EKS (yes/no), if so version? 
* AWS service targeted (S3, RDS, etc.)

PR #530, opened 2020-11-25
Defer modify replication group if any of its node groups is not in available state

kumargauravsharma:
Issue #529 

Description of changes:
- Added code to check node groups state before invoking modify replication group
- Updated unit tests to include corresponding test scenarios

```make test``` passed.

### DIFF
--- a/services/elasticache/pkg/resource/replication_group/custom_update_api.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_update_api.go
@@ -35,12 +35,22 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 ) (*resource, error) {
 
 	latestRGStatus := latest.ko.Status.Status
-	if latestRGStatus != nil && *latestRGStatus != "available" {
+
+	allNodeGroupsAvailable := true
+	if latest.ko.Status.NodeGroups != nil {
+		for _, nodeGroup := range latest.ko.Status.NodeGroups {
+			if nodeGroup.Status == nil || *nodeGroup.Status != "available" {
+				allNodeGroupsAvailable = false
+				break
+			}
+		}
+	}
+	if latestRGStatus == nil || *latestRGStatus != "available" || !allNodeGroupsAvailable {
 		return nil, requeue.NeededAfter(
 			errors.New("Replication Group can not be modified, it is not in 'available' state."),
 			requeue.DefaultRequeueAfterDuration)
 	}
-
+	
 	// Order of operations when diffs map to multiple updates APIs:
 	// 1. When automaticFailoverEnabled differs:
 	//		if automaticFailoverEnabled == false; do nothing in this custom logic, let the modify execute first.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by A-Hilaly** [comment]
I'm wondering if we should split pkg/ unit tests and service custom code unit tests. I think it makes more sense to run a specific service unit tests using something like `make test SERVICE=elasticache`

**Comment by jaypipes** [comment]
> I'm wondering if we should split pkg/ unit tests and service custom code unit tests. I think it makes more sense to run a specific service unit tests using something like `make test SERVICE=elasticache`

Maybe eventually, sure.

File: services/elasticache/pkg/resource/replication_group/custom_update_api.go [hunk 1]
```diff
@@ -35,12 +35,22 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
+			if nodeGroup.Status == nil || *nodeGroup.Status != "available" {
+				allNodeGroupsAvailable = false
+				break
+			}
+		}
+	}
+	if latestRGStatus == nil || *latestRGStatus != "available" || !allNodeGroupsAvailable {
 		return nil, requeue.NeededAfter(
 			errors.New("Replication Group can not be modified, it is not in 'available' state."),
 			requeue.DefaultRequeueAfterDuration)
```
jaypipes, line 51: eventually we might want to have this interval be configurable? or at least specific to Elasticache?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -35,12 +35,22 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
 	}
-
+
```
jaypipes, line 53: whitespace here?

services/elasticache/pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -16,6 +16,8 @@ package replication_group
 import (
 	"context"
 	"fmt"
+	"github.com/aws/aws-controllers-k8s/pkg/requeue"
+	"github.com/pkg/errors"
```
jaypipes, line 20: this is only a test file of course, but common practice is to put third-party imports together separated by a newline after standard library imports.

services/elasticache/pkg/resource/replication_group/custom_update_api_test.go
```diff
@@ -223,6 +234,69 @@ func TestCustomModifyReplicationGroup(t *testing.T) {
...
+		for _, nodeGroup := range latest.ko.Status.NodeGroups {
+			nodeGroup.Status = &unavailableStatus
+		}
+		var diffReporter ackcompare.Reporter
+		var ctx context.Context
+		res, err := rm.CustomModifyReplicationGroup(ctx, desired, latest, &diffReporter)
+		assert.Nil(res)
+		assert.Nil(err)
+	})
+}
```
jaypipes, line 298: ++ nice unit tests.

jaypipes (approved):
Good work @kumargauravsharma. Couple little nits inline but you can address in a followup patch whenever...

Issue #494, opened 2020-11-10
Terminal Condition stabilization


echen-98:
Service: Elasticache

Terminal condition "toggles" when set after invalid operation/input (e.g. negative replica count specified for replication group, e.g. scale out on a cluster mode-disabled replication group): doesn't constantly stay set with "true" status.

jaypipes:
@echen-98  By "toggles" do you mean that the value of the `Condition` **fluctuates** between a `false` and `true` value?

echen-98:
@jaypipes Yes

PR #531, opened 2020-11-25
Terminal Condition stabilization

kumargauravsharma:
Issue #494 

Description of changes:
The changes fix the issue where Terminal condition fluctuates between a `false` and `true` value

Steps to reproduce:
The issue occurs when a custom resource was successfully reconciled and its terminal condition is false or not set.
Then if this resource is updated using yaml which would result in failure during modify call; then this issue surfaces.

Cause:
`rm.ReadOne()` for already created resource returns successfully; the code which invokes `patchResource()` before the `rm.Update()` call sets the terminal condition to `false` (because of successful retrieval of `latest` object not having any failure); this is followed by `rm.Update()` call which fails with terminal condition and results in setting it to `true`.

Fix:
- The call to `patchResource()` before the `rm.Update()` call is removed as it was not required considering that `patchResource` is invoked in case of update failure and latest is not nil. 
- Updated the code in manager `onError()` to ensure that latest object is returned in-case of error too.
  - Generated service controllers (due to updated manager template) using: 
```
SERVICE=apigatewayv2 && make build-controller && \
SERVICE=dynamodb && make build-controller && \
SERVICE=ecr && make build-controller && \
SERVICE=elasticache && make build-controller && \
SERVICE=s3 && make build-controller && \
SERVICE=sagemaker && make build-controller && \
SERVICE=sfn && make build-controller && \
SERVICE=sns && make build-controller;
```

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -207,13 +207,6 @@ func (r *reconciler) sync(
 			"diff", diffReporter.String(),
 			"arn", latest.Identifiers().ARN(),
 		)
-		// Before we update the backend AWS service resources, let's first update
-		// the latest status of CR which was retrieved by ReadOne call.
-		// Else, latest read status is lost in-case Update call fails with error.
-		err = r.patchResource(ctx, desired, latest)
-		if err != nil {
-			return err
-		}
 		latest, err = rm.Update(ctx, desired, latest, diffReporter)
 		if err != nil {
 			if latest != nil {
--- a/services/apigatewayv2/pkg/resource/api/manager.go
+++ b/services/apigatewayv2/pkg/resource/api/manager.go
@@ -192,7 +192,7 @@ func (rm *resourceManager) onError(
 ) (acktypes.AWSResource, error) {
 	r1, updated := rm.updateConditions(r, err)
 	if !updated {
-		return nil, err
+		return r, err
 	}
 	for _, condition := range r1.Conditions() {
 		if condition.Type == ackv1alpha1.ConditionTypeTerminal &&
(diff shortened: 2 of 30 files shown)


pkg/runtime/reconciler.go
```diff
@@ -207,13 +207,6 @@ func (r *reconciler) sync(
 			"diff", diffReporter.String(),
 			"arn", latest.Identifiers().ARN(),
 		)
-		// Before we update the backend AWS service resources, let's first update
-		// the latest status of CR which was retrieved by ReadOne call.
-		// Else, latest read status is lost in-case Update call fails with error.
-		err = r.patchResource(ctx, desired, latest)
-		if err != nil {
-			return err
-		}
```
jaypipes, line 216: This was specifically added by @vijtrip2 to handle scenarios in APIGateway v2. Without a test showing whether removing this would impact other service controllers, I'm hesitant to merge this, even though I believe you are correct in removing it here.

Vijay, could you please advise?

vijtrip2: taking a look

vijtrip2: This logic was added by @kumargauravsharma and it will not impact apigatewayv2 changes that I made in readOne operation.

I will add the update operation in apigatewayv2-e2e test as part of issue #539 

I am fine with merging this change in, but if @kumargauravsharma can wait until I finish #539, that would be great.
I should be able to finish it by 12/1/2020 End of Day PST.

vijtrip2 (approved):

pkg/runtime/reconciler.go
vijtrip2: PR for adding apigatewayv2 update-api e2e test. https://github.com/aws/aws-controllers-k8s/pull/544

jaypipes (approved):
thanks for looking into this @vijtrip2, appreciated. I'm happy to merge this now.

Issue #532, opened 2020-11-25
Configuration option to enable/disable automatic setting AWS tag on ACK controller managed resources


kumargauravsharma:
**Is your feature request related to a problem?**
This feature request is related to #148 which mentions:
> provide an ability to configure the ACK service controller to always set certain well-known AWS tag key/value pairs on resources that it manages. 

This feature request is to provide the configurable option for it.

**Describe the solution you'd like**
A description of what you want to happen.

**Describe alternatives you've considered**
A description of any alternative solutions or features you've considered.

Issue #148, opened 2020-08-06
Defining AWS tags that ACK service controllers should always add to resources


jaypipes:
We should provide an ability to configure the ACK service controller to always set certain well-known AWS tag key/value pairs on resources that it manages. This allows for auditability as well as the ability to get some statistics around adoption of ACK.

At a minimum, I'm thinking this:

* `services.k8s.aws/managed=true` to indicate an ACK service controller is managing the lifecycle of the resource
* `services.k8s.aws/created=true` to indicate an ACK service controller created the resource (as opposed to an externally-created resource being "adopted" by an ACK service controller
* `services.k8s.aws/cluster=$KUBERNETES_CLUSTER` could be an EKS cluster ARN or the name of the cluster created by kops or another external k8s installer. The cluster here should be **the cluster that was running the ACK service controller**.
* `services.k8s.aws/namespace=$KUBERNETES_NAMESPACE`

jaypipes:
related: #63

sellers:
I can see a use case where a PaaS team or SRE group would want to modify the configuration so that all resources are tagged with local specific tags that that org uses in AWS (e.g. for ownership, cost accounting, etc) Have them be part of the K8s CRD deployment plus any custom tags that a "developer" deploys.  I could see this being a separate story but wanted to mention this here in case the technical implementation decision are swayed by this desire. 

ex:
{built-in services. tags as suggested above}
{local-default service tags that everyone in t his cluster/namespace gets}
{deployment-specific tags that this deployment gets}

PR #543, opened 2020-12-01
Config option to specify resource tags

kumargauravsharma:
Issue #532 

Description of changes:
- Added `resource-tags` option by referring to `ACK_RESOURCE_TAGS` environment variable. 
- Generated controllers using:
```
SERVICE=elasticache && make build-controller
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -30,6 +30,7 @@ const (
 	flagAWSAccountID         = "aws-account-id"
 	flagAWSRegion            = "aws-region"
 	flagLogLevel             = "log-level"
+	flagResourceTags         = "resource-tags"
 )
 
 // Config contains configuration otpions for ACK service controllers
@@ -41,6 +42,7 @@ type Config struct {
 	AccountID                string
 	Region                   string
 	LogLevel                 string
+	ResourceTags             []string
 }
 
 // BindFlags defines CLI/runtime configuration options
@@ -81,6 +83,11 @@ func (cfg *Config) BindFlags() {
 		&cfg.LogLevel, flagLogLevel,
 		"info",
 		"The log level. Default is info. We use logr interface which only supports info and debug level",
+	)
+	flag.StringSliceVar(
+		&cfg.ResourceTags, flagResourceTags,
+		[]string{},
+		"Configures the ACK service controller to always set key/value pairs tags on resources that it manages.",
 	)
 }
 
--- a/pkg/runtime/tags.go
+++ b/pkg/runtime/tags.go
@@ -1,0 +1,69 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package runtime
+
+import (
+	"strings"
+	"time"
+
+	ackconfig "github.com/aws/aws-controllers-k8s/pkg/config"
+	acktypes "github.com/aws/aws-controllers-k8s/pkg/types"
+)
+
+// GetDefaultTags provides Default tags (key value pairs) for given resource
+func GetDefaultTags(
+	config *ackconfig.Config,
+	metadata *acktypes.RuntimeMetaObject,
+) map[string]string {
+	if metadata == nil || config == nil || len(config.ResourceTags) == 0 {
+		return nil
+	}
+	var populatedTags = make(map[string]string)
+	for _, tagKeyVal := range config.ResourceTags {
+		keyVal := strings.Split(tagKeyVal, "=")
+		if keyVal == nil && len(keyVal) != 2 {
+			continue
+		}
+		key := strings.TrimSpace(keyVal[0])
+		val := strings.TrimSpace(keyVal[1])
+		if key == "" || val == "" {
+			continue
+		}
+		populatedValue := expandTagValue(&val, metadata)
+		populatedTags[key] = *populatedValue
+	}
+	if len(populatedTags) == 0 {
+		return nil
+	}
+	return populatedTags
+}
+
+func expandTagValue(
+	value *string,
+	metadata *acktypes.RuntimeMetaObject,
+) *string {
+	if value == nil || metadata == nil {
+		return nil
+	}
+	var expandedValue string = ""
+	switch *value {
+	case "%UTCNOW%":
+		expandedValue = time.Now().UTC().String()
+	case "%KUBERNETES_NAMESPACE%":
+		expandedValue = (*metadata).GetNamespace()
+	default:
+		expandedValue = *value
+	}
+	return &expandedValue
+}
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -19,6 +19,7 @@ AWS_ROLE_ARN=${AWS_ROLE_ARN:-""}
 ACK_ENABLE_DEVELOPMENT_LOGGING="true"
 ENABLE_PROMETHEUS=${ENABLE_PROMETHEUS:-"false"}
 ACK_LOG_LEVEL="debug"
+ACK_RESOURCE_TAGS='services.k8s.aws/managed=true, services.k8s.aws/created=%UTCNOW%, services.k8s.aws/namespace=%KUBERNETES_NAMESPACE%'
 DELETE_CLUSTER_ARGS=""
 K8S_VERSION=${K8S_VERSION:-"1.16"}
 PRESERVE=${PRESERVE:-"false"}
@@ -146,6 +147,7 @@ export AWS_REGION
 export AWS_ROLE_ARN
 export ACK_ENABLE_DEVELOPMENT_LOGGING
 export ENABLE_PROMETHEUS
+export ACK_RESOURCE_TAGS
 export ACK_LOG_LEVEL
 
 service_config_dir="$ROOT_DIR/services/$AWS_SERVICE/config"
@@ -190,6 +192,7 @@ kubectl -n ack-system set env deployment/ack-"$AWS_SERVICE"-controller \
     AWS_ACCOUNT_ID="$AWS_ACCOUNT_ID" \
     ACK_ENABLE_DEVELOPMENT_LOGGING="$ACK_ENABLE_DEVELOPMENT_LOGGING" \
     ACK_LOG_LEVEL="$ACK_LOG_LEVEL" \
+    ACK_RESOURCE_TAGS="$ACK_RESOURCE_TAGS" \
     AWS_REGION="$AWS_REGION" 1>/dev/null
 sleep 10
 echo "ok."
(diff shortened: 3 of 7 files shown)


pkg/config/config.go
```diff
@@ -65,6 +67,11 @@ func (cfg *Config) BindFlags() {
 		"Configures the logger to use a Zap development config (encoder=consoleEncoder,logLevel=Debug,stackTraceLevel=Warn, no sampling), "+
 			"otherwise a Zap production config will be used (encoder=jsonEncoder,logLevel=Info,stackTraceLevel=Error), sampling).",
 	)
+	flag.BoolVar(
+		&cfg.EnableDefaultTagging, flagEnableDefaultTagging,
+		true,
+		"Configures the ACK service controller to always set certain well-known AWS tag key/value pairs on resources that it manages.",
+	)
```
jaypipes, line 74: Instead of a boolean configuration variable, how about we have a string configuration variable called `--resource-tags` that takes a comma-delimited list of key=value pairs?

pkg/config/config.go
```diff
@@ -11,7 +11,7 @@
 // express or implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
-package runtime
+package config
```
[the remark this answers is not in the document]
kumargauravsharma: It was required to remove cyclic dependency causing compile failure.

pkg/config/config.go
kumargauravsharma: I think having `--resource-tags` option *in addition* to the boolean flag would be helpful.
The boolean flag would toggle the default tagging behavior with well known default tags, 
while `--resource-tags` would help extend/append to the list of default tags (instead of replacing it).

pkg/config/config.go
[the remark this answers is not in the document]
jaypipes: > It was required to remove cyclic dependency causing compile failure.

You could have just exported the `Reconciler.cfg` field as `Reconciler.Cfg` since the resource manager already has the `rr` field pointing at the `Reconciler` object.

pkg/config/config.go
jaypipes: The problem with that approach is that what happens when the user sets both `--enable-default-tagging` and `--resource-tags`? That user experience is clunky IMHO.

A single `--resource-tags` CLI option is easier to reason about, and more importantly, harder to get wrong since there's no question as to what it would do: specify the tags that resources would get.

pkg/config/config.go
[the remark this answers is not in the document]
kumargauravsharma: I did not export `cfg` as that would have allowed any code with access to it to be able to modify values in it. Thus, I preferred copying it by value while passing as argument to resource manager creation instead of passing its reference.

pkg/config/config.go
kumargauravsharma: The tags mentioned in #148 have dynamic values per custom resource instance/deployment:
- `services.k8s.aws/managed=true`
- `services.k8s.aws/created=true`
- `services.k8s.aws/cluster=$KUBERNETES_CLUSTER`
- `services.k8s.aws/namespace=$KUBERNETES_NAMESPACE` 
These do not seem to fit into static key, value input as `--resource-tags` CLI option.

kumargauravsharma: I find that with `--resource-tags`  option its more likely that default tags might get replaced unintentionally by end user.

jaypipes: I respectfully disagree :)

For the dynamic value interpolation, I do see your point. However, we have a choice here between having an entirely undocumented configuration behaviour (`--enable-default-tagging` option, which does not describe to users what tags would be added to resources) and a configuration behaviour that would rely on some special well-known value substitutions (`--resource-tags=services.k8s.aws/cluster=$KUBERNETES_CLUSTER`). The latter behaviour is at least decently self-documenting and describes to the end user what they should expect.

kumargauravsharma: Updated code to use `--resource-tags` input.

pkg/util/tags.go
```diff
@@ -0,0 +1,66 @@
...
+		populatedValue := expandTagValue(&val)
+		populatedTags[key] = *populatedValue
+	}
+	if len(populatedTags) == 0 {
+		return nil
+	}
+	return populatedTags
+}
+
+func expandTagValue(value *string) *string {
```
jaypipes, line 48: This will likely need to accept the CR as a parameter (in order to expand things like KUBERNETES_NAMESPACE).

pkg/util/tags.go
```diff
@@ -0,0 +1,66 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
jaypipes, line 1: Can we place this into `pkg/runtime/tags.go` please?

jaypipes (changes requested):
This is getting there, @kumargauravsharma. See inline for a few requested changes, though. Thank you!

templates/config/controller/deployment.yaml.tpl
```diff
@@ -50,4 +52,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: ACK_RESOURCE_TAGS
+          value: ""
```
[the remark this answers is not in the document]
kumargauravsharma: This was to ensure that there exist K8s environment variable to refer.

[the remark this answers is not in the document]
jaypipes: But this environment variable does not need to exist.

scripts/kind-build-test.sh
```diff
@@ -189,6 +191,7 @@ kubectl -n ack-system set env deployment/ack-"$AWS_SERVICE"-controller \
     AWS_ACCOUNT_ID="$AWS_ACCOUNT_ID" \
     ACK_ENABLE_DEVELOPMENT_LOGGING="$ACK_ENABLE_DEVELOPMENT_LOGGING" \
     ACK_LOG_LEVEL="$ACK_LOG_LEVEL" \
+    ACK_RESOURCE_TAGS="$ACK_RESOURCE_TAGS"
```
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

templates/config/controller/deployment.yaml.tpl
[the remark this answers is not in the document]
kumargauravsharma: Updated code.

pkg/util/tags.go
```diff
@@ -0,0 +1,66 @@
...
+
+func expandTagValue(value *string) *string {
+	if value == nil {
+		return nil
+	}
+	var expandedValue string = ""
+	switch *value {
+	case "$UTCNOW":
+		expandedValue = time.Now().UTC().String()
+	case "$KUBERNETES_CLUSTER":
```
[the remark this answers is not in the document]
kumargauravsharma: Removed `$KUBERNETES_CLUSTER`

pkg/util/tags.go
kumargauravsharma: Updated code.

pkg/util/tags.go
kumargauravsharma: Moved to `pkg/runtime/tags.go`

pkg/runtime/tags.go
```diff
@@ -0,0 +1,69 @@
...
+) *string {
+	if value == nil || metadata == nil {
+		return nil
+	}
+	var expandedValue string = ""
+	switch *value {
+	case "$UTCNOW":
+		expandedValue = time.Now().UTC().String()
+	case "$KUBERNETES_NAMESPACE":
+		expandedValue = (*metadata).GetNamespace()
```
jaypipes, line 64: :+1:

templates/helm/values.yaml.tpl
```diff
@@ -28,6 +28,13 @@ aws:
   # If specified, use the AWS region for AWS API calls
   region: ""
 
+resourceTags:
+  # Configures the ACK service controller to always set key/value pairs tags on resources that it manages.
+  - services.k8s.aws/managed=true
+  - services.k8s.aws/created=$UTCNOW
+  - services.k8s.aws/cluster=$KUBERNETES_CLUSTER
```
jaypipes, line 35: remove the above?

jaypipes (changes requested):
one teeny thing to fix up, otherwise this is ready to go.

pkg/runtime/tags.go
```diff
@@ -0,0 +1,69 @@
...
+func expandTagValue(
+	value *string,
+	metadata *acktypes.RuntimeMetaObject,
+) *string {
+	if value == nil || metadata == nil {
+		return nil
+	}
+	var expandedValue string = ""
+	switch *value {
+	case "$UTCNOW":
```
A-Hilaly, line 61: Is there a reason behind using `$` as a prefix ? I'm afraid it might be confusing or misleading for folks trying to run services controller locally or using bash scripts

pkg/runtime/tags.go
```diff
@@ -0,0 +1,69 @@
...
+// permissions and limitations under the License.
+
+package runtime
+
+import (
+	acktypes "github.com/aws/aws-controllers-k8s/pkg/types"
+	"strings"
+	"time"
+
+	ackconfig "github.com/aws/aws-controllers-k8s/pkg/config"
```
A-Hilaly, line 21: nit: Import should preferably be organized in groups. https://github.com/golang/go/wiki/CodeReviewComments\#imports

templates/helm/values.yaml.tpl
kumargauravsharma: Removed entry for `KUBERNETES_CLUSTER`

pkg/runtime/tags.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+// GetDefaultTags provides Default tags (key value pairs) for given resource
+func GetDefaultTags(
+	config *ackconfig.Config,
+	metadata *acktypes.RuntimeMetaObject,
+) map[string]string {
+	if metadata == nil || config == nil || config.ResourceTags == nil || len(config.ResourceTags) == 0 {
```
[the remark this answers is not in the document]
kumargauravsharma: Updated the code.

pkg/runtime/tags.go
kumargauravsharma: Removed `$` prefix.

pkg/runtime/tags.go
kumargauravsharma: Organized imports.

A-Hilaly (approved):
LGTM 👍

pkg/runtime/tags.go
```diff
@@ -0,0 +1,69 @@
...
+func expandTagValue(
+	value *string,
+	metadata *acktypes.RuntimeMetaObject,
+) *string {
+	if value == nil || metadata == nil {
+		return nil
+	}
+	var expandedValue string = ""
+	switch *value {
+	case "UTCNOW":
```
jaypipes, line 61: The `$` sign in the expansion strings earlier were to indicate that the strings would be interpolated. If we don't want to use `$`, then we should pick some other interpolation indicator. Perhaps `{UTCNOW}` or `{{ UTCNOW }}`.

Without any interpolation indicator, it's difficult to tell that the string will be replaced.

kumargauravsharma: I think using `{}` or `{{ }}` could create confusion with template files and `{{ }}` syntax used there.

jaypipes: Yes, that's a good point. how about `%UTCNOW%`?

jaypipes (changes requested):
I still believe we need some variable interpolation indicator. How about `%XXXX%`?

pkg/runtime/tags.go
kumargauravsharma: Updated code to use `%UTCNOW%` syntax.

A-Hilaly: I like the idea of using `%`

A-Hilaly (approved):

jaypipes (approved):
:+1:  thanks @kumargauravsharma :)

PR #547, opened 2020-12-02
removed getopts strategy from kind-build-test.sh

niconosenzo:
Efforts part of Issue #504
Refactoring stage: Argument variable values as env variables instead of getopts strategy
Script: kind-build-test.sh

Description of changes:
- Replaced getopts strategy with a fixed number of arguments (aws service name ) and env variable.
- Note that AWS_ROLE_ARN is defined as a variable during script startup:
`AWS_ROLE_ARN=${AWS_ROLE_ARN:-""}`
The above will cause the script to not fail even when the environment variable (previously passed as script argument) is not explicitly defined. I assumed this is intentional.

### DIFF
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -12,18 +12,16 @@ TEST_DIR="$ROOT_DIR/test"
 TEST_E2E_DIR="$TEST_DIR/e2e"
 TEST_RELEASE_DIR="$TEST_DIR/release"
 
-OPTIND=1
-CLUSTER_NAME_BASE="test"
+CLUSTER_NAME_BASE=${CLUSTER_NAME_BASE:-"test"}
 AWS_ACCOUNT_ID=${AWS_ACCOUNT_ID:-""}
 AWS_REGION=${AWS_REGION:-"us-west-2"}
 AWS_ROLE_ARN=${AWS_ROLE_ARN:-""}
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
Nice work @niconosenzo! :)

RedbackThomson:
This appears to have broken the `make kind-test` target - which is failing our automated testing.

RedbackThomson:
Further, I found an `echo` that still references the old command line arguments - https://github.com/aws/aws-controllers-k8s/blob/main/scripts/kind-build-test.sh#L59

There may be more, that I missed.

niconosenzo:
> Further, I found an `echo` that still references the old command line arguments - https://github.com/aws/aws-controllers-k8s/blob/main/scripts/kind-build-test.sh#L59
> 
> There may be more, that I missed.

I will change that line on a separate PR, it doesn't look like there are more out there.

niconosenzo:
> This appears to have broken the `make kind-test` target - which is failing our automated testing.

The PR should have not modified anything but only the way it's called. You may probably create an issue, I will take a look

RedbackThomson:
> > This appears to have broken the `make kind-test` target - which is failing our automated testing.
> 
> The PR should have not modified anything but only the way it's called. You may probably create an issue, I will take a look

It did not modify the Makefile, but the Makefile references the old getopts version of this script.

PR #551, opened 2020-12-03
Bash error handling

tbutler4:
Issue #, if available:
#504 
Description of changes:
1. Added safeguards to make error handling more strict. Included the option -x in hopes to make debugging easier. I can use set -Eo pipefail Instead if that works better.

2. There are three files with trap signals. [generate-crds.sh, kind-build-test.sh, sagemaker.sh]. I left the traps to fill their purpose and added set options to sagemaker which was the only one without them. But I can delete traps and corresponding signals if wanted.

3. Lib and aws folders barely had any set options so any files without set options I left that way. I can also add set options to all files in scripts if wanted


scripts/build-controller-image.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -E
+set -euxo pipefail
```
jaypipes, line 3: Hi @tbutler4! Sorry for the delayed review on this!

The addition of the `-x` option will make the default output of the scripts extremely verbose, so I think that should be hidden behind a conditional on the `DEBUG` environs variable inside `scripts/lib/common.sh`:

```bash
if [[ "${DEBUG}" = "true" ]]; then
    set -x
fi
```

For the `-u` option, unfortunately there are a number of scripts that assume the presence of variables, but they do so in a "safe" manner by encapsulating Bash tests for the value of a potentially unset variable with `"${SOME_VAR}" = "someval"` like I've done in the code above. The `-u` option will just print a whole bunch of error messages that really aren't errors, so I'm afraid you'll need to remove that from the option set.

jaypipes (changes requested):

scripts/build-controller-image.sh
tbutler4: Hi @jaypipes,

With all the exciting news from reInvent, I appreciate how you made time to offer support. So thank you!!

Your suggestions have helped me learn more on how to maneuver through bash scripts. I believe everything is correct now but please let me know if any additionally changes are necessary.

scripts/lib/common.sh
```diff
@@ -81,5 +83,8 @@ debug_msg() {
     if [ -n "$__indent_level" ]; then
         __indent="$( for each in $( seq 0 $__indent_level ); do printf " "; done )"
     fi
+    if [[ "${DEBUG}" = "true" ]]; then
+        set -x
+    fi
```
jaypipes, line 88: Hi @tbutler4! Actually, I thought you should add this block *outside* the `debug_msg()` Bash function at the top of the file. Since this file is `source`'d from almost every other script right at the beginning of the script, having this block at the top of the `scripts/lib/common.sh` file outside of a function scope will essentially set the -x option for the remainder of the calling/sourcing script's execution.

tbutler4: ahhh I see how that is more beneficial. I moved the block of code to the correct spot. This portion of #504 should be completed now I believe.

jaypipes (approved):
:+1:

PR #558, opened 2020-12-07
Rename Status and Spec Field

nmvk:
Added ability to rename the additional Status and Spec field.

For ElastiCache, updated status from ScaleUpModifications to AllowedScaleUpModifications.

```
 Allowed Scale Up Modifications:
    cache.m4.10xlarge
    cache.m4.2xlarge
    cache.m4.4xlarge
    cache.m4.large
    cache.m4.xlarge
    cache.m5.12xlarge
    cache.m5.24xlarge
```

### DIFF
--- a/pkg/generate/config/config.go
+++ b/pkg/generate/config/config.go
@@ -151,10 +151,13 @@ type AdditionalFieldConfig struct {
 	// OperationID refers to the ID of the API Operation where we will
 	// determine the field's Go type.
 	OperationID string `json:"operation_id,omitempty"`
-	// MemberName refers to the name of the member of the
+	// SourceName refers to the name of the member of the
 	// Input shape in the Operation identified by OperationID that
 	// we will take as our additional spec/status field.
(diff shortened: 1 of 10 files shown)


pkg/generate/config/config.go
```diff
@@ -155,6 +155,9 @@ type AdditionalFieldConfig struct {
 	// Input shape in the Operation identified by OperationID that
 	// we will take as our additional spec/status field.
 	MemberName string `json:"member_name"`
+	// RenameMemberName refers to the name that will be used instead of
+	// MemberName in the status.
+	RenameMemberName string `json:"rename_member_name,omitempty"`
```
jaypipes, line 160: Let's call these fields `SourceName` and `TargetName`. I think that most explicitly describes what the values represent. Thoughts?

nmvk: Done updated

jaypipes (approved):
Perfect. :) Thank you!

PR #546, opened 2020-12-02
overhaul generation framework with TemplateSet

jaypipes:
Adds a new `pkg/generate/templateset` containing a `TemplateSet` struct
that makes our template generation extensible. No more hard-coded
template paths or unused templates for Crossplane. Instead, with this
patch, we now have a `pkg/generate/ack` package that contains the
instructions for generating ACK components and a
`pkg/generate/crossplane` that contains instructions for generating
Crossplane components.

Issue #538

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #559, opened 2020-12-08
Crossplane pipeline to adopt new template sets

muvaf:
Description of changes: The generation framework has undergone an overhaul in https://github.com/aws/aws-controllers-k8s/pull/546 and there were a few things broken in Crossplane pipeline. This PR fixes those, removes more template files in `template/crossplane`


jaypipes:
@muvaf could you elaborate on what were the specific things that were broken in the Crossplane pipeline?

cmd/ack-generate/command/crossplane.go
```diff
@@ -67,24 +68,29 @@ func generateCrossplane(_ *cobra.Command, args []string) error {
...
 	}
-
-	ts, err := cpgenerate.Crossplane(g, optTemplatesDir)
+	ts := templateset.New(optTemplatesDir, crossplane.IncludePaths, crossplane.CopyPaths, crossplane.TemplateFuncs)
+	generation := crossplane.NewGeneration(g, ts,
+		crossplane.WithInitializer(crossplane.AddAPIFiles),
+		crossplane.WithInitializer(crossplane.AddCRDFiles),
+		crossplane.WithInitializer(crossplane.AddControllerFiles),
+	)
+	output, err := generation.Run()
```
jaypipes, line 88: I'm a little confused why you went back to the Generation.Run() thing here... what wasn't possible to do with the `pkg/generate/crossplane.Crossplane` function?

pkg/generate/crossplane/crossplane.go
```diff
@@ -14,157 +14,47 @@
...
-				"zz_"+target,
-			)
-			crdVars := &templateCRDVars{
-				metaVars,
-				crd,
-			}
-			if err = ts.Add(outPath, "pkg/"+target, crdVars); err != nil {
-				return nil, err
-			}
-		}
```
jaypipes, line 167: The previous code nicely laid out the generation of templates/files including code comments about what was happening at the different stages. That is now hidden behind the Initializer things. I'm wondering what benefit the Initializers have over the previous code? I understand that the `hooks.go` file should not have had a `zz_` prefix on it, but couldn't that have just been generated once instead of in this CRD loop?

cmd/ack-generate/command/crossplane.go
muvaf: The fixes `pkg/generate/crossplane.Crossplane` needed weren't that big but I'm planning to write the tests separately for each one of those initializers instead of testing the whole pipeline. Also wanted to separate the core logic in `Run` so that when we a developer adds a new initializer, they don't have to understand the whole flow.

pkg/generate/crossplane/crossplane.go
muvaf: Similar to my [comment](https://github.com/aws/aws-controllers-k8s/pull/559#discussion_r541111665) above, I separated the stages in the code so that we can test them separately.

> I understand that the hooks.go file should not have had a zz_ prefix on it, but couldn't that have just been generated once instead of in this CRD loop?

Yes, that's a viable approach. Since `hooks.go` is kind of an exception in the whole pipeline, I wanted to deal with it at a higher level so that the actual generation function can be re-used in different places without being worried about what gets eliminated. Essentially, it's to keep the initializer functions do only one thing (initialize the templates) and removal of `hooks.go` didn't feel like belonging to _templating_ part of the logic.

muvaf:
> @muvaf could you elaborate on what were the specific things that were broken in the Crossplane pipeline?

It was mainly about file names and paths. But since I wanted to write the tests (haven't got to write them yet) per generation logic, I separated those into different objects rather than the one big flow `crossplane.Crossplane`. The actual templating functionality wasn't affected.

cmd/ack-generate/command/crossplane.go
jaypipes: I'm still not really liking the whole "Initializer" thing here... I was trying to encapsulate all the necessary logic regarding building the templates into the `pkg/generate/crossplane`  package instead of having to have the CLI tool contain this logic.

jaypipes: @muvaf OK, I pushed up an alternative fix for this. Please let me know whether you'd be OK with it.

https://github.com/aws/aws-controllers-k8s/pull/593

muvaf: Sounds good to me. I've just tested your [PR](https://github.com/aws/aws-controllers-k8s/pull/593) and I think we can close this one.

muvaf:
Closing in favor of https://github.com/aws/aws-controllers-k8s/pull/593

Issue #431, opened 2020-10-23
404 exception handling needs to include error message 


surajkota:
**Is your feature request related to a problem?**
Some SageMaker services do not have a specific exception for `ResourceNotFound` case. The describe* API call returns a ValidationException with a specific message like (Training) `Requested resource not found`, (Processing) `Could not find requested job`, (Endpoint) `Could not find endpoint`, (Model) `Could not find model` etc.

**Describe the solution you'd like**
Currently generator.yaml ([sample](https://github.com/aws/aws-controllers-k8s/blob/e86d38994ff8f73a5604432e8c634839451e109f/services/ecr/generator.yaml#L5)) only supports error codes for exception codes. Generator yaml needs additional logic to take a string as input for error message and check for its existence in requestFailure.Message()

#385

PR #564, opened 2020-12-09
add exception message check instructions for code gen

surajkota:
#### Description of changes:
Issue #431 

Generator Config offers the option to override the exception name which corresponds to a specific HTTP Status code/error type for a resource. While this is sufficient for most APIs, some need a more comprehensive check because the exception code is too generic. For e.g., describe* API in SageMaker, returns a ValidationException with a specific message like (Training) _Requested resource not found_, (Model) _Could not find model_ etc. 
```
An error occurred (ValidationException) when calling the DescribeModel operation: Could not find model "arn:aws:sagemaker:us-west-2:169544399729:model/xyz".
```
Same API would also throw the same exception if the Name in the API call is [malformed](https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_DescribeModel.html#API_DescribeModel_RequestSyntax)
```
An error occurred (ValidationException) when calling the DescribeModel operation: 1 validation error detected: Value 'XY$$$Z' at 'modelName' failed to satisfy constraint: Member must satisfy regular expression pattern: ^[a-zA-Z0-9](-*[a-zA-Z0-9])*
```

While the former should be treated as job not found and the later is a terminal condition for the controller. This PR adds the ability to code generator to do an additional check to differentiate between a exception treated as 404 vs others in such conditions

The spec for exception codes in **generator.yaml** has changed from:
```
codes:
        404: ResourceNotFoundException
```
to
```
errors:
        404:
          code: ValidationException                     #Required field
          messagePrefix: Could not find model  #Optional field
```

I have made the changes needed for existing services

#### Testing
```
make build-ack-generate

make build-controller SERVICE=apigatewayv2 && \
make build-controller SERVICE=dynamodb && \
make build-controller SERVICE=ecr && \
make build-controller SERVICE=elasticache && \
make build-controller SERVICE=s3 && \
make build-controller SERVICE=sagemaker && \
make build-controller SERVICE=sfn && \
make build-controller SERVICE=sns;

make test

PRESERVE=true AWS_ROLE_ARN=<> TMP_DIR=<> ./scripts/kind-build-test.sh s3
PRESERVE=true AWS_ROLE_ARN=<> TMP_DIR=<> ./scripts/kind-build-test.sh ecr
PRESERVE=true AWS_ROLE_ARN=<> TMP_DIR=<> ./scripts/kind-build-test.sh sagemaker
```


pkg/generate/ack/controller.go
```diff
@@ -51,6 +51,9 @@ var (
 		"ResourceExceptionCode": func(r *ackmodel.CRD, httpStatusCode int) string {
 			return r.ExceptionCode(httpStatusCode)
 		},
+		"ResourceExceptionMessagePrefixCheck": func(r *ackmodel.CRD, httpStatusCode int) string {
```
jaypipes, line 54: Let's call the function `GoCodeResourceExceptionMessagePrefixCheck` since it outputs Go code (and the other functions that do so begin with `GoCode` :) )

pkg/model/crd.go
```diff
@@ -449,6 +447,29 @@ func (r *CRD) ExceptionCode(httpStatusCode int) string {
...
+		if found && resGenConfig.Exceptions != nil {
+			if excConfig, present := resGenConfig.Exceptions.Errors[httpStatusCode]; present &&
+				resGenConfig.Exceptions.Errors[httpStatusCode].MessagePrefix != nil {
+				return fmt.Sprintf("&& strings.HasPrefix(awsErr.Message(), \"%s\") ",
+					*excConfig.MessagePrefix)
+			}
+		}
+	}
+	return ""
+}
```
jaypipes, line 471: :+1:

jaypipes (changes requested):
@surajkota just a quick request to rename the template function. otherwise, looks great! (though you might also update the PR summary to be more descriptive please) :)

surajkota:
Thanks @jaypipes, addressed all the comments

pkg/generate/config/config.go
```diff
@@ -274,13 +274,30 @@ type FieldConfig struct {
...
+	Code string `json:"code"`
+	// MessagePrefix is an optional string field to be checked as prefix of the
+	// exception message in addition to exception name. This is needed for HTTP codes
+	// where the exception name alone is not sufficient to determine the type of error.
+	// Example: SageMaker service throws ValidationException if job does not exist
+	// as well as if IAM role does not have sufficient permission to fetch the dataset
+	// For the former controller should proceed with creation of job whereas the
+	// later is a terminal state.
+	// In Go SDK terms - awsErr.Message()
+	MessagePrefix *string `json:"messagePrefix,omitempty"`
```
jaypipes, line 298: sorry, I forgot to catch this previously... the `json` tag here should be `message_prefix` not `messagePrefix` to match the style of the other configuration elements in the YAML file (see `terminal_codes` for example)

jaypipes (changes requested):
One little thing I missed on a previous revision.. see inline.

pkg/generate/config/config.go
surajkota: good catch, fixed!

jaypipes (approved):
Great work, @surajkota, thank you sir! :+1: :)

PR #569, opened 2020-12-10
rm getopts strategy from provision-kind-cluster.sh

niconosenzo:
Efforts part of Issue #504
Refactoring stage: Argument variable values as env variables instead of getopts strategy
Script: provision-kind-cluster.sh

Description of changes:
- Replaced getopts strategy with a fixed number of arguments and env variables.
- Removed unused variable OVERRIDE_PATH.
- Fixed reference to this script from kind-build-test.sh.

### DIFF
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -104,7 +104,8 @@ if [ -z "$TMP_DIR" ]; then
     CLUSTER_NAME_BASE=$(uuidgen | cut -d'-' -f1 | tr '[:upper:]' '[:lower:]')
     CLUSTER_NAME="ack-test-$CLUSTER_NAME_BASE"-"${TEST_ID}"
     TMP_DIR=$ROOT_DIR/build/tmp-$CLUSTER_NAME
-    $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}" -v "${K8S_VERSION}"
+    export K8S_VERSION
+    $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}"
 fi
 export PATH=$TMP_DIR:$PATH
 
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -104,8 +104,7 @@ if [ -z "$TMP_DIR" ]; then
     CLUSTER_NAME_BASE=$(uuidgen | cut -d'-' -f1 | tr '[:upper:]' '[:lower:]')
     CLUSTER_NAME="ack-test-$CLUSTER_NAME_BASE"-"${TEST_ID}"
     TMP_DIR=$ROOT_DIR/build/tmp-$CLUSTER_NAME
-    export K8S_VERSION
-    $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}"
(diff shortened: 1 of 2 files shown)


scripts/provision-kind-cluster.sh
```diff
@@ -24,18 +23,17 @@ K8_1_16="kindest/node:v1.16.9@sha256:7175872357bc85847ec4b1aba46ed1d12fa054c83ac
...
-
 USAGE="
 Usage:
-  $(basename "$0") CLUSTER_NAME [-v K8S_VERSION]
+  $(basename "$0") CLUSTER_NAME
 
 Provisions a KinD cluster for local development and testing.
 
-Example: $(basename "$0") my-test -v 1.16
+Example: export K8S_VERSION=1.18; $(basename "$0") my-test
```
jaypipes, line 32: I'd remove the `export K8S_version=1.18` here, since it's an optional argument and described in the `Environment variables:` section in the USAGE already.

scripts/provision-kind-cluster.sh
```diff
@@ -45,26 +43,21 @@ if [[ -z "$cluster_name" ]]; then
...
+    
+    # Check if version is supported:
+    if [ -z $K8_VERSION ]; then
+        echo "Version set: $K8_VER"
+        echo "K8s version not supported" 1>&2
+        exit 2
+    fi
+else
+    K8_VERSION=${K8_1_16}
+fi
```
jaypipes, line 60: The above is fine (since it works, of course). However, in a future PR, I'd love to see the above converted into code that is more readable and less Bash-magic.

Perhaps add a function into `scripts/lib/k8s.sh` that takes a version parameter and returns a full SHA1 string from a Bash array lookup (instead of the `sed` and `${!K8_VER}` stuff which I find pretty inscrutable :)

scripts/kind-build-test.sh
```diff
@@ -104,7 +104,8 @@ if [ -z "$TMP_DIR" ]; then
     CLUSTER_NAME_BASE=$(uuidgen | cut -d'-' -f1 | tr '[:upper:]' '[:lower:]')
     CLUSTER_NAME="ack-test-$CLUSTER_NAME_BASE"-"${TEST_ID}"
     TMP_DIR=$ROOT_DIR/build/tmp-$CLUSTER_NAME
-    $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}" -v "${K8S_VERSION}"
+    export K8S_VERSION
+    $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}"
```
jaypipes, line 108: since the `K8S_VERSION` variable isn't used anywhere else, you don't need to export the variable here. Instead, just set the variable in the call signature:

```bash
K8S_VERSION="$K8S_VERSION" $SCRIPTS_DIR/provision-kind-cluster.sh "${CLUSTER_NAME}"
```

jaypipes (changes requested):
couple teeny nits inline, @niconosenzo :)

scripts/provision-kind-cluster.sh
niconosenzo: done :)

scripts/provision-kind-cluster.sh
niconosenzo: totally agree @jaypipes , will work on that at next.

jaypipes (approved):
:+1:

PR #573, opened 2020-12-10
fix e2e workflow to only pass one arg

jaypipes:
I goofed in #572. There should only be a single parameter passed to the
kind-build-test.sh script.

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -37,6 +37,6 @@ jobs:
         run: |
           export AWS_ROLE_ARN=$(aws ssm get-parameter --name ACK_ROLE_ARN | jq -r '.Parameter.Value')
           export AWS_ROLE_ARN_ALT=$(aws ssm get-parameter --name ACK_ROLE_ARN_ALT | jq -r '.Parameter.Value')
-          ./scripts/kind-build-test.sh $SERVICE $AWS_ROLE_ARN
+          ./scripts/kind-build-test.sh $SERVICE
         env:
           SERVICE: ${{ matrix.service }}


.github/workflows/e2e-test.yaml
```diff
@@ -37,6 +37,6 @@ jobs:
         run: |
           export AWS_ROLE_ARN=$(aws ssm get-parameter --name ACK_ROLE_ARN | jq -r '.Parameter.Value')
           export AWS_ROLE_ARN_ALT=$(aws ssm get-parameter --name ACK_ROLE_ARN_ALT | jq -r '.Parameter.Value')
-          ./scripts/kind-build-test.sh $SERVICE $AWS_ROLE_ARN
+          ./scripts/kind-build-test.sh $SERVICE
```
surajkota, line 40: why dont we use the make targets directly? that will keep the docs and workflows in sync

jaypipes: Because the `make kind-test` target uses a persistent Kind cluster :)

surajkota: so we can set PRESERVE to false here then

surajkota: reason I am saying is last week a PR broke the make command

jaypipes: we used to use the make targets directly in the workflow scripts. I changed that a few months ago because the make targets assume a local build environment and were doing things like running all unit tests (`make test`) as a target dependency of other targets like `make kind-test`. This makes sense on a local development environment where you make changes and want to run `make kind-test` as a singular command to functionally test everything you are working on. It does not make sense, however, when you have separate test runners (running on entirely different underlying compute engines -- some containers, some VMs -- to have the make targets represent dependencies in this way. We have separate test runners that run the build and unit tests. There's no reason to re-run those build and unit tests on the test runners that are running e2e tests. It just wastes time and cycles.

PR #574, opened 2020-12-10
Update terminal conditions for SDK Find

nmvk:
InvalidParameterCombination exception can be thrown for Elasticache snapshot find. This results in endless exceptions and no status change in `kubectl get or describe`

After the change terminal condition is set.

```
"conditions": [
            {
                "message": "Cannot specify both ReplicationGroupId and CacheClusterId",
                "status": "True",
                "type": "ACK.Terminal"
            }
        ]
```


jaypipes (approved):
:+1:

PR #576, opened 2020-12-11
option to setup prometheus during kind-test

kumargauravsharma:
Issue #, if available:

Description of changes:
* Adds container ports in deployment files for metrics discovery
* Option to setup prometheus during kind-test (default false):

Use with `make kind-test`:
```
ENABLE_PROMETHEUS=true make kind-test
```
Upon setup, the prometheus URL (http://localhost:9090) can be accessed on host machine.

### DIFF
--- a/kind-setup/prometheus/prometheus-setup.yaml
+++ b/kind-setup/prometheus/prometheus-setup.yaml
@@ -1,0 +1,97 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus-config
+  namespace: ack-system
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval: 15s
+    scrape_configs:
+      - job_name: 'service-endpoints'
+        kubernetes_sd_configs:
+          - role: endpoints
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: prometheus-cluster-role
+rules:
+  - apiGroups: [""]
+    resources:
+      - nodes
+      - services
+      - endpoints
+      - pods
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources:
+      - configmaps
+    verbs: ["get"]
+  - nonResourceURLs: ["/metrics"]
+    verbs: ["get"]
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: prometheus-service-account
+  namespace: ack-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: prometheus-role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: prometheus-cluster-role
+subjects:
+  - kind: ServiceAccount
+    name: prometheus-service-account
+    namespace: ack-system
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: prometheus-deployment
+  namespace: ack-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: prometheus
+  template:
+    metadata:
+      labels:
+        app: prometheus
+    spec:
+      containers:
+        - name: prometheus-cont
+          image: prom/prometheus
+          ports:
+            - containerPort: 9090
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/prometheus/prometheus.yml
+              subPath: prometheus.yml
+      volumes:
+        - name: config-volume
+          configMap:
+            name: prometheus-config
+      serviceAccountName: prometheus-service-account
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: prometheus-service
+  namespace: ack-system
+spec:
+  selector:
+    app: prometheus
+  ports:
+    - name: prometheusmetricsport
+      nodePort: 30900
+      protocol: TCP
+      port: 9090
+      targetPort: 9090
+  type: NodePort
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -17,6 +17,7 @@ AWS_ACCOUNT_ID=${AWS_ACCOUNT_ID:-""}
 AWS_REGION=${AWS_REGION:-"us-west-2"}
 AWS_ROLE_ARN=${AWS_ROLE_ARN:-""}
 ACK_ENABLE_DEVELOPMENT_LOGGING="true"
+ENABLE_PROMETHEUS=${ENABLE_PROMETHEUS:-"false"}
 ACK_LOG_LEVEL="debug"
 DELETE_CLUSTER_ARGS=""
 K8S_VERSION=${K8S_VERSION:-"1.16"}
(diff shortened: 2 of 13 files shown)


kind-setup/prometheus/prometheus-setup.yaml
```diff
@@ -0,0 +1,75 @@
...
+    metadata:
+      labels:
+        app: prometheus
+    spec:
+      containers:
+        - name: prometheus-cont
+          image: custom-prometheus:embedded_config
+          ports:
+            - containerPort: 9090
+      serviceAccountName: prometheus-service-account
```
jaypipes, line 59: Add a volume mount here instead of needing to build a special Docker image for Prometheus...

```yaml
        volumeMounts:
          - mountPath: /etc/prometheus/prometheus.yml
            name: etc-prometheus
        volumes:
          - name: etc-prometheus
            hostPath:
              path: $(PROMETHEUS_SETUP_DIR)/prometheus.yml
```

kind-setup/prometheus/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM prom/prometheus
+ADD prometheus.yml /etc/prometheus/
```
jaypipes, line 2: There's no need to build a separate Docker image for Prometheus. We can simply specify a volume mount in the `Deployment.Spec.Template.Spec` (see below)

scripts/kind-build-test.sh
```diff
@@ -190,5 +193,10 @@ echo "==========================================================================
 
 export KUBECONFIG
 
+if [[ "$ENABLE_PROMETHEUS" == true ]]; then
+    echo -n "Setting up prometheus into the cluster ... "
+    setup_prometheus "$CLUSTER_NAME"
+fi
```
jaypipes, line 199: I think you can remove `scripts/lib/prometheus.sh` entirely and just add this into the `kind-build-test.sh` script above around line 130 where we are loading the other images into the KinD cluster:

```bash
if [[ "$ENABLE_PROMETHEUS" == true ]]; then
    echo -n "Loading prometheus image into the cluster ... "
    kind load docker-image --quiet --name "${CLUSTER_NAME}" \
        --nodes="${CLUSTER_NAME}"-worker,"${CLUSTER_NAME}"-control-plane \
        prom/prometheus || exit 1
    echo "ok."
fi
```

and then here, just do:

```bash
if [[ "$ENABLE_PROMETHEUS" == true ]]; then
    PROMETHEUS_SETUP_DIR=$ROOT_DIR/kind-setup/prometheus
    PROMETHEUS_SETUP_FILE_PATH=$PROMETHEUS_SETUP_DIR/prometheus-setup.yaml
    echo -n "Deploying prometheus into the cluster ... "
    kubectl apply -f "$PROMETHEUS_SETUP_FILE_PATH" 1>/dev/null
    echo "ok."
fi
```
```

File: templates/config/controller/deployment.yaml.tpl [hunk 4]
```diff
@@ -49,3 +51,18 @@ spec:
...
+  namespace: ack-system
+spec:
+  selector:
+    control-plane: controller
+  ports:
+    - name: metricsport
+      port: 8080
+      targetPort: 8080
+      protocol: TCP
+  type: NodePort
```
jaypipes, line 68: Would be great to put this into a separate `service.yaml.tpl` file in a future PR...

jaypipes (changes requested):
Thanks @kumargauravsharma! Please see inline for requested changes to this. Basically, there's no need to build a special Docker image for Prometheus. Instead, we should just add a volume mount to the deployment that the kind test sets up for Prometheus server.

kind-setup/prometheus/prometheus-setup.yaml
A-Hilaly: ++ We can also use a `ConfigMap` and mount it as a volume

kumargauravsharma: Updated code to use config map.

scripts/kind-build-test.sh
kumargauravsharma: Updated the code per the comments.

kind-setup/prometheus/Dockerfile
kumargauravsharma: Removed the Dockerfile.

templates/config/controller/deployment.yaml.tpl
kumargauravsharma: created `service.yaml.tpl` file.

kind-setup/prometheus/prometheus-setup.yaml
```diff
@@ -0,0 +1,97 @@
...
+    spec:
+      containers:
+        - name: prometheus-cont
+          image: prom/prometheus
+          ports:
+            - containerPort: 9090
+          volumeMounts:
+            - name: config-volume
+              mountPath: /etc/prometheus/prometheus.yml
+              subPath: prometheus.yml
```
A-Hilaly, line 76: 👍

scripts/kind-build-test.sh
```diff
@@ -127,6 +128,13 @@ echo "$AWS_SERVICE_DOCKER_IMG" > "${TMP_DIR}"/"${AWS_SERVICE}"_docker-img
 echo -n "loading the images into the cluster ... "
 kind load docker-image --quiet --name "${CLUSTER_NAME}" --nodes="${CLUSTER_NAME}"-worker,"${CLUSTER_NAME}"-control-plane "${AWS_SERVICE_DOCKER_IMG}" || exit 1
 echo "ok."
+if [[ "$ENABLE_PROMETHEUS" == true ]]; then
+    echo -n "Loading prometheus image into the cluster ... "
+    kind load docker-image --quiet --name "${CLUSTER_NAME}" \
+        --nodes="${CLUSTER_NAME}"-worker,"${CLUSTER_NAME}"-control-plane \
+        prom/prometheus || exit 1
+    echo "ok."
+fi
```
A-Hilaly, line 137: I'm not sure if this is needed. `prom/prometheus` is a public container image, the nodes will be able to download the image if needed.

A-Hilaly: Thinking more about this, i realised that we might need a `kubectl wait` to wait for the prometheus pod to be running

kumargauravsharma: Added wait logic.

kumargauravsharma: Not loading the image before hand, lead to `ErrImagePull` error with message: `Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit.`

A-Hilaly: Ahh :/ .. Docker hub pull limits ¯\\_(ツ)_/¯

A-Hilaly (approved):
LGTM 👍

kind-setup/prometheus/prometheus-setup.yaml
```diff
@@ -0,0 +1,97 @@
...
+  name: prometheus-config
+  namespace: ack-system
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval: 15s
+    scrape_configs:
+      - job_name: 'service-endpoints'
+        kubernetes_sd_configs:
+          - role: endpoints
```
jaypipes, line 13: :+1:

jaypipes (approved):
Good patience on this, @kumargauravsharma, thank you!

PR #579, opened 2020-12-15
Update Snapshot sdkFind

nmvk:
Use only snapshot name to find the resource. Snapshot name is unique so we do not need
other parameters. 


kumargauravsharma (approved):

jaypipes (changes requested):
Chatted with @nmvk on Slack... I think if SnapshotName is nil, then this PR will produce a DescribeSnapshotInput shape that has no fields filled in, which would result in a failure...

jaypipes (approved):

PR #580, opened 2020-12-15
Fixes for Snapshot failures

kumargauravsharma:
Fixes for Snapshot failures:
```
1. Fix snapshot crud failures:
   Cause: update did not save the sync status condition
   Fix: update() needed to return latest retrieved.
2. Fix endless reconciler loop:
   Cause: ReadOne->onError() sets terminal condition to false using patchResource,
     followed by error returned from Create() leads to
     setting terminal condition to true using patchResource;
     causing reconciler loop.
   Fix: ensure patchResource is invoked once in the reconciler::sync()
3 Snapshots e2e tests fix - updated scenarios where create snapshot expects either Replication Group or Cache Cluster Id and fails if other is supplied.
```

Testing done:
```
1. Snapshot e2e tests passed.
2. apigatewayv2 e2e tests passed.
```

Issue #, if available:

Description of changes:

### DIFF
--- a/pkg/runtime/reconciler.go
+++ b/pkg/runtime/reconciler.go
@@ -156,15 +156,15 @@ func (r *reconciler) sync(
 
 	latest, err := rm.ReadOne(ctx, desired)
 	if err != nil {
-		if latest != nil {
-			// this indicates, that even though ReadOne failed
-			// there is some changes available in the latest.RuntimeObject()
-			// (example: ko.Status.Conditions) which have been
-			// updated in the resource
-			// Thus, patchResource() call should be made here
-			_ = r.patchResource(ctx, desired, latest)
-		}
 		if err != ackerr.NotFound {
+			if latest != nil {
+				// this indicates, that even though ReadOne failed
+				// there is some changes available in the latest.RuntimeObject()
+				// (example: ko.Status.Conditions) which have been
+				// updated in the resource
+				// Thus, patchResource() call should be made here
+				_ = r.patchResource(ctx, desired, latest)
+			}
 			return err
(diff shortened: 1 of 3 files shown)


nmvk (approved):
LGTM

A-Hilaly (approved):
/lgtm

echen-98 (approved):

pkg/runtime/reconciler.go
```diff
@@ -156,15 +156,15 @@ func (r *reconciler) sync(
...
 	latest, err := rm.ReadOne(ctx, desired)
 	if err != nil {
-		if latest != nil {
-			// this indicates, that even though ReadOne failed
-			// there is some changes available in the latest.RuntimeObject()
-			// (example: ko.Status.Conditions) which have been
-			// updated in the resource
-			// Thus, patchResource() call should be made here
-			_ = r.patchResource(ctx, desired, latest)
-		}
```
jaypipes, line 166: I don't understand why this change was necessary. Why would ReadOne be returning err == NotFound && a non-nil `latest` variable?

kumargauravsharma: The earlier version of the code was having the err == NotFound && a non-nil latest scenario.
This PR fixes that.
With the changes in this PR; the resource would be patched when `err != NotFound && a not nil latest`. This scenario occurs when invalid input is given to SDK find.

jaypipes: gotcha, thanks for the explanation. This is partly why I was a little hesitant about #574

jaypipes (approved):

PR #587, opened 2020-12-17
E2E test framework written in Python

RedbackThomson:
Issue #, if available: N/A

Description of changes:
The aim of this PR is to migrate our integration tests away from pure Bash scripts. Instead, I have written the existing SageMaker integration tests using PyTest. I have also provided a Dockerfile and a runner script (that builds and runs said Dockerfile) that encapsulates all of the dependencies and passes through the necessary variables to run on KIND (or EKS).

The flow of the integration test framework is as such:
- Run a bootstrap Python script (currently invoked using `./bootstrap <service>`)
  - Invokes the individual service's `service_bootstrap.py` file, creating AWS resources
  - Stores the exported properties of each resource into a `bootstrap.yaml` file in the service directory
- Run the PyTest module in the service directory.
  - Helper methods load templated resources (CR YAMLs) and injects values
  - Runs the PyTest fixtures and tests as expected
- Run a cleanup Python script (`./cleanup <service>`)
  - Invokes the individual service`s `service_cleanup`py` file, deleting the bootstrapped resources

Features:
- Tests can be marked as `canary`, if you wish to only run a subset of integration tests for canaries
- Helper methods provide abstractions over the K8s API for common tasks (those previously found in `lib/aws.sh`)

TODO:
- Add support for job parallelism (using `pytest-xdist`)
- Set up Github Actions workflow
- Support passing test markers through to Docker


test/e2e/run-tests.sh
```diff
@@ -3,11 +3,6 @@
 set -eo pipefail
 
 THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-ROOT_DIR="$THIS_DIR/../.."
-SCRIPTS_DIR="$ROOT_DIR/scripts"
-BIN_DIR="$ROOT_DIR/bin"
-
-source "$SCRIPTS_DIR/lib/common.sh"
```
RedbackThomson, line 10: FYI: I removed this to ensure that the `test/e2e` directory could be run without any context of any parent directories. These scripts were not being used in this file.

test/e2e/run-tests.sh
```diff
@@ -35,13 +30,29 @@ if [ ! -d "$service_test_dir" ]; then
     exit 0
 fi
 
+# check to see if the tests use pytest
+[[ -f "$service_test_dir/__init__.py" ]] && enable_python_tests="true" || enable_python_tests="false"
```
jaypipes, line 34: :+1:

test/e2e/run-tests.sh
```diff
@@ -35,13 +30,29 @@ if [ ! -d "$service_test_dir" ]; then
...
+  done
+else
+  python bootstrap.py "${SERVICE}"
+
+  set +e
+
+  pushd "$service_test_dir" > /dev/null
+    pytest --log-cli-level INFO .
+  popd > /dev/null
+  python cleanup.py "${SERVICE}"
```
jaypipes, line 55: Is this guaranteed to be executed if `pytest .` fails?

jaypipes (changes requested):
I'm really diggin' this, @RedbackThomson :) Thank you!

A few requests/suggestions inline, but I think we can get this merged today and start followup PRs that add documentation about how this new Pytest system works and begin converting the Bash-based stuff.

test/e2e/run-tests.sh
RedbackThomson: I `set +e` just above this in an attempt to ensure they do. I'm not certain that's foolproof, although I couldn't imagine `pytest` could force an exit.

test/e2e/common/k8s.py
```diff
@@ -0,0 +1,199 @@
...
+        logging.debug(f"Waiting for resource {reference} to be synced")
+
+        reference = get_resource(reference)
+        if 'status' not in reference or 'conditions' not in reference['status']:
+            logging.error(
+                f"Expected .status.conditions property to exist in {reference}")
+            return False
+
+        conditions: Dict = reference['status']['conditions']
+        if 'ACK' not in conditions or 'ResourceSynced' not in conditions['ACK']:
```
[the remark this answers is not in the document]
RedbackThomson: Ohhh actually this line of code is a result of a faulty refactoring. The parameter used to be called `resource` rather than `reference`, but then I was reusing the variable name. When I renamed `resource` to `reference` it modified the local variable as well (on line 143):
https://github.com/aws/aws-controllers-k8s/pull/587/files/cc33198c1102ad5dc100a4d5706fc5d2ce69c8e7#diff-c17786fb0613026f7c25ee4300969ab7ed42e3e6b219d5f871345cf7aefa9af5R143

So actually, this process of splitting up the status fields by path name is from the K8s API (or the Python interpretation) itself, nothing to do with `dataclass` (pitchforks can be lowered again).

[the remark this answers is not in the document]
RedbackThomson: Nevertheless I will provide a helper method to pull out both this and `ACK.Terminal`, just to break up the code flow. (And I've renamed the variables to make sense again)

test/e2e/run-tests.sh
jaypipes: Maybe we refactor the framework a bit (in a future PR) to use pytest.fixture and add finalizers to a module-scoped fixture?

https://docs.pytest.org/en/latest/fixture.html

jaypipes (approved):
thoroughly awesome. :)

thanks for putting together a readme, too!

PR #592, opened 2020-12-18
Fix replication group tests failures

kumargauravsharma:
Issue #, if available:

Description of changes:

- Fix credential expire issue during elasticache tests failures. This caused the controller to fail to get the latest details from service API and the test would timeout waiting for latest updates.
- Fixes for replication group tests failures by
    - Include security group Ids and engine versions properties
      in modify replication group API call
      only if there is change in their values
    - Updated sync condition logic to: 
       a/ check all cache clusters' available status from node groups and 
       b/ check if the cacheclusters count from node groups match the count from member clusters
    - Added logs messages to help debugging
    - Updated replication group tests that were failing due to invalid input yaml
- Added cache subnet group terminal codes in generator config

### DIFF
--- a/scripts/lib/testutil.sh
+++ b/scripts/lib/testutil.sh
@@ -75,24 +75,34 @@ assert_pod_not_restarted() {
 # k8s_wait_resource_synced checks the given resource for an ACK.ResourceSynced condition in its
 #   k8s status.conditions property. Times out if condition has not been met for a long time. This function
 #   is intended to be used after yaml application to await creation of a resource.
-# k8s_wait_resource_synced requires 2 arguments:
+# k8s_wait_resource_synced requires 3 arguments:
 #   k8s_resource_name: the name of the resource, e.g. "snapshots/test-snapshot"
 #   wait_periods: the number of 60-second periods to wait for the resource before timing out
+#   service_name: the aws service name. It is used to determine the controller deployment, in case credentials need
+#                 to be rotated during the wait.
 k8s_wait_resource_synced() {
-  if [[ $# -ne 2 ]]; then
+  if [[ $# -ne 3 ]]; then
     echo "FATAL: Wrong number of arguments passed to ${FUNCNAME[0]}"
-    echo "Usage: ${FUNCNAME[0]} k8s_resource_name wait_periods"
+    echo "Usage: ${FUNCNAME[0]} k8s_resource_name wait_periods service_name"
     exit 1
   fi
 
   local k8s_resource_name="$1"
   local wait_periods="$2"
+  local service_name="$3"
 
   kubectl get "$k8s_resource_name" 1>/dev/null 2>&1
   assert_equal "0" "$?" "Resource $k8s_resource_name doesn't exist in k8s cluster" || exit 1
 
   local wait_failed="true"
   for i in $(seq 1 "$wait_periods"); do
+    # Test role credentials expire after 15 minutes (Refer: aws.sh::aws_generate_temp_creds)
+    # Ensure that credentials are reloaded at the beginning and after every 15 minutes.
+    # else the controller fails to get the latest details from aws service api
+    # and the test fails on sync status.
+    if [[ "$((i % 15))" == "0" || "$i" == "1" ]]; then
+      k8s_controller_reload_credentials "$service_name"
+    fi
     debug_msg "waiting for resource $k8s_resource_name to be synced ($i)"
     sleep 60
 
--- a/scripts/lib/testutil.sh
+++ b/scripts/lib/testutil.sh
@@ -97,10 +97,10 @@ k8s_wait_resource_synced() {
   local wait_failed="true"
   for i in $(seq 1 "$wait_periods"); do
     # Test role credentials expire after 15 minutes (Refer: aws.sh::aws_generate_temp_creds)
-    # Ensure that credentials are reloaded at the beginning and after every 15 minutes.
+    # Ensure that credentials are reloaded after first iteration and after every 15 minutes.
     # else the controller fails to get the latest details from aws service api
     # and the test fails on sync status.
-    if [[ "$((i % 15))" == "0" || "$i" == "1" ]]; then
+    if [[ "$((i % 15))" == "0" || "$i" == "2" ]]; then
       k8s_controller_reload_credentials "$service_name"
     fi
     debug_msg "waiting for resource $k8s_resource_name to be synced ($i)"
--- a/services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
+++ b/services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
@@ -92,6 +92,7 @@ func (rm *resourceManager) provideEvents(
 	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
 	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents-CacheParameterGroup", err)
 	if err != nil {
+		rm.log.V(1).Info("Error during DescribeEvents-CacheParameterGroup", "error", err)
 		return nil, err
 	}
 	events := []*svcapitypes.Event{}
@@ -133,6 +134,7 @@ func (rm *resourceManager) describeCacheParameters(
 			if awsErr, ok := ackerr.AWSError(respErr); ok && awsErr.Code() == "CacheParameterGroupNotFound" {
 				return nil, ackerr.NotFound
 			}
+			rm.log.V(1).Info("Error during DescribeCacheParameters", "error", respErr)
 			return nil, respErr
 		}
 
@@ -198,6 +200,7 @@ func (rm *resourceManager) resetAllParameters(
 	_, err := rm.sdkapi.ResetCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ResetCacheParameterGroup-ResetAllParameters", err)
 	if err != nil {
+		rm.log.V(1).Info("Error during ResetCacheParameterGroup-ResetAllParameters", "error", err)
 		return false, err
 	}
 	return true, nil
@@ -228,6 +231,7 @@ func (rm *resourceManager) resetParameters(
 	_, err := rm.sdkapi.ResetCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ResetCacheParameterGroup", err)
 	if err != nil {
+		rm.log.V(1).Info("Error during ResetCacheParameterGroup", "error", err)
 		return false, err
 	}
 	return true, nil
@@ -288,6 +292,7 @@ func (rm *resourceManager) modifyCacheParameterGroup(
 	_, err := rm.sdkapi.ModifyCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ModifyCacheParameterGroup", err)
 	if err != nil {
+		rm.log.V(1).Info("Error during ModifyCacheParameterGroup", "error", err)
 		return false, err
 	}
 	return true, nil
--- a/services/elasticache/pkg/resource/cache_subnet_group/custom_set_output.go
+++ b/services/elasticache/pkg/resource/cache_subnet_group/custom_set_output.go
@@ -70,6 +70,7 @@ func (rm *resourceManager) provideEvents(
 	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
 	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents-CacheSubnetGroup", err)
 	if err != nil {
+		rm.log.V(1).Info("Error during DescribeEvents-CacheSubnetGroup", "error", err)
 		return nil, err
 	}
 	events := []*svcapitypes.Event{}
--- a/services/elasticache/pkg/resource/replication_group/custom_update_api.go
+++ b/services/elasticache/pkg/resource/replication_group/custom_update_api.go
@@ -19,6 +19,7 @@ import (
 	ackv1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
 	"github.com/aws/aws-controllers-k8s/pkg/requeue"
 	"github.com/pkg/errors"
+	"sort"
 
 	ackcompare "github.com/aws/aws-controllers-k8s/pkg/compare"
 	svcapitypes "github.com/aws/aws-controllers-k8s/services/elasticache/apis/v1alpha1"
@@ -37,6 +38,7 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
 	latestRGStatus := latest.ko.Status.Status
 
 	allNodeGroupsAvailable := true
+	nodeGroupMembersCount := 0
 	if latest.ko.Status.NodeGroups != nil {
 		for _, nodeGroup := range latest.ko.Status.NodeGroups {
 			if nodeGroup.Status == nil || *nodeGroup.Status != "available" {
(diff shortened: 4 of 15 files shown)


services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+		var value string
+		value = *id
+		desiredIds = append(desiredIds, &value)
+	}
+	sort.Slice(desiredIds, func(i, j int) bool {
+		return *desiredIds[i] < *desiredIds[j]
+	})
+
+	latestIds := []*string{}
+	if latestCacheCluster.SecurityGroups != nil {
```
jaypipes, line 637: This needs a nil-guard. `latestCacheCluster` can be `nil`.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+	rm.metrics.RecordAPICall("UPDATE", "ModifyReplicationGroup-SecurityGroupsIds", respErr)
+	if respErr != nil {
+		rm.log.Error(respErr, "Error during ModifyReplicationGroup-SecurityGroupsIds")
+		return nil, respErr
+	}
+
+	return rm.provideUpdatedResource(desired, resp.ReplicationGroup)
+}
+
+// securityGroupIdsDiffer return true if
```
jaypipes, line 697: s/securityGroupIdsDiffer/engineVersionDiffer/

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+	latest *resource,
+	latestCacheCluster *svcsdk.CacheCluster,
+) bool {
+	if desired.ko.Spec.EngineVersion == nil {
+		return false
+	}
+	desiredEV := *desired.ko.Spec.EngineVersion
+
+	var latestEV string = ""
+	if latestCacheCluster.EngineVersion != nil {
```
jaypipes, line 710: needs a nil-guard since `latestCacheCluster` can be `nil`

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -87,7 +107,41 @@ func (rm *resourceManager) CustomModifyReplicationGroup(
...
+	}
+
+	// SecurityGroupIds
+	if rm.securityGroupIdsDiffer(desired, latest, latestCacheCluster) {
+		return rm.updateSecurityGroupIds(ctx, desired, latest)
+	}
+	// EngineVersion
+	if rm.engineVersionDiffer(desired, latest, latestCacheCluster) {
+		return rm.updateEngineVersion(ctx, desired, latest)
+	}
```
jaypipes, line 142: Is there a reason you are making multiple API calls if both sec groups and engine version are different? Why not roll all updates into a single ModifyReplicationGroup API call?

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+			if nodeGroupMember.CacheClusterID == nil {
+				continue
+			}
+			cacheClusterId = nodeGroupMember.CacheClusterID
+			break
+		}
+		if cacheClusterId != nil {
+			break
+		}
+	}
```
jaypipes, line 586: Consider breaking the above out into a `getCacheClusterIDFromNodeGroups` helper function.

Also, the above assumes that all NodeGroups have the same `CacheClusterID` value. Is this always the case?

(if it is the case, one wonders why this isn't a field on the ReplicationGroup and not the NodeGroup...) :(

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+
+	if cacheClusterId == nil {
+		return nil, nil
+	}
+
+	input.SetCacheClusterId(*cacheClusterId)
+	resp, respErr := rm.sdkapi.DescribeCacheClustersWithContext(ctx, input)
+	rm.metrics.RecordAPICall("READ_MANY", "DescribeCacheClusters", respErr)
+	if respErr != nil {
+		rm.log.Error(respErr, "Error during DescribeCacheClusters")
```
jaypipes, line 596: Same comment about removing log.Error

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+			value = *id
+			ids = append(ids, &value)
+		}
+		input.SetSecurityGroupIds(ids)
+	}
+
+	resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
+	rm.metrics.RecordAPICall("UPDATE", "ModifyReplicationGroup-SecurityGroupsIds", respErr)
+	if respErr != nil {
+		rm.log.Error(respErr, "Error during ModifyReplicationGroup-SecurityGroupsIds")
```
jaypipes, line 690: Same comment about removing log.Error

services/elasticache/pkg/resource/replication_group/custom_update_api.go
```diff
@@ -490,6 +555,193 @@ func (rm *resourceManager) newUpdateShardConfigurationRequestPayload(
...
+	}
+
+	if desired.ko.Spec.EngineVersion != nil {
+		input.SetEngineVersion(*desired.ko.Spec.EngineVersion)
+	}
+
+	resp, respErr := rm.sdkapi.ModifyReplicationGroupWithContext(ctx, input)
+	rm.metrics.RecordAPICall("UPDATE", "ModifyReplicationGroup-EngineVersion", respErr)
+	if respErr != nil {
+		rm.log.Error(respErr, "Error during ModifyReplicationGroup-EngineVersion")
```
jaypipes, line 738: Same comment about removing log.Error

templates/pkg/resource/manager.go.tpl
```diff
@@ -177,6 +177,7 @@ func (rm *resourceManager) onError(
 	r *resource,
 	err error,
 ) (acktypes.AWSResource, error) {
+	rm.log.Info("Error occurred.", "Error: ", err.Error())
```
jaypipes, line 180: remove this please.

templates/pkg/resource/manager.go.tpl
```diff
@@ -186,6 +187,7 @@ func (rm *resourceManager) onError(
 			condition.Status == corev1.ConditionTrue {
 			// resource is in Terminal condition
 			// return Terminal error
+			rm.log.Info("Terminal condition is set to true for resource.")
```
jaypipes, line 190: remove this please.

scripts/lib/testutil.sh
```diff
@@ -75,24 +75,34 @@ assert_pod_not_restarted() {
...
   assert_equal "0" "$?" "Resource $k8s_resource_name doesn't exist in k8s cluster" || exit 1
 
   local wait_failed="true"
   for i in $(seq 1 "$wait_periods"); do
+    # Test role credentials expire after 15 minutes (Refer: aws.sh::aws_generate_temp_creds)
+    # Ensure that credentials are reloaded after first iteration and after every 15 minutes.
+    # else the controller fails to get the latest details from aws service api
+    # and the test fails on sync status.
+    if [[ "$((i % 15))" == "0" || "$i" == "2" ]]; then
+      k8s_controller_reload_credentials "$service_name"
```
jaypipes, line 104: this is a hack but I'm OK doing this for now. With the new Python-based testing framework we should be able to make this more elegant.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -92,6 +92,7 @@ func (rm *resourceManager) provideEvents(
 	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
 	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents-CacheParameterGroup", err)
 	if err != nil {
+		rm.log.Error(err, "Error during DescribeEvents-CacheParameterGroup")
```
jaypipes, line 95: I generally dislike using the log.Error call anywhere other than the root calling location. The controller-runtime will automatically log any error it gets back from the resourceManagers, so why not just return nil, respErr here and remove the log.Error calls

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -133,6 +134,7 @@ func (rm *resourceManager) describeCacheParameters(
 			if awsErr, ok := ackerr.AWSError(respErr); ok && awsErr.Code() == "CacheParameterGroupNotFound" {
 				return nil, ackerr.NotFound
 			}
+			rm.log.Error(err, "Error during DescribeCacheParameters")
```
jaypipes, line 137: Same comment about removing log.Error

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -198,6 +200,7 @@ func (rm *resourceManager) resetAllParameters(
 	_, err := rm.sdkapi.ResetCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ResetCacheParameterGroup-ResetAllParameters", err)
 	if err != nil {
+		rm.log.Error(err, "Error during ResetCacheParameterGroup-ResetAllParameters")
```
jaypipes, line 203: Same comment about removing log.Error

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -228,6 +231,7 @@ func (rm *resourceManager) resetParameters(
 	_, err := rm.sdkapi.ResetCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ResetCacheParameterGroup", err)
 	if err != nil {
+		rm.log.Error(err, "Error during ResetCacheParameterGroup")
```
jaypipes, line 234: Same comment about removing log.Error

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
```diff
@@ -288,6 +292,7 @@ func (rm *resourceManager) modifyCacheParameterGroup(
 	_, err := rm.sdkapi.ModifyCacheParameterGroupWithContext(ctx, input)
 	rm.metrics.RecordAPICall("UPDATE", "ModifyCacheParameterGroup", err)
 	if err != nil {
+		rm.log.Error(err, "Error during ModifyCacheParameterGroup")
```
jaypipes, line 295: Same comment about removing log.Error

services/elasticache/pkg/resource/cache_subnet_group/custom_set_output.go
```diff
@@ -70,6 +70,7 @@ func (rm *resourceManager) provideEvents(
 	resp, err := rm.sdkapi.DescribeEventsWithContext(ctx, input)
 	rm.metrics.RecordAPICall("READ_MANY", "DescribeEvents-CacheSubnetGroup", err)
 	if err != nil {
+		rm.log.Error(err, "Error during DescribeEvents-CacheSubnetGroup")
```
jaypipes, line 73: Same comment about removing log.Error

jaypipes (changes requested):
See inline please :)

Also, needs a rebase.

templates/pkg/resource/manager.go.tpl
kumargauravsharma: During debugging, I found this log helpful.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: done

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: done

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: done

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated the logic to make single modify API call.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated method name to be `getAnyCacheClusterIDFromNodeGroups`.
Any cluster Id suffice because it is used to get details about the `SecurityGroups` and `EngineVersion` data which is same for all clusters.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: The error messages logged here were helpful in debugging.
Please revisit.

jaypipes: Please change to `rm.log.V(1)` then. The error is already logged higher in the reconciler.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
jaypipes: Please note that `err` is the wrong variable above. Should be `respErr`

templates/pkg/resource/manager.go.tpl
jaypipes: Please remove this from the template, Kumar. If we want to keep the original error reference, we should be using `errors.WithMessage()` and returning the wrapped error. The top-level reconciler logs errors so we shouldn't duplicate that logging here.

https://godoc.org/github.com/pkg/errors#WithMessage

For example, we could do the following here:

```go
errMsg := fmt.Sprintf("terminal condition: %s", err)
return r1, errors.WithMessage(ackerr.Terminal, errMsg)
```

templates/pkg/resource/manager.go.tpl
jaypipes: @kumargauravsharma please remove the above.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated code.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated code.

services/elasticache/pkg/resource/replication_group/custom_update_api.go
kumargauravsharma: Updated code.

templates/pkg/resource/manager.go.tpl
kumargauravsharma: Removed the changes done to template file.

templates/pkg/resource/manager.go.tpl
kumargauravsharma: Removed the changes done to template file.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Used `rm.log.V(1).Info`

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Updated code

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Updated code.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Updated code.

services/elasticache/pkg/resource/cache_parameter_group/custom_api.go
kumargauravsharma: Updated code.

services/elasticache/pkg/resource/cache_subnet_group/custom_set_output.go
kumargauravsharma: Updated code.

jaypipes (approved):
Thanks for your patience on this @kumargauravsharma. LGTM!

PR #593, opened 2020-12-18
crossplane: only generate hooks.go once

jaypipes:
The crossplane code generation was generating a `zz_hooks.go` file in
each resource subpackage. This was incorrect in two ways: first, the
file should not have been prefixed with `zz_`. Secondly, there only
needs to be a single `pkg/controller/$SERVICE/$RESOURCE/hooks.go` file
generated, not one per resource.

This PR simply moves the generation of the `hooks.go` template out of
the loop over CRDs inside the crossplane TemplateSet constructor and
ensures that a single hooks.go file is generated from the first CRD's
structure.

Also removes unncessary templates that crossplane doesn't use and
corrects a mistake in the hooks.go.tpl for the ReadMany code path's
isUpToDate() function.


pkg/generate/crossplane/crossplane.go
```diff
@@ -169,5 +168,21 @@ func Crossplane(
 			}
 		}
 	}
+
+	// Finally, generate the single hooks.go file from one of the CRDs (doesn't
```
muvaf, line 172: We actually generate a `hooks.go` for every controller, hence for every CRD in that service. The only difference between other controller files is that `hooks.go` is not generated if it already exists because it's meant to be only a boilerplate to be edited by the developer after the code generation step and we don't want to override their custom code.

muvaf (commented):
This looks good to me, thank you @jaypipes !

A few path strings need to be fixed but since it's not in diff, I couldn't suggest changes. Also we should not generate `hooks.go` if it already exists.

Here is a small diff commit that I tested with this one to validate this PR https://github.com/aws/aws-controllers-k8s/commit/b7324b1b37cd633263e2a5d888db7bf2d2ecb921

pkg/generate/crossplane/crossplane.go
```diff
@@ -152,13 +152,12 @@ func Crossplane(
 	targets := []string{
```
muvaf, line 152: Line 132 should have the full path like:
```
outPath := filepath.Join("apis", metaVars.ServiceIDClean, metaVars.APIVersion, "zz_"+strings.TrimSuffix(filepath.Base(path), ".tpl"))
```

jaypipes: sorry. fixed.

PR #599, opened 2020-12-21
combine FieldConfig and AdditionalFieldConfig

jaypipes:
The generator config package had two structs that were serving similar
purposes: `AdditionalFieldConfig` and `FieldConfig`. This patch combines
these two structs into a single `FieldConfig` struct that contains
instructions for handling any field in a resource, regardless of whether
it's a Spec or a Status field (or whether it's "additional" or not).

When informing the code generator about fields for a resource that
should be included, the `FieldConfig.IsReadOnly` attribute indicates
that the field goes in the `Status` struct of a resource (when `true`).

The `AdditionalFieldConfig`'s information is now included in a
`SourceFieldConfig` struct. The `AdditionalFieldConfig.OperationID` and
`AdditionalFieldConfig.SourceName` attributes have been renamed to
`SourceFieldConfig.Operation` and `SourceFieldConfig.Path`; the latter
rename is to support path-based dot-notation subfield access in an input
or output shape. The `FieldConfig.From` attribute is a pointer to one
of these `SourceFieldConfig` structs.

When discovering CRDs, the code generator looks for a
`ResourceConfig.Fields`, which is a map, keyed by the target field name
to use, of `FieldConfig` structs describing interpretation instructions
for how to handle that particular field.

It's easiest to show an example of how the config changed with this
patch. Here is the Elasticache `generator.yaml` section for the
`CacheParameterGroup` resource BEFORE this patch:

```yaml
 resources:
   CacheParameterGroup:
    spec_fields:
      - operation_id: ModifyCacheParameterGroup
        source_name: ParameterNameValues
    status_fields:
      - operation_id: DescribeCacheParameters
        source_name: Parameters
      - operation_id: DescribeEvents
        source_name: Events
```

And here it is AFTER this patch:

```yaml
 resources:
   CacheParameterGroup:
    fields:
      ParameterNameValues:
        from:
          operation: ModifyCacheParameterGroup
          path: ParameterNameValues
      Parameters:
        is_read_only: true
        from:
          operation: DescribeCacheParameters
          path: Parameters
      Events:
        is_read_only: true
        from:
          operation: DescribeEvents
          path: Events
```

I tried to make the new configuration fields more descriptive and
human-readable.

For example, you might read the new configuration like so:

> for the CacheParameterGroup resource, there are three fields on the resource:
> ParameterNameValues, Parameters, and Events. Parameters and Events are
> read only and therefore are in the resource's Status field.
> ParameterNameValues is not read only and so is in the Spec field. The value
> of Spec.ParameterNameValues is set from the
> DescribeCacheParametersInput.Parameters member field. The value of the
> Status.Parameters field is set from the
> DescribeCacheParametersOutput.Parameters member field. The value of the
> Status.Events field is set from the DescribeEventsOutput.Events member
> field


pkg/generate/config/resource.go
```diff
@@ -0,0 +1,342 @@
...
+func (c *Config) OverrideValues(operationName string) (map[string]string, bool) {
+	if c == nil {
+		return nil, false
+	}
+	oConfig, ok := c.Operations[operationName]
+	if !ok {
+		return nil, false
+	}
+	return oConfig.OverrideValues, ok
+}
```
nmvk, line 271: Should we move this to `operation.go`?

nmvk (approved):
One minor comment mostly LGTM

A-Hilaly (approved):
/lgtm

vijtrip2 (approved):
* Watching for the merge of this. My local ImportAPI changes will need to be refactored after this merge. : )
* Meanwhile working with your branch to accommodate the refactoring. If e2e tests pass, I will send a PR today.

services/elasticache/generator.yaml
```diff
@@ -9,9 +9,12 @@ resources:
         - CacheSubnetQuotaExceededFault
         - SubnetInUse
         - InvalidSubnet
-    status_fields:
-      - operation_id: DescribeEvents
-        source_name: Events
+    fields:
+      Events:
+        is_read_only: true
+        setFrom:
```
A-Hilaly, line 15: I think we missed this during the review. I believe it should be just `from`

A-Hilaly: opened #606

PR #611, opened 2020-12-29
standardizing default value expression

niconosenzo:
Efforts part of Issue #504
Refactoring stage: Standardize default value expression.

Description of changes:
Standardize default value expression, moving from :
https://github.com/aws/aws-controllers-k8s/blob/079fa26fe98aa289e5770e8226129917971473f8/scripts/build-controller.sh#L23

To:
https://github.com/aws/aws-controllers-k8s/blob/079fa26fe98aa289e5770e8226129917971473f8/scripts/kind-build-test.sh#L17

This way, becomes easier to read it and Bash only will test for a unset or null parameter before substitute it.


jaypipes (approved):
:+1: thanks @niconosenzo!

PR #617, opened 2021-01-04
E2E Tests for Events, Allowed node modifications

nmvk:
Add new asserts for the existing tests to validate events, allowed node modifications.

Asserting on exact  allowed node type modifications and events is difficult so we assert on length.


jaypipes (approved):
Yep, everything here looks fine, thank you @nmvk! :)

PR #625, opened 2021-01-08
Remove hooks.go and use available Update functionality in Crossplane pipeline

muvaf:
Description of changes:

* Removes `hooks.go` hack by placing the no-op functions in `controller.go` in a way that can be overridden by the developer.
* Fixes the case where there is no `generator-config.yaml` available.
* Uses available Update functionality in Crossplane pipeline as well.

There is no change in parts that are not Crossplane-specific.


templates/crossplane/pkg/controller.go.tpl
```diff
@@ -73,13 +64,18 @@ func (e *external) Observe(ctx context.Context, mg cpresource.Managed) (managed.
 
 {{- if .CRD.Ops.ReadOne }}
 	input := Generate{{ .CRD.Ops.ReadOne.InputRef.Shape.ShapeName }}(cr)
+	if err := e.preObserve(ctx, cr, input); err != nil {
+  	return managed.ExternalObservation{}, errors.Wrap(err, "pre-observe failed")
+  }
```
jaypipes, line 69: nit: indentation is off above andf throughout the added lines here in this template. For `.tpl` files, probably need to `s/\ \ \ \ /\t/gI`

jaypipes (approved):
Awesomesauce. :) Thanks @muvaf!

Issue #41, opened 2020-05-27
Possibility of "adopting" existing AWS resources


marcindulak:
Cloudformation supports importing existing resources https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import.html

A parallel issue is opened in https://github.com/aws/aws-cdk-rfcs/issues/52

mhausenblas:
Thanks @marcindulak and can you flesh out your proposal a little, please?

marcindulak:
It's just a question about ASO that had to be asked at some point.

By "adopting" I mean starting to manage, using ASO, the already existing AWS resources. The term "adopting" is used by pulumi https://www.pulumi.com/blog/adopting-existing-cloud-resources-into-pulumi

An example use case:
- create an s3 bucket by hand
- use the ASO s3 resource to "adopt" the bucket, without changing any of the bucket parameters in AWS. If the state of the ASO s3 resource differs from the state in AWS the application of the ASO resource should fail
- from now on, use the ASO s3 resource to manage the bucket instead of doing this by hand. If the state of the ASO s3 resource differs from the state in AWS the AWS state should be updated

mhausenblas:
Thanks for the clarification @marcindulak. Adopting of existing AWS resources is out of scope, at least for now. I will keep this issue open as a reminder, though for if and when we revisit this topic again.

hikhvar:
Adopting existing resources should be a first class feature. I have two use cases for adopting resources:
1) Migrating existing infrastructure to the aws-controller-k8s. Recreation on migration is simple for resources like Route53 records, security group rules, etc. However, it will be more complicated if you have to manage stateful services like a RDS database or a S3 bucket. Imagine an already deployed application. They may have non-trivial amount of data in those services. Without an "adopt resources" feature, I have to migrate the data to migrate my service to aws-controller-k8s. Depending on the nature of the data, this migration is a non trivial task.

2) The second use case is disaster recovery. Imaging you have to rebuild the k8s cluster but the databases/s3 buckets are still present. Since you lost the state of the cluster, you have to combine the already existing AWS resources with the aws-controller-k8s CRDs.

Vlaaaaaaad:
As discussed on Twitch for _Containers Day_, posting some thoughts here: adopting/importing resources is also vital for handling failures, and upgrades.

[...]

PR #639, opened 2021-01-13
Proposal: Resource Adoption

RedbackThomson:
Issue #, if available: #41 

Description of changes:
- Creates a proposal for a resource adoption system
- *Inviting community participation*


docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+kind: Bucket
+metadata:
+  annotations:
+    services.k8s.aws/arn: arn:aws:s3:::my-existing-bucket
+  ...
+spec:
+   ...
+```
+
+### Notable Pitfalls
```
bwagner5, line 151: In addition to these pitfalls, ACK could adopt a resource that it only has read access to. Reconciliation via describe calls would work, but in the event of a user trying to mutate via ACK, permissions errors would be encountered.  I suppose this could happen to resources even if they are created by ACK if there are only partial write permissions and an update happens later on. Just noting it here, although I don't think it's a huge deal.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Adoption Process Diagram
+
+![sequence-diagram](./images/adoption-sequence-diagram.png)
+# Potential Alternatives
+
+## Resource Adoption Annotation
+
+The existing custom resource definitions (every service resource) would accept an additional annotation that the manager interprets as an indication that the resource specification should be ignored and instead that it should be formed through adoption. For any resource marked with this annotation, upon creation, the manager would attempt to describe the AWS resource associated with the annotated value. If none such resource exists, the manager would update the state of the custom resource into a `AdoptionFailed` terminal state. If the resource does exist, the manager would continue the follow the same adoption process as defined in the [Proposed Implementation](https://quip-amazon.com/lXxVAKU9dv16#SZW9CAMeXE1) section.
```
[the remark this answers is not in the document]
RedbackThomson: This is meant to be an anchor link to the above section - I drafted the document in quip. Thank you.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+* Updates to generator for service-specific configuration
+* Code paths for service managers to receive informer updates
+* Unit and E2E tests with a specific ACK controller (e.g. S3 buckets)
+* Verbose logging for adoption lifecycle updates
+
+# Out of Scope
+
+* Integration of adopted resource types for every ACK controller
+* E2E tests for every ACK controller
+* Tooling for generating `AdoptedResource` CRs
```
marcindulak, line 169: To clarify, in order to adopt a resource, will a user need to add  "an additional annotation" (line 134) on the existing ACK manifest and to create the corresponding `AdoptedResource` CRs?

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+1. Export the identifiers (ARNs/IDs) from the current installation resources’ states
+2. Create `AdoptedResource` CRs for each of these identifiers
+3. Install and configure all necessary controllers to the new cluster
+4. Apply the `AdoptedResource` CRs to the cluster
+    1. Wait for all resources to end up in terminal state
+5. Delete the ACK controller deployments from the old installation
+    1. Do NOT delete the custom resources, this will delete the underlying AWS resources
+6. Apply existing manifests to override the newly created resources
+    1. *Note: If their state is already as desired, this should not make any updates to the AWS resources*
```
marcindulak, line 223: 1. will "existing manifests" be still modified to include "an additional annotation" (line 134)?

2. where is the place and role for "The proposed k8s adopted tag" (line 109)  in this context?

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: The additional annotation is an alternative we explored but found was not suitable. Instead, the user only needs to create an `AdoptedResource` CR.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: 1. No (see https://github.com/aws/aws-controllers-k8s/pull/639#discussion_r557764856)
2. The manager that creates the resource will apply the adopted tag to the new resource automatically. The user doesn't have to take any additional steps.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+* `id` - A resource ID (an integer value)
+
+The custom resource definition makes no assumptions about which fields should be present for any given target kind. It is therefore up to the manager to validate the identifier input shape for existence and correctness.
+
+The target ACK type is specified through the following required fields:
+
+* [API group and version](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning)
+* [Resource kind](https://kubernetes.io/docs/reference/kubectl/overview/#resource-types)
+
+A validating webhook for the `AdoptedResource` custom resources will validate the existence of the target type in the cluster and ensure the requested type is able to be adopted.
```
jaypipes, line 44: Do you envision the webhook to also validate the existence of the AWS resource referred to by the ARN/name/id? Or will the individual ACK service controllers be responsible for that?

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+The target ACK type is specified through the following required fields:
+
+* [API group and version](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning)
+* [Resource kind](https://kubernetes.io/docs/reference/kubectl/overview/#resource-types)
+
+A validating webhook for the `AdoptedResource` custom resources will validate the existence of the target type in the cluster and ensure the requested type is able to be adopted. 
+
+The target metadata field is an arbitrary mapping of key-value pairs that will be applied as the metadata for the created custom resource. If a `name` field is specified in this mapping, the manager should use this value as the custom resource `name` . Otherwise, the manager will use the name specified in the metadata of the `AdoptedResource` as the default name for the resource. This same method will also apply to the `namespace` field. All other fields in the metadata mapping are applied to the created resource without validation.
```
jaypipes, line 46: > If a `name` field is specified in this mapping, the manager should use this value as the custom resource `name` . Otherwise, the manager will use the name specified in the metadata of the `AdoptedResource` as the default name for the resource.

I would advise instead having an `AdoptedResource.Spec.Name` field and `AdoptedResource.Spec.Namespace` field and not used metadata key/value mappings for a 'name' and 'namespace' key.

I realize the reason you need this is because `AdoptedResource` is going to be a CRD and you need to have unique names for AdoptedResource CRs across a Kubernetes namespace, even if you have, say, two AdoptedResources for a resource named "production" and one of those resources is an S3 Bucket and the other is an SNS Topic.

However, I think that `AdoptedResource.Spec.Name` and `AdoptedResource.Spec.Namespace` is the appropriate place to put the Kubernetes name and Kubernetes namespace of the CR to be created, just like the `APIVersion` and `Kind` fields indicate the CR's Kubernetes `APIVersion` and `Kind`. Alternatively, you could do something like this:

```yaml
apiVersion: core.services.k8s.aws/v1alpha1
kind: AdoptedResource
metadata:
  name: adopted-data-bucket
spec:
  kubernetes:
    apiVersion: s3.services.k8s.aws/v1alpha
    kind: Bucket
    name: data-bucket
    namespace: team-a
  aws:
    arn:  arn:aws:s3:::data-bucket
```

And group the Kubernetes-identifying information together and separate from the AWS-identifying information.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Resource Type Blocklisting
+
+Adoption, for some resources, is not possible due to any number of the following reasons:
+
+* They are ephemeral resources which don’t support additional lifecycle operations (one-time jobs)
+* The service cannot provide adequate information to produce a full specification
+* Unforeseen limitations within ACK or the service directly
+
+For any of these cases, the resource can be marked as “not adoptable” and the validation webhook will subsequently reject any adopted resources of this type. Therefore, the manager will never attempt to run the adoption process. This flag can be set in the generator configuration file with the `cannot_adopt` boolean option supplied for any resource.
```
jaypipes, line 125: ++ nice.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+## Adoption Process Diagram
+
+![sequence-diagram](./images/adoption-sequence-diagram.png)
+# Potential Alternatives
+
+## Resource Adoption Annotation
+
+The existing custom resource definitions (every service resource) would accept an additional annotation that the manager interprets as an indication that the resource specification should be ignored and instead that it should be formed through adoption. For any resource marked with this annotation, upon creation, the manager would attempt to describe the AWS resource associated with the annotated value. If none such resource exists, the manager would update the state of the custom resource into a `AdoptionFailed` terminal state. If the resource does exist, the manager would continue the follow the same adoption process as defined in the [Proposed Implementation](#proposed-implementation) section.
```
jaypipes, line 134: This harkens back to my comment above in The Problem section to describe the difficulty right now with this annotation and needing to fill in required fields in the Spec when the annotation essentially signals to the service controller to ignore the values of those Spec fields. :)

jaypipes: I'd almost prefer that we remove this `services.k8s.aws/arn` annotation from ACK entirely and have a single way of signaling resource adoption (using AdoptedResource CRs)

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+### Notable Pitfalls
+
+In the event that the user is attempting to restore resources created in a previous installation of ACK (in the event of an etcd corruption, for example), each manifest will need to be updated with the adoption annotation. Therefore the restoration process would require modifying desired state files (k8s manifests) with actual state fields (the created ARN/ID fields). At the least this introduces a temporary change in k8s manifests - add ARNs to manifests,  apply and adopt, delete ARNs from manifests. More so, it could introduce private field information (generated public API IDs, RDS instance ARNs, etc.) into the manifests repository.
+
+Another pitfall is that, given this solution aims to re-use the existing custom resource definitions, all adopted resources would need to provide a specification body that contains all required fields. Every service contains a distinct set of required values, and as such the user would need to modify every adopted resource type with a distinct set of valid dummy values. Assuming a user provides dummy values for these fields, they would subsequently be overwritten by the manager logic when updating the resource.
```
jaypipes, line 155: bingo.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+# In Scope
+
+* Declaration of the new `AdoptedResource` CRD
+* Updates to generator for service-specific configuration
+* Code paths for service managers to receive informer updates
+* Unit and E2E tests with a specific ACK controller (e.g. S3 buckets)
+* Verbose logging for adoption lifecycle updates
```
jaypipes, line 163: remove this line. discussion of how verbose logging should be is not needed in the scope section :)

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+Another pitfall is that, given this solution aims to re-use the existing custom resource definitions, all adopted resources would need to provide a specification body that contains all required fields. Every service contains a distinct set of required values, and as such the user would need to modify every adopted resource type with a distinct set of valid dummy values. Assuming a user provides dummy values for these fields, they would subsequently be overwritten by the manager logic when updating the resource.
+
+# In Scope
+
+* Declaration of the new `AdoptedResource` CRD
+* Updates to generator for service-specific configuration
+* Code paths for service managers to receive informer updates
```
jaypipes, line 161: hmmm, actually you may not need this. Since all resource managers implement the ReadOne and Create methods, the common ACK runtime can handle all the Informer logic (see `pkg/runtime/cache` for how it handles Informers for Namespaces and ConfigMaps currently). Let's discuss implementation details a little later. I have thoughts here... :)

jaypipes (commented):
Awesome start on this @RedbackThomson! Comments and thoughts inline for you to digest and respond to. :)

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: I think this section is worth removing. I have listed it under "Potential Alternatives" but realistically I don't think we will ever go down this route because of the problems we have both listed. Plus it's not super clear in this format that this is not part of the proposed solution.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
+# Resource Adoption 
+
+# The Problem
+
+New ACK users typically already have implemented their infrastructure in existing solutions, meaning they have deployed production AWS resources into their accounts. While ACK provides good tooling for creating new resources, the path to migration from existing declarative infrastructure technologies into ACK is, so far, bumpy. The current method for migration requires users to store the current state of their applications, delete the resources, and re-create them using ACK. This process introduces numerous places for bugs, loss of the state of existing deployments and downtime in production workloads. Furthermore, considering the case where a K8s cluster loses state, users have no path to recover the previous resources without having to recreate all of them using ACK.
```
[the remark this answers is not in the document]
RedbackThomson: > we cannot use the same CRD as the Kind itself because the fields in the Spec that describe the desired state of the Kind are typically marked as required

I would say this is a potential problem with other of the possible solutions, but it's not a problem that exists in ACK today.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+Users have requested this functionality as being pivotal to their adoption of ACK (see [#41](https://github.com/aws/aws-controllers-k8s/issues/41)). Information from user feedback has been considered as part of this document.
+
+Below are a number of use cases for requesting the ability this feature.
+
+1. Migrating existing architecture from another platform (CFN, CDK, Terraform, etc.)
+    1. Resources exist but have never been managed by K8s - no spec files
+2. Restoring from a broken K8s environment (etcd failure, complete state loss)
+    1. Resources exist and have previously been managed by K8s, but are not being actively reconciled
+3. Migrating architecture across K8s environments
+    1. Resource exist and are actively being managed by another K8s environment
```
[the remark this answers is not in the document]
RedbackThomson: I wanted to call out that it could be any other k8s environment, because that is directly what the SageMaker team hopes to do from our current operator.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+The manager will describe the referenced AWS resource, typically through a `Describe*` SDK call, and use the returned values to compile a new custom resource - with filled spec and status fields. The manager will then apply this custom resource to the cluster using the metadata applied in the `AdoptedResource` spec.
+
+## `AdoptedResource` Specification
+
+The `AdoptedResource` custom resource definition will allow for specification of an AWS resource unique identifier, a target ACK type (with API group and kind), and target metadata fields.
+The AWS resource unique identifier allows identification through at most one of:
+
+* `arn` - An [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) (Amazon Resource Name)
+* `name` - A resource name (a string value)
+* `id` - A resource ID (an integer value)
```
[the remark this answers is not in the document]
RedbackThomson: I wonder if it would then be better to differentiate them, rather than by `name` and `id`, by `stringIdentifier` or `integerIdentifier` maybe? The reason I pointed out validation type was so each manager would already be provided with the cast type, rather than each of them having to parse to integer and give a corresponding error at runtime.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: Absolutely not. I want the validation webhook to be super lightweight, and not care at all about AWS - only about the cluster configuration. Just to provide fast, "apply-time" (idk if that's a concept?) feedback to the user if they're doing something terminally incorrect.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: I still don't know if I like this boolean name. Some other ideas:
- `is_adoptable`
- `block_adoption`

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: Yeah I had that same idea, I suppose with this line I meant indirectly handle `AdoptedResource` updates.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: I don't necessarily see how (in the example above) your `kubernetes` object is any different from my arbitrary metadata key/value mapping.

Say you wanted to attach annotations to this S3 bucket, the `kubernetes` object would now have to support an `annotations` map. Then you wanted to attach a label to the bucket, the object would not have to support a `labels` map. This `kubernetes` object is practically replicating the [`ObjectMeta` type](https://github.com/kubernetes/apimachinery/blob/4f505736214f6a32e70f37c6bae217360fe227b2/pkg/apis/meta/v1/types.go#L110).

Are you suggesting the `kubernetes` object should just be of type `ObjectMeta`? I think that was the direction I was intending when describing this arbitrary map, so that's probably my suggestion now.

docs/design/proposals/adoption/resource-adoption.md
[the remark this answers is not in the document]
jaypipes: > > we cannot use the same CRD as the Kind itself because the fields in the Spec that describe the desired state of the Kind are typically marked as required
> 
> I would say this is a potential problem with other of the possible solutions, but it's not a problem that exists in ACK today.

Sure. I say "exists today" just because the `services.k8s.aws/arn` "solution" exists today -- but use of that annotation is problematic for the reasons described.

docs/design/proposals/adoption/resource-adoption.md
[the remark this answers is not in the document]
jaypipes: Ah, gotcha. ++

docs/design/proposals/adoption/resource-adoption.md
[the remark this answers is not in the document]
jaypipes: What @vijtrip2 said :) The service controller will know how to look up a resource it is responsible for (either via a `name` field, an `id` field or an `arn` field). The AdoptedResource structure should not need to embed this controller-specific type-casting logic.

docs/design/proposals/adoption/resource-adoption.md
jaypipes: > I don't necessarily see how (in the example above) your `kubernetes` object is any different from my arbitrary metadata key/value mapping.

One is a strongly-structured set of fields. The other is a bag of key/value strings with no structure or schema.

> Say you wanted to attach annotations to this S3 bucket, the `kubernetes` object would now have to support an `annotations` map. Then you wanted to attach a label to the bucket, the object would not have to support a `labels` map. This `kubernetes` object is practically replicating the [`ObjectMeta` type](https://github.com/kubernetes/apimachinery/blob/4f505736214f6a32e70f37c6bae217360fe227b2/pkg/apis/meta/v1/types.go#L110).
> 
> Are you suggesting the `kubernetes` object should just be of type `ObjectMeta`? I think that was the direction I was intending when describing this arbitrary map, so that's probably my suggestion now.

Sure, you could definitely use the `corev1.ObjectMeta` type as the Go type of the `spec.kubenetes` field, that's a good idea.

The thing is that we need to differentiate between metadata that describes the adopted resource CR itself and metadata that describes the CR that will be created *from* the adopted resource identifiers.

docs/design/proposals/adoption/resource-adoption.md
jaypipes: `is_adoptable` would be fine. I prefer it, actually, versus `cannot_adopt` because I prefer boolean attributes to be positive versus negative default. `can_adopt` would also work.

docs/design/proposals/adoption/resource-adoption.md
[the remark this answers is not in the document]
Vlaaaaaaad: > you might consider listing a set of Use Cases in a separate section, one of which might mention the use case of an existing Terraform or CloudFormation user where TF/CF has been responsible for managing the lifecycle of an existing S3 Bucket and the user is interested in describing that S3 Bucket resource as a Kubernetes CRD and having ACK manage the resource's lifecycle.

+1 to this. I'd make it clear very early on that resource adoption will just "adopt" the AWS resources, not making any YAML/HCL -> CRD translation or automatic conversion.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+2. The target resource would override an existing resource
+3. The target resource type cannot be adopted
+
+## Adopted K8s Resource Tagging
+
+All adopted resources will be annotated with an ACK-specific annotation that differentiates it from a natively created resource. This annotation serves to ensure clarity for users about possible specification differences introduced during the adoption process, or for filtering by adopted status. However, once adopted into a cluster, a resource will be reconciled with the same logic as any natively created resource. 
+
+The proposed k8s adopted tag is `services.k8s.aws/adopted: true`
+
+The adopted annotation would remain on the custom resource for the remainder of its lifecycle, and should not be modified by any other manager operation.
```
Vlaaaaaaad, line 111: What's the desire driving this limitation?

On the one hand, I get it: when I am in process of migrating from say Terraform to the ACK I want to have the labels and easily identify the resources. Makes total sense.

On the other hand: it's 1 year later, I am no longer using Terraform or care about the adoption process. I still have the `adopted: true` tag which is now redundant. Is there a use for a tag after the adoption process is complete?

---

Thinking more on this, "_not be modified by any other manager operation_" just blocks the ACK Manager from changing it, not the user. So the user could delete the tag? I am a bit confused.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,225 @@
...
+
+All adopted resources will be annotated with an ACK-specific annotation that differentiates it from a natively created resource. This annotation serves to ensure clarity for users about possible specification differences introduced during the adoption process, or for filtering by adopted status. However, once adopted into a cluster, a resource will be reconciled with the same logic as any natively created resource. 
+
+The proposed k8s adopted tag is `services.k8s.aws/adopted: true`
+
+The adopted annotation would remain on the custom resource for the remainder of its lifecycle, and should not be modified by any other manager operation. 
+
+## AWS Resource Tagging
+
+AWS resources adopted by an ACK controller will have tags retroactively applied (whenever possible) to indicate its association with an ACK deployment. The details of these tags are discussed in another issue ([#148](https://github.com/aws/aws-controllers-k8s/issues/148)). ACK managers should not place any conditions on the existence of ACK tags on the AWS resources to gate the adoption process. The manager should should override any existing ACK tags.
```
Vlaaaaaaad, line 115: > The manager should override any existing ACK tags

Hm... could the ACK Manager overwriting the tags maybe lead to fights over resources? Say I have an S3 bucket that is managed from ACK running in cluster `one`. I accidentally create an `AdoptedResource` for that same bucket in ACK running in cluster `two`. In that case, will `two` just take over management of that S3 bucket?

docs/design/proposals/adoption/resource-adoption.md
Vlaaaaaaad: I love the idea of using dummy fields that are then filled when the `Describe` call is done! 

It's such a user-friendly approach: just provide the minimal details needed to identify the resource and we'll do the rest. ❤️

Vlaaaaaaad (commented):
Loving the overall flow of "_submit an `AdoptedResource`, we do some magic, ACK CRD appears_"!

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: > The thing is that we need to differentiate between metadata that describes the adopted resource CR itself and metadata that describes the CR that will be created from the adopted resource identifiers.

So from this discussion it sounds like there should be a field marked `metadata` of type `ObjectMeta` in the `spec` of the `AdoptedResource`. The controller will use this as the source of the created CR's `ObjectMeta`, or if not defined then to default to the `name` and `namespace` of the `AdoptedResource` CR?

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: > It's such a user-friendly approach: just provide the minimal details needed to identify the resource and we'll do the rest. ❤️

The problem is that you would need to provide way more than just the "minimal details needed to identify the resource". You'd need to provide a value for every required field. Even if we did only need to give one identifier field, this doesn't work for some resources where they generate the identifier after creation (it's not a spec field)

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: The idea behind this was as your first case points out, I don't want anything to wipe away the fact that this was an adopted resource in case we need that information for debugging or logical grouping.

However, I think it's totally fine if the user wants to delete it manually. I just wouldn't want some controller deciding when is appropriate.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: This is a point we brought up. Ultimately this is the same if the resource is managed by CDK, TF, Cfn, other operator, etc. Without a solid way of locking a resource to a given cluster, there is no way to strictly ensure it is only reconciled by at most one service.
The ACK controller should have the privilege to override a tag, though, as the use case of migration from one cluster to another would require this.

docs/design/proposals/adoption/resource-adoption.md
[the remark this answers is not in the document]
RedbackThomson: > not making any YAML/HCL -> CRD translation or automatic conversion.

By this do you mean Cfn YAML -> CRD? Because this design does suggest we will be able to generate an ACK CRD for each resource, as part of the adoption.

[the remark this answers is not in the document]
Vlaaaaaaad: Yup, when I said YAML I was thinking of CloudFormation YAML. I should've been more specific 🤦‍♂️

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,202 @@
...
+
+## `AdoptedResource` Specification
+
+The `AdoptedResource` custom resource definition will allow for specification of an AWS resource unique identifier (in the `aws` field) and a target `kubernetes` field - containing Kubernetes `GroupVersionKind` and `metadata` sub-fields.
+
+The `aws` resource unique identifier field allows identification through at most one of:
+
+* `arn` - An [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) (Amazon Resource Name)
+* `name` - A resource name (a string value)
+* `id` - A resource ID (a string value)
```
jaypipes, line 34: I have been chatting with @surajkota about the object relationship feature stuff and it's pretty clear to me that we will want to have some sort of `ackcorev1alpha1.AWSIdentifiers` struct type that contains the above oneOf {name,id,arn}. Please consider making such a struct as part of your implementation so that it may be re-used in things like a concrete object relationship type, for example:

In `apis/core/v1alpha1/identifiers.go`:

```go
type AWSIdentifiers struct {
    // ARN is the AWS Resource Name for the resource. It is a globally
    // unique identifier.
    ARN *AWSResourceName `json:"arn"`
    // Name is a user-supplied string identifier for the resource. It may
    // or may not be globally unique, depending on the type of resource.
    Name *string `json:"name"`
    // ID is an identifier that has been generated by the AWS service for
    // the resource. ID fields are *usually*, but not always, globally unique.
    // Sometimes ID fields are numeric, other times they are strings.
    ID *string `json:"id"`
}
```

and then in, for example, `services/sagemaker/apis/v1alpha1/relations.go`, we might find:

```go
import (
    ackcorev1alpha1 "github.com/aws/aws-controllers-k8s/apis/core/v1alpha1"
)

// ModelReference is a pointer to a Model resource in the Sagemaker API
// and/or the Kubernetes API.
type ModelReference struct {
    // AWS indicates the AWS external identifiers for the Model.
    // If nil, the `Kubernetes` field must not be nil.
    AWS *ackv1alpha1.AWSIdentifiers `json:"aws,omitempty"`
    // Kubernetes indicates the Kubernetes internal identifiers for
    // the Model CR that references the AWS resource. If nil, the
    // Model CR has not yet been created by an ACK service
    // controller and the `AWS` field above should be non-nil.
    Kubernetes *ackcorev1alpha1.PartialObjectMeta `json:"aws,omitempty"
}
```

RedbackThomson:
@jaypipes Do you think it makes sense to have both `ID` and `Name` as AWS identifiers? From an implementation standpoint, this adds unnecessary complexity - we need to pass around extra information for each CRD about whether the primary identifier should come from an `ID` or a `Name` field. Furthermore, since both are string, we would need to cast to integer either way so it's not like one would provide any value over the other in that regard.

RedbackThomson:
Still left outstanding with this design:
- How would we support resources which require multiple identifier fields?
  - For example in `apigatewayv2` a `Route` resource is uniquely identified using an `AppID` and `RouteID`

mwarkentin:
@RedbackThomson some prior art from Terraform, they support multi-part identifiers split with a `/` - here's the command to import a apigatewayv2 route: `terraform import aws_apigatewayv2_route.example aabbccddee/1122334`

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+1. Export the identifiers (ARNs/IDs) from the current installation resources’ states
+2. Create `AdoptedResource` CRs for each of these identifiers
+3. Install and configure all necessary controllers to the new cluster
+4. Apply the `AdoptedResource` CRs to the cluster
+    1. Wait for all resources to end up in terminal state
+5. Delete the ACK controller deployments from the old installation
+    1. Do NOT delete the custom resources, this will delete the underlying AWS resources
+6. Apply existing manifests to override the newly created resources
+    1. *Note: If their state is already as desired, this should not make any updates to the AWS resources*
+7. Delete the custom resources from the old installation
```
RedbackThomson, line 199: This step will necessitate manually removing the finalizers - through force deletion.

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
+# Resource Adoption 
+
+# The Problem
+
+New ACK users typically have existing AWS resources in their accounts. These resources were not originally created by an ACK service controller. While ACK can create new resources, getting ACK service controllers to take over management of an existing resource is not yet supported.
+
+The current method for migrating management of resources over to ACK requires users to store the current state of their applications, delete the resources, and re-create them using ACK. This process introduces places for bugs, loss of the state of existing deployments and downtime in production workloads.
+
+ACK aims to provide the ability to “adopt” existing resources into your Kubernetes environment. Given the user can provide an IAM role which has access to existing resource, and that ACK operators exist for each resource’s associated service, the ACK controller will be able to determine the current specification and status of the AWS resource and reconcile said resource as if it were created natively in k8s.
```
jaypipes, line 9: s/ACK operators/ACK controllers/gI :)

jaypipes: s/as if it were created natively in k8s/as if the ACK controller had originally created it/

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Use Cases
+
+1. Migrating existing architecture from another platform (CFN, CDK, Terraform, etc.)
+    1. Resources exist but have never been managed by K8s - no spec files
```
jaypipes, line 14: s/managed by K8s - no spec files/managed by an ACK controller/

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+3. Migrating architecture across K8s environments
+    1. Resource are being managed by another K8s environment
```
jaypipes, line 16: s/Resource are/Resources are/

s/another K8s environment/an ACK controller installed in another Kubernetes cluster/

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+
+# Proposed Implementation
+
+## Overview
+
+The proposed solution will provide an alternate workflow by which to adopt resources through a separate set of manifests. A new custom resource definition, the `AdoptedResource` CRD, provides a signal to ACK to adopt an existing resource.
```
jaypipes, line 22: s/provide an alternate workflow/provide an explicit workflow/

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+The custom resource definition makes no assumptions about which fields should be present for any given target kind. It is therefore up to the manager to validate the identifier input shape for existence and correctness.
+
+The `kubernetes` field wraps the structures from `GroupVersionKind`, which specify the target ACK type through the following required fields:
+
+* [API group and version](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning)
+* [Resource kind](https://kubernetes.io/docs/reference/kubectl/overview/#resource-types)
+
+A validating webhook for the `AdoptedResource` custom resources will validate the existence of the target type in the cluster and ensure the requested type is able to be adopted. 
+
+The `kubernetes.metadata` field is an optional [`ObjectMeta`](https://github.com/kubernetes/apimachinery/blob/4f505736214f6a32e70f37c6bae217360fe227b2/pkg/apis/meta/v1/types.go#L110) type that will be applied as the metadata for the created custom resource. If the `name` and/o `namespace` fields within the `metadata` spec is not defined, the respective fields from the `AdoptedResource` will be used as the value for the custom resource.
```
jaypipes, line 44: Technically, it's a `PartialObjectMeta`, not an `ObjectMeta`, right?

docs/design/proposals/adoption/resource-adoption.md
```diff
@@ -0,0 +1,201 @@
...
+
+* They are ephemeral resources which don’t support additional lifecycle operations (one-time jobs)
+* The service cannot provide adequate information to produce a full specification
+* Unforeseen limitations within ACK or the service directly
+
+For any of these cases, the resource can be marked as "not adoptable" and the validation webhook will subsequently reject any adopted resources of this type. Therefore, the manager will never attempt to run the adoption process. This option can be managed with a flag in the generator configuration file labelled `is_adoptable`.
+
+## Adoption Process Diagram
+
+![sequence-diagram](./images/adoption-sequence-diagram.png)
```
jaypipes, line 130: remove the `/images` part of the above, since you placed the file into the `docs/design/propsals/adoption/` folder.

jaypipes (changes requested):
@RedbackThomson please see inline for a few minor nits and path changes, otherwise I'm ready to merge this today.

docs/design/proposals/adoption/resource-adoption.md
RedbackThomson: Yes. Implementation detail, or worth putting in the proposal? Functionally identical to the end user

jaypipes: nope, just wanted to make sure I wasn't crazy :)

docs/design/proposals/adoption/resource-adoption.md
jaypipes: I see you moved the png into an `images/` subdirectory, which is fine. Only problem is that the png file is an empty file according to Github :)

RedbackThomson: It's showing as fine to me. 166 KB on the Github UI, and when viewing as raw I see the image.

jaypipes: That is soooooo weird. It says "Empty file" for me in the pull request, but yes, I absolutely see it when I view it on your personal branch

jaypipes (approved):
:+1:

PR #640, opened 2021-01-13
Resource configuration to generate CRD shortnames/aliases

A-Hilaly:
Issue N/A

Description of changes:
- Generate CRD shortnames using resource generator configuration

### DIFF
--- a/pkg/generate/ack/apis.go
+++ b/pkg/generate/ack/apis.go
@@ -37,7 +37,9 @@ var (
 		"apis/type_def.go.tpl",
 	}
 	apisCopyPaths = []string{}
-	apisFuncMap   = ttpl.FuncMap{}
+	apisFuncMap   = ttpl.FuncMap{
+		"Join": strings.Join,
+	}
 )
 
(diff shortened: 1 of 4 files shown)


RedbackThomson (approved):

jaypipes (approved):
Very cool. ++ :)

RedbackThomson:
Amine and I had some discussion about possible issues with short names, noting here for future reference:
- The cluster will not accept duplicate short names
- We currently do not have any tests that install all CRDs (that would catch duplicate shortnames)
- CRDs cannot be installed in namespaces, so all CRDs must have unique shortnames from any other
- We cannot parameterise shortnames (using Helm values, for instance) because CRD manifests are generated using `controller-gen`. Hacking a solution is possible but super gross.

So for now it seems that short names are on a first come, first serve basis

jaypipes:
@RedbackThomson good notes. Perhaps worth placing those into the code comment for the `ResourceConfig.ShortNames` attribute?

Issue #545, opened 2020-12-01
support resource relationships


jaypipes:
Some APIs -- e.g. API Gateway v2 or Sagemaker -- have a set of resources in their APIs that are related to each other. For example, in API Gateway v2, the `API` resource has a set of `Route` resources.

We need a way of describing these relationships and generating the code that pushes and pops finalizers from CRs of that Kind.

PR #643, opened 2021-01-14
crossplane: referencer and selector fields generation

muvaf:
Description of changes: Crossplane uses two additional fields for the ones that can be referenced by another CRD - `Reference` and `Selector`. This PR allows users to enter the type of the referenced object so that we automatically generate these fields. The type information is necessary for the resolver but since we're not yet generating it, only the existence of a value in `referenced_type` is used. When we generate the resolver code, we'll need type information as well.

One caveat is that any field on the CR (including the ones in `types.go`) could be a candidate as a reference field but from what I can tell, we are only able to address the top level fields in `Config` object. For now, that will be a limitation we'll live with.

Fixes https://github.com/crossplane/provider-aws/issues/484

### DIFF
--- a/pkg/generate/config/field.go
+++ b/pkg/generate/config/field.go
@@ -136,6 +136,10 @@ type FieldConfig struct {
 	// that owns the resource. This is a special field that we direct to
 	// storage in the common `Status.ACKResourceMetadata.OwnerAccountID` field.
 	IsOwnerAccountID bool `json:"is_owner_account_id"`
+	// ReferencedType is the Group Version Kind of another CRD that can be referenced
+	// to get the value of this field. The format is <group>/<version>.<kind>.
+	// For example: "sns/v1alpha1.Topic"
+	ReferencedType *string `json:"referenced_type,omitempty"`
 	// From instructs the code generator that the value of the field should
 	// be retrieved from the specified operation and member path
 	From *SourceFieldConfig `json:"from,omitempty"`
--- a/templates/crossplane/apis/crd.go.tpl
+++ b/templates/crossplane/apis/crd.go.tpl
@@ -7,7 +7,7 @@ package {{ .APIVersion }}
 import (
 {{- if .CRD.TypeImports }}
 {{- range $packagePath, $alias := .CRD.TypeImports }}
-    {{ if $alias }}{{ $alias }} {{ end }}"{{ $packagePath }}"
+		{{ if $alias }}{{ $alias }} {{ end }}"{{ $packagePath }}"
 {{ end }}
 
 {{- end }}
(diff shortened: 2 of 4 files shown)


muvaf:
Here is an example output https://github.com/crossplane/provider-aws/pull/493/files#diff-69196dc22922f6f3397e1cef87e44ae6594c4e6d863fa1d4aa042f45fc89efebR35

muvaf:
@jaypipes I don't know the plans regarding referencing in ACK but I think this new config field could be useful for ACK pipeline, too.

jaypipes:
> @jaypipes I don't know the plans regarding referencing in ACK but I think this new config field could be useful for ACK pipeline, too.

Hi @muvaf! @surajkota and I have been having conversations about how to represent relationships between resources. #545 is the main GH issue we will use to track this functionality, and Suraj should be pushing a PR soon with a proposal for this.

templates/crossplane/apis/crd.go.tpl
```diff
@@ -21,15 +21,29 @@ type {{ .CRD.Kind }}Parameters struct {
...
+	{{ $field.Names.Camel }} {{ $field.GoType }} `json:"{{ $field.Names.CamelLower }},omitempty"`
+
+	// {{ $field.Names.Camel }}Ref is a reference to an {{ $field.ReferencedType }} used
+	// to set the {{ $field.Names.Camel }} field.
+	// +optional
+	{{ $field.Names.Camel }}Ref {{if Contains $field.GoType "[]" -}}[]xpv1.Reference{{- else -}}*xpv1.Reference{{- end -}} `json:"{{ $field.Names.CamelLower }}Ref,omitempty"`
+
+	// {{ $field.Names.Camel }}Selector selects references to {{ $field.ReferencedType }}
+	// used to set the {{ $field.Names.Camel }}.
+  // +optional
```
jaypipes, line 38: indent :)

pkg/generate/config/field.go
```diff
@@ -136,6 +136,10 @@ type FieldConfig struct {
 	// that owns the resource. This is a special field that we direct to
 	// storage in the common `Status.ACKResourceMetadata.OwnerAccountID` field.
 	IsOwnerAccountID bool `json:"is_owner_account_id"`
+	// ReferencedType is the Group Version Kind of another CRD that can be referenced
+	// to get the value of this field. The format is <group>/<version>.<kind>.
+	// For example: "sns/v1alpha1.Topic"
+	ReferencedType *string `json:"referenced_type,omitempty"`
```
jaypipes, line 142: What about using the `Type` field being introduced in this PR?

https://github.com/aws/aws-controllers-k8s/pull/636

templates/crossplane/apis/crd.go.tpl
```diff
@@ -21,15 +21,29 @@ type {{ .CRD.Kind }}Parameters struct {
...
+	{{ $field.Names.Camel }}Selector *xpv1.Selector `json:"{{ $field.Names.CamelLower }}Selector,omitempty"`
+	{{- else if $field.IsRequired }}
```
jaypipes, line 40: Does the above mean that no referenced type may ever be required?

templates/crossplane/apis/crd.go.tpl
```diff
@@ -21,15 +21,29 @@ type {{ .CRD.Kind }}Parameters struct {
...
+	// +kubebuilder:validation:Required
 	{{ $field.Names.Camel }} {{ $field.GoType }} `json:"{{ $field.Names.CamelLower }}"`
-	{{- else }} {{ $field.Names.Camel }} {{ $field.GoType }} `json:"{{ $field.Names.CamelLower }},omitempty"` {{ end }}
+	{{- else }}
+	{{ $field.Names.Camel }} {{ $field.GoType }} `json:"{{ $field.Names.CamelLower }},omitempty"`
+	{{- end }}
 {{- end }}
-	Custom{{ .CRD.Kind }}Parameters `json:",inline"`
```
jaypipes, line 32: did you mean to remove the CustomXXXParameters field inheritance?

muvaf: Yes, we needed only for the additional referencer fields and now that we generate them, it's not necessary anymore.

templates/crossplane/apis/crd.go.tpl
muvaf: Yes, for the referenced fields, we don't use CRD validation because having no value is valid, i.e. CR creation should succeed if `vpcId` is required but only `vpcIdRef` is given because we'll resolve it before doing the call to AWS.

pkg/generate/config/field.go
muvaf: I am not sure but from what I can tell, `Type` is used to override the type of the field, right? `ReferencedType` is the type of the referenced object. For example, `tableNameRef` resolver should [look](https://github.com/crossplane/provider-aws/blob/f554eb9/apis/dynamodb/v1alpha1/referencers.go#L37) for `Table` resources.

templates/crossplane/apis/crd.go.tpl
muvaf: Ugh it's Goland again, thanks for the heads-up. Fixed it.

templates/crossplane/apis/crd.go.tpl
muvaf: Actually I think we still might need it. Some calls are not recognized by ACK, like additional calls of a resource besides the main ones. I'll add it back.

muvaf:
@jaypipes I added `Custom*Parameters` inline struct back. It should be ready for another review.

muvaf:
I've moved this PR to the new repo https://github.com/aws-controllers-k8s/code-generator/pull/11

PR #645, opened 2021-01-14
crossplane: fix read many template

muvaf:
Description of changes: `ReadMany` template was using `.Items` subfield for all `ReadMany` outputs but that's not true for all `ReadMany` resources. For example, RDS `DBCluster` has `.DBClusters` in the returned object. ACK already addresses this in set output code but Crossplane does a filtering before that step so we needed to check whether the filtered list still has any items.

Fixes https://github.com/crossplane/provider-aws/issues/492

### DIFF
--- a/pkg/generate/code/set_resource.go
+++ b/pkg/generate/code/set_resource.go
@@ -290,6 +290,21 @@ func SetResource(
 	return out
 }
 
+func ListMemberNameInReadManyOutput(
+	r *model.CRD,
+) string {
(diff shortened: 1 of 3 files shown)


jaypipes (approved):
++ looks fine to me, thanks Muvaffak.

muvaf:
@jaypipes thanks for the review! I just rebased and it should be good to go.

jaypipes (approved):
:+1:

PR #654, opened 2021-01-20
Update crossplane controller template

muvaf:
Description of changes: There are cases where isUpToDate might make an API call which can result in error, so developers should be able to return that error.

Fixes https://github.com/crossplane/provider-aws/issues/501


Henni (approved):
LGTM

jaypipes (approved):
:+1:

PR #659, opened 2021-01-22
SageMaker: Add Custom Callbacks for TrainingJob, Model and ProcessingJob Update Ops

mbaijal:
### Description of changes:
1. Add Custom Callbacks for TrainingJob, Model and ProcessingJob Update No-op since these resources do not support updates as of aws_sdk_go 1.34.32. This change fixes the issue where we saw a `not implemented` in controller log for update actions. 
Note: However the latest version of gosdk v1 does support updateTrainingJob but this needs to be a part of a separate PR.

2. Adds back the failureReason as a printer column for TrainingJob (I am not sure why the order of these fields keeps changing across different controller-gen versions)

### Testing:
Ran the integ tests and checked the logs to ensure no errors 

### DIFF
--- a/services/sagemaker/apis/v1alpha1/processing_job.go
+++ b/services/sagemaker/apis/v1alpha1/processing_job.go
@@ -57,8 +57,8 @@ type ProcessingJobStatus struct {
 // ProcessingJob is the Schema for the ProcessingJobs API
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="FailureReason",type=string,JSONPath=`.status.failureReason`
 // +kubebuilder:printcolumn:name="ProcessingJobStatus",type=string,JSONPath=`.status.processingJobStatus`
+// +kubebuilder:printcolumn:name="FailureReason",type=string,JSONPath=`.status.failureReason`
 type ProcessingJob struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
--- a/services/sagemaker/pkg/resource/model/custom_update_api.go
+++ b/services/sagemaker/pkg/resource/model/custom_update_api.go
@@ -1,0 +1,29 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package model
+
+import (
+	"context"
+	ackcompare "github.com/aws/aws-controllers-k8s/pkg/compare"
+)
+
+// TrainingJob has no update op (as of aws_sdk_go_v1.34.32)
+func (rm *resourceManager) customUpdateModel(
+	ctx context.Context,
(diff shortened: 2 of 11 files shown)


services/sagemaker/apis/v1alpha1/processing_job.go
```diff
@@ -57,8 +57,8 @@ type ProcessingJobStatus struct {
 // ProcessingJob is the Schema for the ProcessingJobs API
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="FailureReason",type=string,JSONPath=`.status.failureReason`
 // +kubebuilder:printcolumn:name="ProcessingJobStatus",type=string,JSONPath=`.status.processingJobStatus`
+// +kubebuilder:printcolumn:name="FailureReason",type=string,JSONPath=`.status.failureReason`
```
jaypipes, line 61: in order to prevent this kind of change, we need to sort the slice of `PrinterColumn` struct pointers.

Instead of iterating directly over the `CRD.AdditionalPrinterColumns` field:

https://github.com/aws/aws-controllers-k8s/blob/main/templates/apis/crd.go.tpl#L44

we need to make the AdditionalPrinterColumns field unexported and add an `AdditionalPrinterColumns` exported method of CRD that sorts the slice of `PrinterColumn` structs by the `Name` field of the `PrinterColumn` struct.

/cc @RedbackThomson

services/sagemaker/pkg/resource/model/custom_update_api.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// TrainingJob has no update op (as of aws_sdk_go_v1.34.32)
+func (rm *resourceManager) customUpdateModel(
+	ctx context.Context,
+	desired *resource,
+	latest *resource,
+	diffReporter *ackcompare.Reporter,
+) (*resource, error) {
+	return latest, nil
```
jaypipes, line 28: s/TrainingJob/Model/

If the Model doesn't have an update op, why aren't we keeping the default behaviour of returning a NotImplemented error into the reconciler loop?

services/sagemaker/pkg/resource/processing_job/custom_update_api.go
```diff
@@ -0,0 +1,29 @@
...
+// permissions and limitations under the License.
+
+package processing_job
+
+import (
+	"context"
+	ackcompare "github.com/aws/aws-controllers-k8s/pkg/compare"
+)
+
+// TrainingJob has no update op (as of aws_sdk_go_v1.34.32)
```
jaypipes, line 21: ProcessingJob.

Also, same comment as above :)

services/sagemaker/pkg/resource/model/custom_update_api.go
mbaijal: @surajkota @RedbackThomson Jay's point makes sense. This change was in keeping with the current operator experience. Opinions ?

RedbackThomson: Is there a clean way to report back to the user that this operation is not supported?

jaypipes: Unfortunately, I think the "cleanest" way is the default behaviour of returning a NotImplemented error into the reconciler loop and relying on the error logging that occurs already.

surajkota: Should we set a condition for such cases then? maybe OutOfSync? because if a user updates this resource and it cannot reach the desired state

jaypipes: @surajkota yeah, we should probably just consider NotImplemented to be a Terminal Condition...

mbaijal:
@jaypipes this PR needs to be recreated to the new repo (only) as well, correct ?

PR #670, opened 2021-01-29
Proposal: Support Resource Relationships

surajkota:
Related issues: [#545](https://github.com/aws/aws-controllers-k8s/issues/545), [crossplane/provider-aws#484](https://github.com/crossplane/provider-aws/issues/484)

Description of changes:
- Proposal to support resource relationships
- Inviting community participation


docs/design/proposals/references/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+### Modification to CRDs
+Currently, each top level resource in API model of the AWS service has a CRD. The field types in the CRD are structs, lists or primitive datatypes. A resource in K8s can be identified by a combination of metadata - `APIVersion, Kind, Name and Namespace`. Since AWS APIs have a deterministic input shape, APIVersion and Kind can be fixed at code generation time while Namespace and Name of resource will be a user input at runtime. To facilitate this, we propose to extend the K8s style of referencing resources in ACK by introducing a new datatype `AWSResourceReference` in `apis/core/v1alpha1/common.go`:
+
+```
+// AWSResourceReference represents a reference to another CR which refers to a AWS resource in Kubernetes 
+type AWSResourceReference struct {
+    // Namespace of the resource being referred to
+    Namespace string `json:"namespace"`
```
muvaf, line 19: In Kubernetes, namespaces are security boundary. So, the ability to reference an object in another namespace might result in unexpected consequences. In fact, this is one of the main reasons that managed resources are cluster-scoped in Crossplane, so that they can be shared and everyone knows they can be shared, similar to how you can use other team's resources in AWS console.

We actually had a [chat](https://kubernetes.slack.com/archives/C0EG7JC6T/p1611967173070000) in sig-api-machinery about this because I was looking for a more explicit statement in the API conventions and now there is an [open PR](https://github.com/kubernetes/community/pull/5455) to Kubernetes API conventions doc listing the reasons why references across namespaces should be avoided. A concrete example is the ownership relation built by object reference, i.e. `ownerReference`. It's [explicitly forbidden](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) for that reference to span across namespaces.

It has some consequences in controller deployment as well. It allows a controller to see and fetch values from other resources that it otherwise might not have permission for. For example, I set up a SageMaker controller in my namespace `A` and reference a VPC in namespace `B`:
* The SageMaker controller has to have `VPC` RBAC permissions either cluster-wide or in namespace `B` so that it can query that custom resource.
* The IAM permissions of SageMaker controller may or may not cover the permissions to know about the `VPC` in namespace `B`, i.e. it could be in another AWS account or region or just simply no `describe-vpc` permission on SageMaker IAM role.

That kinds of impersonation on controller level has to happen for any kind of referencing mechanism since it's the controller that fetches the value, but in native Kubernetes controllers it's always either in the same namespace or namespace->cluster in order not to breach the namespace boundary expectation of users.

To conclude, I'd suggest having only the `Name` in this struct so that teams do not end up referencing resources in other namespaces while the owner of those namespaces feel a false sense of security.

docs/design/proposals/references/README.md
```diff
@@ -0,0 +1,243 @@
...
+  * Drift detection
+  * Detect destructive operations [#82](https://github.com/aws/aws-controllers-k8s/issues/82)
+
+#### Limitation (to be solved):
+
+Resource exists in cache/etcd not a sufficient condition if X refers to a field in Y.Spec. E.g.
+  * Endpoint resource (X) refers to Spec.ModelName from resource (Y). User submitted X & Y to cluster. Y.Spec has modelName when the CR is submitted but resource does not get created in AWS because of error. Since X is only looking for is Y to exist in shared informer cache, X will run create API by looking at Y.Spec.ModelName & will go into an exponential back off instead of periodic requeue
+  * If X had a dependency on Y's ARN/id (returned by AWS) then it would have requeueAfter() until Y.Status.*.ARN != nil
+
+Although, we do not want to complicate things by mixing the state of the referred-to resource with the state of the referencing resource, if the reference is created from a spec field, it needs to be treated like ARN/id. Is there a better way to solve this ?
```
jaypipes, line 163: Right, exactly what I am saying in my comment above :)

docs/design/proposals/references/README.md
```diff
@@ -0,0 +1,243 @@
...
+    * This is a referenced resource. [Updating configmap does not update the deployment/restart the pods](https://github.com/kubernetes/kubernetes/issues/22368). Doesnt look like there will be a fix as well (issue opened 4 years ago). There are [custom controllers](https://github.com/stakater/Reloader) to deal with this issue.
+
+* ReplicaSet in Deployment:
+    * This is not a referenced resource but controller watches its resources. If you delete a pod from the replicaset, it gets recreated. This cannot happen unless controller is continuously watching the pods.
+    * ```
+      k create deploy -—image nginx nginx
+      k scale deploy nginx -—replicas=3
+      k get pods
+      k delete pod <>
+      ```
```
jaypipes, line 178: In the case of the ConfigMap referred to in a Deployment being changed and the Deployment controller not taking any action, that is a tricky issue indeed because there is a camp of people that *want* the Deployment controller to terminate the Pods that refer to the old ConfigMap value (and have the Deployment controller naturally spawn replacement Pods that will have the latest version of the ConfigMap).

And there is a camp of people that do *not want* the Deployment controller to do that and instead want ConfigMaps to be content-addressable and immutable -- in other words, if a Pod is launched by a Deployment that refers (via the Pod template) to ConfigMap with content at version X, then all Pods in that Deployment controller should be launched referring to that specific content in ConfigMap version X. Only if the Deployment's Pod template ConfigMap version changes should the deployment controller delete Pods and have them reload with the newly-referred ConfigMap version.

I would prefer that this proposal focus mostly on the *structure* of the AWSResourceReference object and walk through whether the fields contained in an AWSResourceReference will be sufficient for a controller to actuate some behaviour based on changes to the desired state of a referring resource or the observed state of a referred-to resource. Let's not get too in the weeds about implementation details about how controllers should/would respond to those changes.

docs/design/proposals/references/README.md
```diff
@@ -0,0 +1,243 @@
...
+* APIVersion and Kind:
+    * Kind is pre-determined based on the AWS API parameter. So we dont need this to be input by the user
+    * APIVersion needs to be fixed at code generation time so that controller is not affected by updates to another service controller. Kubernetes follows a [hub and spoke design](https://book.kubebuilder.io/multiversion-tutorial/conversion-concepts.html#what-does-that-have-to-do-with-webhooks) allows converting CRs from one version to other using webhooks.
+
+```
+# Initial draft, expect to be changed
+
+resources:
+  Endpoint:
+    references:
```
muvaf, line 122: I am wondering whether we can find a way to get this information from the API spec, i.e. the model JSON, somehow. I know our friends at Azure Service Operator are trying to convince their teams to share some kind of metadata, like `vpcId points to VPC`, in the OpenAPI schema they publish.

I can imagine that model JSON is published for many SDKs to use is pretty constrained due to having to be very generic, but from service owner perspective, it might be more manageable for other tooling they need to maintain to have the reference metadata in the source of truth. Though it may not be worth it either since the change there is probably higher cost than here.

jaypipes: @muvaf yeah, I've considered this. If all the AWS API models were consistent, this would be possible... :) However, the reality is that the APIs have inconsistent patterns in how they refer to other objects, which makes this kind of thing pretty difficult to standardize on.

vijtrip2:
This proposal was carried forward in PR #1057

PR #677, opened 2021-02-04
Elasticache: initial support for User

echen-98:
Description of changes:

- Terminal codes
- Additional `Status` fields for statekeeping: Elasticache's CreateUser and ModifyUser operations return an AccessString in a different representation than the AccessString requested by the user. We have to compensate for this in some way, as the generated code does not handle this case. The current solution is to add a `Status.LastAppliedAccessString` to determine when the desired AccessString has changed and `Status.ResponseAccessString` so that the user can see the server representation of the AccessString (without using the AWS CLI). Another possible solution would be to allow patching to apply to an object's `Spec` field, although if I'm not mistaken this was explicitly disabled in an earlier PR.
- Custom set output and modify logic to populate these additional `Status` fields and use them to determine when to invoke a ModifyUser versus when to do nothing.

Behavior verified via manual testing; automated tests and other changes to follow in later PRs.

### DIFF
--- a/services/elasticache/apis/v1alpha1/user.go
+++ b/services/elasticache/apis/v1alpha1/user.go
@@ -44,10 +44,12 @@ type UserStatus struct {
 	// contains a collection of `ackv1alpha1.Condition` objects that describe
 	// the various terminal states of the CR and its backend AWS service API
 	// resource
-	Conditions     []*ackv1alpha1.Condition `json:"conditions"`
-	Authentication *Authentication          `json:"authentication,omitempty"`
-	Status         *string                  `json:"status,omitempty"`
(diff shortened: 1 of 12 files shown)


jaypipes (changes requested):
@echen-98 you will need to split this PR into a PR against the github.com/aws-controllers-k8s/elasticache-controller repository (all changes to the original `/services/elasticache` directory) and a PR to this repository containing only the e2e Python test changes.

Lemme know if you need assistance with this.

echen-98:
@jaypipes sounds good, in the process of setting up my new workspace right now. We've also decided to park work on the User resource for now, so I'll close this PR as well

PR #680, opened 2021-02-07
Add support for non-default AWS profile

abyss:
Description of changes:
Currently, the kind-test uses the `daws` script to run AWS cli commands. However, it does not respect $AWS_PROFILE, and instead always uses default.

This pull request changes it, such that if $AWS_PROFILE is set, it will be passed as an environment variable to the Docker run.

There is a tiny bit of wonkiness so that it does not pass anything when $AWS_PROFILE is not set - because in testing, I could not get the --profile option (which may never be used in this setting, either way) to override the environment variable even if it's set to empty.


jaypipes (approved):
same PR merged in code-generator repo, too. See discussion on aws-controllers-k8s/code-generator#13

PR #682, opened 2021-02-09
Allow reference to local runtime module from service controller go.mod file during kind-test

kumargauravsharma:
Issue #, if available:

During development, workspace directory setup looks like:
Example setup including ElastiCache service controller along with ACK packages:
```
aws-controllers-k8s
├── code-generator
├── community
├── elasticache-controller
└── runtime
```
To build service controller image against local changes in ```runtime``` module, `elasticache-controller/go.mod` file contains:
```
replace github.com/aws-controllers-k8s/runtime => ../runtime
```

However to run local kind tests (using ```SERVICE=elasticache make kind-test``` from `community` directory) , the controller docker image needs to be built by referring to local `runtime` module from the Dockerfile.

Description of changes:

Created a ```Dockerfile.localmodules``` file that refers ```runtime``` module during build/
Updated ```community/scripts/build-controller-image.sh``` to use ```Dockerfile.localmodules``` if local workspace contains ```runtime``` module.

Note: It is upto the `elasticache-controller/go.mod` to use `runtime` module from local workspace or from its github version; ```Dockerfile.localmodules``` just makes it available during built.

### DIFF
--- a/Dockerfile.local
+++ b/Dockerfile.local
@@ -1,0 +1,60 @@
+# Build the manager binary
+FROM golang:1.14.1 as builder
+
+ARG service_alias
+# The tuple of controller image version information
+ARG service_controller_git_version
+ARG service_controller_git_commit
+ARG build_date
+# The directory within the builder container into which we will copy our
+# service controller code.
+ARG work_dir=/github.com/aws-controllers-k8s/$service_alias-controller
+WORKDIR $work_dir
+# For building Go Module required
+ENV GOPROXY=https://proxy.golang.org,direct
+ENV GO111MODULE=on
+ENV GOARCH=amd64
+ENV GOOS=linux
+ENV CGO_ENABLED=0
+ENV VERSION_PKG=github.com/aws-controllers-k8s/$service_alias-controller/pkg/version
+# Copy the Go Modules manifests and LICENSE/ATTRIBUTION
+COPY $service_alias-controller/LICENSE $work_dir/LICENSE
+COPY $service_alias-controller/ATTRIBUTION.md $work_dir/ATTRIBUTION.md
+# use local mod files
+COPY $service_alias-controller/go.local.mod $work_dir/go.local.mod
+COPY $service_alias-controller/go.local.sum $work_dir/go.local.sum
+COPY $service_alias-controller/go.mod $work_dir/go.mod
+
+# for local development, copy local runtime repo
+COPY runtime/pkg $work_dir/../runtime/pkg
+COPY runtime/apis $work_dir/../runtime/apis
+COPY runtime/go.mod $work_dir/../runtime/go.mod
+COPY runtime/go.sum $work_dir/../runtime/go.sum
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+WORKDIR $work_dir/../runtime
+RUN  go mod download
+WORKDIR $work_dir
+RUN  go mod download
+
+# Now copy the go source code for the service controller...
+COPY $service_alias-controller/apis $work_dir/apis
+COPY $service_alias-controller/cmd $work_dir/cmd
+COPY $service_alias-controller/pkg $work_dir/pkg
+# Build
+RUN GIT_VERSION=$service_controller_git_version && \
+    GIT_COMMIT=$service_controller_git_commit && \
+    BUILD_DATE=$build_date && \
+    go build -ldflags="-X ${VERSION_PKG}.GitVersion=${GIT_VERSION} \
+    -X ${VERSION_PKG}.GitCommit=${GIT_COMMIT} \
+    -X ${VERSION_PKG}.BuildDate=${BUILD_DATE}" \
+    -modfile="${work_dir}/go.local.mod" \
+    -a -o $work_dir/bin/controller $work_dir/cmd/controller/main.go
+
+FROM amazonlinux:2
+ARG service_alias
+ARG work_dir=/github.com/aws-controllers-k8s/$service_alias-controller
+WORKDIR /
+COPY --from=builder $work_dir/bin/controller $work_dir/LICENSE $work_dir/ATTRIBUTION.md /bin/
+ENTRYPOINT ["/bin/controller"]
--- a/scripts/build-controller-image.sh
+++ b/scripts/build-controller-image.sh
@@ -5,8 +5,10 @@ set -eo pipefail
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 SCRIPTS_DIR=$DIR
 ROOT_DIR=$DIR/..
-DOCKERFILE_PATH=$DIR/../Dockerfile
+DOCKERFILE_PATH=$ROOT_DIR/Dockerfile
+ACK_DIR=$ROOT_DIR/..
 DOCKERFILE=${DOCKERFILE:-"$DOCKERFILE_PATH"}
+LOCAL_MODULES=${LOCAL_MODULES:-"false"}
 BUILD_DATE=$(date +%Y-%m-%dT%H:%M)
 QUIET=${QUIET:-"false"}
 
@@ -28,6 +30,8 @@ Example: $(basename "$0") ecr
 
 Environment variables:
   QUIET:                            Build controller container image quietly (<true|false>)
+                                    Default: false
+  LOCAL_MODULES:                    Enables use of local modules during AWS Service controller docker image build
                                     Default: false
   AWS_SERVICE_DOCKER_IMG:           Controller container image tag
                                     Default: aws-controllers-k8s:\$AWS_SERVICE-\$VERSION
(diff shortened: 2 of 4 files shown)


scripts/build-controller-image.sh
```diff
@@ -70,15 +71,24 @@ if [[ $QUIET = "false" ]]; then
...
 fi
 
+# if workspace has local 'runtime' module
+# then use Dockerfile which allows references to local module from service controller go.mod file
+if [[ -d "${ACK_DIR}/runtime" ]]; then
+  DOCKER_BUILD_CONTEXT="$ACK_DIR"
+  DOCKERFILE="${ROOT_DIR}"/Dockerfile.localmodules
+else
+  DOCKER_BUILD_CONTEXT="$SERVICE_CONTROLLER_SOURCE_PATH"
+fi
```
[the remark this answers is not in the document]
A-Hilaly: The reasons I'm saying this are:
- Building the official controller images should not use the local `runtime` repository, but rather a specific published version/commit.
- To develop controllers we will be, most likely, doing changes to the `controller-gen` and rarely having to change something in `runtime`.
- It is possible to forget a local repository unsynced/out of date/unrebased, which might be a little hard to debug if users are not aware of this behaviour.

Dockerfile.localmodules
```diff
@@ -0,0 +1,54 @@
+# Build the manager binary
```
A-Hilaly, line 1: nit: Maybe we can just name it `Dockerfile.local` ?

Dockerfile.localmodules
```diff
@@ -0,0 +1,54 @@
...
+
+# for local development, copy local runtime repo
+COPY runtime/pkg $work_dir/../runtime/pkg
+COPY runtime/apis $work_dir/../runtime/apis
+COPY runtime/go.mod $work_dir/../runtime/go.mod
+COPY runtime/go.sum $work_dir/../runtime/go.sum
+
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
+RUN  go mod download
```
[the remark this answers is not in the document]
jaypipes: Yes, we should run `go mod download` from both the copied-in runtime folder as well as the top-level working directory (for the service controller itself).

scripts/build-controller-image.sh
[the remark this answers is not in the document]
jaypipes: Agree with @A-Hilaly here. The local development option should be explicitly-stated.

@kumargauravsharma, if you want to create some additional helper Makefile targets like `local-kind-test` that set `LOCAL=true` environment variable, that's cool.

scripts/build-controller-image.sh
```diff
@@ -70,15 +71,24 @@ if [[ $QUIET = "false" ]]; then
     echo " git commit: $SERVICE_CONTROLLER_GIT_COMMIT"
 fi
 
+# if workspace has local 'runtime' module
+# then use Dockerfile which allows references to local module from service controller go.mod file
+if [[ -d "${ACK_DIR}/runtime" ]]; then
+  DOCKER_BUILD_CONTEXT="$ACK_DIR"
```
jaypipes, line 77: I don't quite understand why you need a separate build context for `Dockerfile.local`. Don't we want to keep the build context pointed at the service controller's directory?

scripts/build-controller-image.sh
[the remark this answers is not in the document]
kumargauravsharma: Considering that
> It is upto the `elasticache-controller/go.mod` to use runtime module from local workspace or from its github version; `Dockerfile.localmodules` just makes the local `runtime` directory available during built.

The use of local vs remote `runtime` is controlled by the consumer module `elasticache-controller/go.mod` file and not by `Dockerfile.localmodules`.

Please review.

scripts/build-controller-image.sh
kumargauravsharma:
```
aws-controllers-k8s
├── code-generator
├── community
├── elasticache-controller
└── runtime
```
To copy the `runtime` during docker build, I supplied the build context as path leading upto `aws-controllers-k8s`

Dockerfile.localmodules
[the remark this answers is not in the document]
kumargauravsharma: Updated `Dockerfile.local` file.

Dockerfile.localmodules
kumargauravsharma: Renamed the file.

scripts/build-controller-image.sh
jaypipes: I think what @A-Hilaly and I were saying is that the way this is currently written, if the scripts/build-controller-image.sh script is run by a developer locally when they have the set of repositories git clone'd, then the Dockerfile.local will *always* be used. And Amine and I were suggesting to use a specific `LOCAL` variable to trigger using the Dockerfile.local instead of just checking for whether the `../runtime` directory existed.

That way, I can run ./scripts/build-controller-image.sh locally (even with a git clone'd runtime repository) and build a controller image that uses the runtime version from its go.mod file.

kumargauravsharma: Considering that

> It is upto the `elasticache-controller/go.mod` to use runtime module from local workspace or from its github version; `Dockerfile.local` just makes the local runtime directory available during build.

The use of local vs remote runtime is controlled by the consumer module elasticache-controller/go.mod file and not by Dockerfile.localmodules as the presence of `runtime` module locally during docker build will not have any impact if the go.mod file refers it from remote.

Please validate.

scripts/build-controller-image.sh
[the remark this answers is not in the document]
A-Hilaly: If i understand your point @kumargauravsharma the `go.mod` might contain a `replace` statement or not and that will control which version/commit of `runtime` is used.

However i still don't get why would we change the Dockerfile and build context, just because a repository is cloned locally.. This way we can never use the normal Dockerfile just because we have local `runtime` clone.. and if we ever want to release the official controller images we will have to delete the local `runtime` clone.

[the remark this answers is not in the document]
A-Hilaly: Another point to take in consideration is that, the build time using `Dockerfile.local` is defintly higher than the build time using `Dockerfile` because:
- We run extra `go mod download` commands
- the Docker daemon have to load more files that necessary because the Build context used for `Dockerfile.local` is way bigger than the one used for `Dockerfile`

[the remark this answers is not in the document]
A-Hilaly: IMO, ideally the default behavior should be using the normal `Dockerfile`, and only switch to `Dockerfile.local` if `../runtime` exists and `LOCAL=true`

scripts/build-controller-image.sh
kumargauravsharma: I had discussion with Jay on it. Per that discussion, considering following approach:
* Add a flag (example: `LOCAL_MODULES`) to select either `Dockerfile.local` or `Dockerfile`
* Supply the flag from makefile targets and name the targets to be explicit about the 'local' build (example: `make local-kind-test`)
* Expect service controller modules to have a `go.local.mod` and `go.local.sum` files for local builds
* Inside `Dockerfile.local` use `go.local.mod` file and `go.local.sum` files during `go build`

This PR has been updated per the above approach.

A-Hilaly: Maybe i'm missing something here but I'm not sure about the `go.local.mod` and `go.local.sum` solution, how would users maintain their "local module files", is it possible to run `go mod tidy` and expect the `go.local.sum` to change? IMO it should be the responsibility of the developers to manage and maintain their modules/dependency files.

A-Hilaly: If we are going with this solution, shall we add `go.local.mod` and `go.local.sum` to the `.gitignore` file?

Dockerfile.local
```diff
@@ -0,0 +1,60 @@
...
+ENV GO111MODULE=on
+ENV GOARCH=amd64
+ENV GOOS=linux
+ENV CGO_ENABLED=0
+ENV VERSION_PKG=github.com/aws-controllers-k8s/$service_alias-controller/pkg/version
+# Copy the Go Modules manifests and LICENSE/ATTRIBUTION
+COPY $service_alias-controller/LICENSE $work_dir/LICENSE
+COPY $service_alias-controller/ATTRIBUTION.md $work_dir/ATTRIBUTION.md
+# use local mod files
+COPY $service_alias-controller/go.local.mod $work_dir/go.local.mod
```
A-Hilaly, line 24: Go compiler doesn't recognize `go.local.mod` files, it should be copied as `go.mod`:
`COPY $service_alias-controller/go.local.mod $work_dir/go.mod`

Dockerfile.local
```diff
@@ -0,0 +1,60 @@
...
+COPY $service_alias-controller/go.local.sum $work_dir/go.local.sum
+COPY $service_alias-controller/go.mod $work_dir/go.mod
```
A-Hilaly, line 26: Unless I'm missing something, this line shouldn't exist? The dockerfile is currently copying `go.mod` and `go.local.mod`

A-Hilaly (approved):

Dockerfile.local
jaypipes: The line below:

`-modfile="${work_dir}/go.local.mod"`

makes this work.

scripts/build-controller-image.sh
jaypipes: We should probably add `go.local.sum` to the .gitignore file, yeah. But `go.local.mod` should stay checked into git. This is the Go mod file that will have the `replace` directive(s).

jaypipes (approved):
@kumargauravsharma lgtm, thank you for this. Would be great to update the .gitignore to ignore `go.local.sum` files, but otherwise, this is great.

Dockerfile.local
A-Hilaly: Ahhh, thanks for the pointer TIL!

Dockerfile.local
kumargauravsharma: Reference: https://golang.org/ref/mod#build-commands
> A file named go.mod must still be present in order to determine the module root directory, but it is not accessed

scripts/build-controller-image.sh
kumargauravsharma: Added `elasticache-controller/go.local.sum` to `.gitignore` file, reference PR:
https://github.com/aws-controllers-k8s/elasticache-controller/pull/4

Dockerfile.local
A-Hilaly: thank you! 🙇

jaypipes (approved):
Awesome, thanks @kumargauravsharma!

PR #690, opened 2021-02-16
ElastiCache controller e2e test workflow with split repo

kumargauravsharma:
Issue #, if available:

* Add unit test workflow for ElastiCache using the split repositories.

Description of changes:

* Added unit test workflow for ElastiCache using the split repositories.

### DIFF
--- a/.github/workflows/e2e-test-elasticache.yaml
+++ b/.github/workflows/e2e-test-elasticache.yaml
@@ -1,0 +1,46 @@
+name: e2e-test-elasticache
+on:
+  # Allow manual trigger of e2e tests
+  workflow_dispatch:
+
+jobs:
+  test-elasticache-controller:
+    name: test elasticache controller
+    runs-on: [self-hosted, aws-controllers-k8s]
+    steps:
+      - name: Set up Go 1.15
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+      - name: checkout community code
+        uses: actions/checkout@v2
+        with:
+          path: aws-controllers-k8s/community
+      - name: checkout code-generator code
+        uses: actions/checkout@v2
+        with:
+          repository: aws-controllers-k8s/code-generator
+          path: aws-controllers-k8s/code-generator
+      - name: checkout elasticache-controller code
+        uses: actions/checkout@v2
+        with:
+          repository: aws-controllers-k8s/elasticache-controller
+          path: aws-controllers-k8s/elasticache-controller
+      - name: build ack-generate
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-test-elasticache.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - name: install controller-gen
+        run: ./scripts/install-controller-gen.sh
+        working-directory: ./aws-controllers-k8s/community
+      - name: build elasticache controller
+        run: ./scripts/build-controller.sh $SERVICE
+        working-directory: ./aws-controllers-k8s/community
+        env:
+          SERVICE: elasticache
+      - name: install helm
+        run: ./scripts/install-helm.sh
```
jaypipes, line 42: All the calls to `./scripts/install-XXX` above won't work, unfortunately. For self-hosted runners, all these base requirements like controller-gen, helm, etc, need to be installed in the EC2 worker node's cloudinit script :( The reason is because of the way that the GH self-hosted test runner's context works and how $PATH cannot be updated to include newly-downloaded and installed tools.

The alternative would be to create a Docker image that had the specific required versions of dependencies and then just install docker in the base self-hosted GH runner image and then execute all tests within that Docker container. The problem with THAT is that running KinD inside Docker is, well, not going to work out well.

What we really need I believe is a Prow cluster that can spawn temporary Kubernetes clusters (via a warm pool or some KinD setup) that can be meted out to Prow test jobs. These Prow test jobs would then install the appropriate CRDs for the controller under test, install the service controller as a Deployment into the test cluster, and then run the tests in a Python-based system.

jaypipes (commented):
I think we should put these in the individual service controller repositories, but there are problems with the GH self-hosted runner actions, see inline.

.github/workflows/e2e-test-elasticache.yaml
kumargauravsharma: The `e2e-test.yaml`: https://github.com/aws-controllers-k8s/community/blob/main/.github/workflows/e2e-test.yaml does not have step to install `controller-gen` however this workflow used to run earlier.

Was the `controller-gen` installation done by updating `cloudinit` script for the EC2 worker node?
If so, could the same be followed for installing the required dependencies as current workaround?

kumargauravsharma: @jaypipes to unblock, I have update this PR to skip the helm test step during kind test based on environment variable.
This is considering that the past runs failed to install helm, indicating that other dependencies are available on test runners.

jaypipes (approved):
:+1:

PR #703, opened 2021-03-02
add simple create-delete non-public Rabbit test

jaypipes:
I learned quite a bit about the AmazonMQ API while putting together this
(very) simple test that creates a non-public-accessible RabbitMQ broker
and then deletes that broker after verifying it is queryable via the
AmazonMQ APIs.

The first thing I learned was that on average it takes more than 5
minutes to launch a single-instance RabbitMQ instance.

The second thing I learned was that if you create a new subnet to house
a public-accessible broker in and specify that subnet without
associating it with an internet gateway, AmazonMQ will complain that the
subnet doesn't exist even though it does exist but isn't publicly
accessible.

The third thing I learned was that you cannot delete a broker if it is
not in a RUNNING state. A broker gets to RUNNING from
CREATING_IN_PROGRESS state after 5+ minutes.

The PR I pushed to the code-generator for [generic hooks](https://github.com/aws-controllers-k8s/code-generator/pull/26) will be used to
create custom callbacks that check the Status.BrokerState field on both
Update and Delete calls to ensure we don't try to update something that
is still being created or delete something that isn't yet running.

I will add additional tests for publicly-accessible broker creation as
well as update code paths for both non and publicly-accessible brokers
once I merge the generic hooks PR in the code-generator and regen the MQ
controller.


vijtrip2 (approved):
LGTM

panazzo (commented):
Just curious and wanted to know more about the decision to have the e2e tests in this repo instead of each specific service controller repo.

jaypipes:
> Just curious and wanted to know more about the decision to have the e2e tests in this repo instead of each specific service controller repo.

@panazzo yeah, it's not ideal. :) The end goal is to have the e2e (and soak) testing framework housed here:

https://github.com/aws-controllers-k8s/test-infra

and then have e2e tests in each individual service controller.

The reason we haven't done this yet is mainly due to not wanting to copy (and therefore get out of sync over time) all of the common test and build scripts.

PR #696, opened 2021-02-23
Enhanced diagnostic logs for ElastiCache tests and fixes

kumargauravsharma:
Issue #, if available:

Looking at e2e test run failure logs for ElastiCache controller: https://github.com/aws-controllers-k8s/community/runs/1956783005?check_suite_focus=true
There seems need to enhance failure logs.

Description of changes:

* Enhanced diagnostic logs for ElastiCache tests and fix for test methods.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #704, opened 2021-03-05
Elasticache: enhance logging for test assertion failures

echen-98:
Description of changes:
Similar to #696, but also logging an AWS describe call and k8s object state where applicable (in addition to controller logs), to help us debug issues with our e2e test setup on Github Workflows

Also changing config for github workflow (script is complaining about AWS_PROFILE being unbound)

### DIFF
--- a/.github/workflows/e2e-test-elasticache.yaml
+++ b/.github/workflows/e2e-test-elasticache.yaml
@@ -37,6 +37,8 @@ jobs:
           SERVICE: elasticache
       - name: execute elasticache e2e tests
         run: |
+          export AWS_PROFILE=default
+          export AWS_REGION=$(aws configure get region --profile default)
           export TEST_HELM_CHARTS=false
           export AWS_ROLE_ARN=$(aws ssm get-parameter --name ACK_ROLE_ARN | jq -r '.Parameter.Value')
           export AWS_ROLE_ARN_ALT=$(aws ssm get-parameter --name ACK_ROLE_ARN_ALT | jq -r '.Parameter.Value')
--- a/scripts/lib/aws/elasticache.sh
+++ b/scripts/lib/aws/elasticache.sh
@@ -21,6 +21,69 @@ print_k8s_ack_controller_pod_logs() {
   kubectl logs -n ack-system "$ack_ctrl_pod_id"
 }
 
+# To be called upon test failure: prints resource state from k8s,
+#   aws resource state, and k8s controller logs. The correct aws CLI describe call is
+#   identified from the resource plural in k8s_resource_name.
+
+# log_and_exit requires 1 or 2 arguments:
+#   k8s_resource_name: the full name of the k8s resource, including the plural (e.g. "replicationgroups/rg-1")
+#   aws_name_override (optional): used in the AWS CLI describe call if the .metadata.name field does not match
+#                                 the AWS resource name
+log_and_exit() {
+  if [[ $# < 1 || $# > 2 ]]; then
+    echo "FATAL: Wrong number of arguments passed to ${FUNCNAME[0]}"
+    echo "Usage: ${FUNCNAME[0]} k8s_resource_name [aws_name_override]"
+    exit 1
+  fi
+  local k8s_resource_name="$1"
+
+  # validate and process input
+  if [[ "$k8s_resource_name" != *"/"* ]]; then
+    echo "FATAL: k8s_resource_name must be of form 'plural/resource_name'"
+    exit 1
+  fi
+  local resource_plural=$(echo $k8s_resource_name | cut -f 1 -d "/")
+  local aws_resource_name=$(echo $k8s_resource_name | cut -f 2 -d "/")
+  aws_resource_name="${2:-$aws_resource_name}" # apply aws_name_override if provided
+
+  # controller logs
+  echo
+  print_k8s_ack_controller_pod_logs
+
+  # k8s resource state
+  printf "\nk8s state of %s:\n" "$k8s_resource_name"
+  kubectl get "$k8s_resource_name" -o json
+
+  # aws CLI describe call
+  printf "\nAWS CLI describe call\n"
+  resource_plural=$(echo "$resource_plural" | tr '[:upper:]' '[:lower:]')
+  case $resource_plural in
+
+    cacheparametergroups)
+      daws elasticache describe-cache-parameter-groups --cache-parameter-group-name "$aws_resource_name"
+      ;;
+
+    cachesubnetgroups)
+      daws elasticache describe-cache-subnet-groups --cache-subnet-group-name "$aws_resource_name"
+      ;;
+
+    replicationgroups)
+      daws elasticache describe-replication-groups --replication-group-id "$aws_resource_name"
+      ;;
+
+    snapshots)
+      daws elasticache describe-snapshots --snapshot-name "$aws_resource_name"
+      ;;
+
+    *)
+      echo "FATAL: unkown resource plural $resource_plural"
+      exit 1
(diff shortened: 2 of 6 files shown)


echen-98:
Also I still can't add reviewers, I'm a member of the aws-controllers-k8s organization but I believe I also need to be added as a contributor for this repo?

.github/workflows/e2e-test-elasticache.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
           SERVICE: elasticache
       - name: execute elasticache e2e tests
         run: |
+          export AWS_PROFILE=default
```
kumargauravsharma, line 40: `AWS_REGION` is set in the script (default `us-west-2`): https://github.com/aws-controllers-k8s/community/blob/main/scripts/kind-build-test.sh#L17 thus, e2e tests can be invoked by specifying aws region.
However, `daws()` function https://github.com/aws-controllers-k8s/community/blob/main/scripts/lib/aws.sh#L27 mounts the aws credentials from the host (test runner) machine to the aws cli docker container. 
As the aws credentials on the host machine can have different default aws region specified for it, the aws command would fail to find resource, if region differs.

Thus, consider setting the value for `AWS_REGION` as environment variable in github workflow.
Also, consider modifying the `daws()` method to supply the aws region value to aws cli docker container as `AWS_DEFAULT_REGION` environment variable as:

```
export AWS_DEFAULT_REGION="$AWS_REGION"
docker run --rm -v ~/.aws:/root/.aws:z $(echo $aws_cli_profile_env) -v $(pwd):/aws "$aws_cli_img" -e AWS_DEFAULT_REGION "$@"
```

scripts/lib/aws/elasticache.sh
```diff
@@ -21,6 +21,68 @@ print_k8s_ack_controller_pod_logs() {
...
+#   k8s_resource_name: the full name of the k8s resource, including the plural (e.g. "replicationgroups/rg-1")
+#   aws_name_override (optional): used in the AWS CLI describe call if the .metadata.name field does not match
+#                                 the AWS resource name
+log_and_exit() {
+  if [[ $# < 1 || $# >2 ]]; then
+    echo "FATAL: Wrong number of arguments passed to ${FUNCNAME[0]}"
+    echo "Usage: ${FUNCNAME[0]} k8s_resource_name [aws_name_override]"
+    exit 1
+  fi
+  local k8s_resource_name=$1
```
kumargauravsharma, line 38: Please use double quotes: `local k8s_resource_name="$1"`

scripts/lib/aws/elasticache.sh
```diff
@@ -21,6 +21,68 @@ print_k8s_ack_controller_pod_logs() {
...
+# To be called upon test failure: prints resource state from k8s,
+#   aws resource state, and k8s controller logs. The correct aws CLI describe call is
+#   identified from the resource plural in k8s_resource_name.
+
+# log_and_exit requires 1 or 2 arguments:
+#   k8s_resource_name: the full name of the k8s resource, including the plural (e.g. "replicationgroups/rg-1")
+#   aws_name_override (optional): used in the AWS CLI describe call if the .metadata.name field does not match
+#                                 the AWS resource name
+log_and_exit() {
+  if [[ $# < 1 || $# >2 ]]; then
```
kumargauravsharma, line 33: nit: spaces around `>`

scripts/lib/aws/elasticache.sh
```diff
@@ -21,6 +21,68 @@ print_k8s_ack_controller_pod_logs() {
...
+  local resource_plural=$(echo $k8s_resource_name | cut -f 1 -d "/")
+  local aws_resource_name=$(echo $k8s_resource_name | cut -f 2 -d "/")
+  aws_resource_name="${2:-$aws_resource_name}" # apply aws_name_override if provided
+
+  # controller logs
+  echo
+  print_k8s_ack_controller_pod_logs
+
+  # k8s resource state
+  printf "\nk8s object state:\n"
```
kumargauravsharma, line 54: nit: please add `$k8s_resource_name` to the print message as well.

scripts/lib/aws/elasticache.sh
```diff
@@ -21,6 +21,68 @@ print_k8s_ack_controller_pod_logs() {
...
+  kubectl get "$k8s_resource_name" -o json
+
+  printf "\nAWS CLI describe call\n"
+  # aws CLI describe call
+  case $resource_plural in
```
kumargauravsharma, line 59: As the conditions do lowercase matching consider doing:
```
resource_plural=$(shell echo $(resource_plural) | tr '[:upper:]' '[:lower:]')
```
before the case statement.

kumargauravsharma:
Nice work @echen-98 .
I have added few comments. Please review.

.github/workflows/e2e-test-elasticache.yaml
echen-98: If our goal is to match the region used in `kind-build-test.sh` to the region used by `daws()`, shouldn't it be enough to add the following to `e2e-test-elasticache.yaml `? 

```
export AWS_REGION=$(aws configure get region --profile default)
```

Now that we've specified the default profile via `AWS_PROFILE` (although I'm pretty sure the default profile was being used before anyway),  and `daws()` uses the region specified in that profile, we just need to tell `kind-build-test.sh` to use that region, right?

kumargauravsharma: It would work as well, and I think its better (though it takes the flexibility of specifying the region out).
But, we shall use the profile name which is actually used on host.

echen-98: Sounds good. Just pushed a second revision with this change (along with the other requested ones).

I'll be out tomorrow, but if there are any further requested changes from you or Jay I can get to them Monday.

kumargauravsharma (approved):

echen-98:
@jaypipes could you take a look when possible?

scripts/lib/aws/elasticache.sh
```diff
@@ -21,6 +21,69 @@ print_k8s_ack_controller_pod_logs() {
...
+    replicationgroups)
+      daws elasticache describe-replication-groups --replication-group-id "$aws_resource_name"
+      ;;
+
+    snapshots)
+      daws elasticache describe-snapshots --snapshot-name "$aws_resource_name"
+      ;;
+
+    *)
+      echo "FATAL: unkown resource plural $resource_plural"
```
jaypipes, line 79: s/unkown/unknown/

jaypipes (approved):

jaypipes:
@echen-98 I presume you're working on the Python-based versions of these tests, yes?

echen-98:
> @echen-98 I presume you're working on the Python-based versions of these tests, yes?

Sort of – any new test we write should be in Python. I'm also currently fixing an issue with one of the snapshot tests, and as part of that I'm rewriting it in Python, but we haven't started an all-out effort to convert all of our existing bash tests yet.

I think this PR was mainly us trying to get our tests working on Github Actions for the moment rather than a further commitment to keeping our tests in bash.

jaypipes:
> > @echen-98 I presume you're working on the Python-based versions of these tests, yes?
> 
> Sort of – any new test we write should be in Python. I'm also currently fixing an issue with one of the snapshot tests, and as part of that I'm rewriting it in Python, but we haven't started an all-out effort to convert all of our existing bash tests yet.
> 
> I think this PR was mainly us trying to get our tests working on Github Actions for the moment rather than a further commitment to keeping our tests in bash.

Sounds good, @echen-98!

PR #706, opened 2021-03-05
Integration tests for SageMaker Transform Job

jkuruba:
### Description of changes:

Integration tests for Transform Job in SageMaker - https://github.com/aws-controllers-k8s/sagemaker-controller/pull/4

Tests: create, read, update, delete

### Testing:
Using local kind cluster and pytest

Minor bug fix: Fixed a race condition between role creation and usage by switching the role creation and data copy operations. By the time data is copied, role is created and available to use by the integration test.

### DIFF
--- a/test/e2e/sagemaker/resources/xgboost_transformjob.yaml
+++ b/test/e2e/sagemaker/resources/xgboost_transformjob.yaml
@@ -1,0 +1,19 @@
+apiVersion: sagemaker.services.k8s.aws/v1alpha1
+kind: TransformJob
+metadata:
+  name: $TRANSFORM_JOB_NAME
+spec:
+  transformJobName: $TRANSFORM_JOB_NAME
+  roleARN: $SAGEMAKER_EXECUTION_ROLE_ARN
+  modelName: $MODEL_NAME
+  transformInput:
+    contentType: text/csv
+    dataSource:
+      s3DataSource:
+        s3DataType: S3Prefix
+        s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/batch/input-data
+  transformOutput:
+    s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/batch/output
+  transformResources:
+    instanceCount: 1
+    instanceType: ml.m5.large
--- a/test/e2e/sagemaker/service_bootstrap.py
+++ b/test/e2e/sagemaker/service_bootstrap.py
@@ -16,6 +16,7 @@
 import boto3
 import json
 import logging
+import time
 
 from common.aws import get_aws_account_id, get_aws_region, duplicate_s3_contents
 from common.resources import random_suffix_name
@@ -45,6 +46,11 @@ def create_execution_role() -> str:
 
     iam_resource = iam.get_role(RoleName=role_name)
     resource_arn = iam_resource['Role']['Arn']
+
+    # There appears to be a delay in role availability after role creation
+    # resulting in failure that role is not present. So adding a delay
+    # to allow for the role to become available
+    time.sleep(10)
     logging.info(f"Created SageMaker execution role {resource_arn}")
 
     return resource_arn
--- a/test/e2e/sagemaker/tests/test_transformjob.py
+++ b/test/e2e/sagemaker/tests/test_transformjob.py
@@ -1,0 +1,160 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""Integration tests for the SageMaker TransformJob API.
+"""
+
+import boto3
+import pytest
+import logging
+from typing import Dict
+import time
+
+from sagemaker import SERVICE_NAME, service_marker, CRD_GROUP, CRD_VERSION
+from sagemaker.replacement_values import REPLACEMENT_VALUES
+from common.resources import load_resource_file, random_suffix_name
+from common import k8s
+from common.aws import get_aws_region
+from sagemaker.bootstrap_resources import get_bootstrap_resources
+
+RESOURCE_PLURAL = 'transformjobs'
+
+
+@pytest.fixture(scope="module")
+def sagemaker_client():
+    return boto3.client('sagemaker')
+
+
+@pytest.fixture(scope="module")
+def xgboost_transformjob(sagemaker_client):
+    #Create model using boto3 for TransformJob
+    transform_model_file = "s3://{d}/sagemaker/batch/model.tar.gz".format(
+            d=get_bootstrap_resources().DataBucketName)
+    model_name = random_suffix_name("xgboost-model", 32)
+
+    create_response = sagemaker_client.create_model(ModelName=model_name,
+                        PrimaryContainer={
+                                        'Image': REPLACEMENT_VALUES["XGBOOST_IMAGE_URI"],
+                                        'ModelDataUrl': transform_model_file, 
+                                        'Environment': {}
+                                    },
+                        ExecutionRoleArn=REPLACEMENT_VALUES["SAGEMAKER_EXECUTION_ROLE_ARN"]
+                       )
+    logging.debug(create_response)
+
+    #Check if the model is created successfully
+    describe_model_response = sagemaker_client.describe_model(ModelName=model_name)
+    assert describe_model_response["ModelName"] is not None
+    
+    resource_name = random_suffix_name("xgboost-transformjob", 32)
+
+    #Use the model created above
+    replacements = REPLACEMENT_VALUES.copy()
+    replacements["MODEL_NAME"] = model_name
+    replacements["TRANSFORM_JOB_NAME"] = resource_name
+
+    transformjob = load_resource_file(
+        SERVICE_NAME, "xgboost_transformjob", additional_replacements=replacements)
+    logging.debug(transformjob)
+
+    # Create the k8s resource
+    reference = k8s.CustomResourceReference(
+        CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL, resource_name, namespace="default")
+    resource = k8s.create_custom_resource(reference, transformjob)
+    resource = k8s.wait_resource_consumed_by_controller(reference)
+
+    assert resource is not None
+
+    yield (reference, resource)    
+
+    try:
+        # Delete the k8s resource if not already deleted by tests
+        k8s.delete_custom_resource(reference)
+    except:
+        pass
+
+    try:
+        # Delete the model created
+        sagemaker_client.delete_model(ModelName=model_name)
+    except:
+        pass
+
+
+
+@service_marker
+@pytest.mark.canary
+class TestTransformJob:
+    def _get_created_transformjob_status_list(self):
+        return ["InProgress"]
+    
+    def _get_stopped_transformjob_status_list(self):
+        return ["Stopped", "Stopping", "Completed"]
+
+    def _get_resource_transformjob_arn(self, resource: Dict):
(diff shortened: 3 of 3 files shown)


test/e2e/sagemaker/bootstrap_resources.py
```diff
@@ -22,13 +22,13 @@
 
 @dataclass
 class TestBootstrapResources:
-    DataBucketName: str
     ExecutionRoleARN: str
+    DataBucketName: str
```
RedbackThomson, line 26: Was this an automatic formatter change?

test/e2e/sagemaker/replacement_values.py
```diff
@@ -71,5 +71,6 @@
     "SAGEMAKER_DATA_BUCKET": get_bootstrap_resources().DataBucketName,
     "XGBOOST_IMAGE_URI": f"{XGBOOST_IMAGE_URIS[get_aws_region()]}/sagemaker-xgboost:1.0-1-cpu-py3",
     "PYTORCH_TRAIN_IMAGE_URI": f"{PYTORCH_TRAIN_IMAGE_URIS[get_aws_region()]}/pytorch-training:1.5.0-cpu-py36-ubuntu16.04",
-    "SAGEMAKER_EXECUTION_ROLE_ARN": get_bootstrap_resources().ExecutionRoleARN
+    "SAGEMAKER_EXECUTION_ROLE_ARN": get_bootstrap_resources().ExecutionRoleARN,
+    "TRANSFORM_MODEL": f"s3://{get_bootstrap_resources().DataBucketName}/sagemaker/batch/model.tar.gz"
```
RedbackThomson, line 75: For our other CRDs we hardcode the path values into the YAML files. I understand this model is not created from a YAML (you're calling boto3 directly), but I feel for consistency we should not have references to data paths in this file.

test/e2e/sagemaker/resources/xgboost_transformjob.yaml
```diff
@@ -0,0 +1,22 @@
...
+        s3DataType: S3Prefix
+        s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/batch/input-data
+  transformOutput:
+    s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/batch/output
+  transformResources:
+    instanceCount: 1
+    instanceType: ml.m5.large
+  tags:
+    - key: key
+      value: value
```
RedbackThomson, line 22: Please remove. We don't have tests for tags and these are an ugly artifact of our previous tests.

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+        (reference, _) = xgboost_transformjob
+        resource = k8s.get_resource(reference)
+        transformjob_name = resource['spec'].get('transformJobName', None)
+
+        assert transformjob_name is not None
+
+        # Delete the k8s resource.
+        k8s.delete_custom_resource(reference)
+        # TODO: This sleep could be replaced by a wait loop but this is sufficient for now.
+        time.sleep(5)
```
RedbackThomson, line 149: `delete_custom_resource` seems to wait until the custom resource is deleted from the API server before it returns. @surajkota Are these sleeps still necessary? I understand ACK does not wait for the status to change to `Stopping`, but surely once SageMaker has received the `StopTransformJob` API call, it changes the state to `Stopping` instantly?

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            transformjob = sagemaker_client.describe_transform_job(TransformJobName=transformjob_name)
+            return transformjob['TransformJobStatus']
+        except BaseException:
+            logging.error(
+                f"SageMaker could not find a transformJob with the name {transformjob_name}")
+            return None
+
+    def test_create_transformjob(self, xgboost_transformjob):
+        (reference, resource) = xgboost_transformjob
```
surajkota, line 110: unused var

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+
+        current_transformjob_status = self._get_sagemaker_transformjob_status(
+            sagemaker_client, transformjob_name)
+        expected_transformjob_status_list = self._get_created_transformjob_status_list()
+        assert current_transformjob_status in expected_transformjob_status_list
+        
+
+    def test_transformjob_has_stopped_status(self, sagemaker_client, xgboost_transformjob):
+        (reference, _) = xgboost_transformjob
+        resource = k8s.get_resource(reference)
```
surajkota, line 141: @RedbackThomson do you have any suggestion to avoid this duplicate code in every test. I also faced this issue

dont need to change this PR, just looking for options

test/e2e/sagemaker/tests/test_transformjob.py
surajkota: right, dont need sleep now. Check other test files on latest commit

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+        assert resource_transformjob_arn == expected_transformjob_arn
+
+    def test_transformjob_has_created_status(self, sagemaker_client, xgboost_transformjob):
+        (reference, _) = xgboost_transformjob
+        resource = k8s.get_resource(reference)
+        transformjob_name = resource['spec'].get('transformJobName', None)
+
+        assert transformjob_name is not None
+
+        current_transformjob_status = self._get_sagemaker_transformjob_status(
```
surajkota, line 133: nit: should we create temp variables if they are only used once?


same for arn test

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+    assert resource is not None
+
+    yield (reference, resource)    
+
+
+@service_marker
+@pytest.mark.canary
+class TestTransformJob:
+    def _get_created_transformjob_status_list(self):
+        return ["InProgress", "Completed"]
```
surajkota, line 81: `Completed` should be in _get_stopped_transformjob_status_list rather than this one

surajkota (changes requested):

test/e2e/sagemaker/bootstrap_resources.py
jkuruba: Actually, this is a fix for the race condition involving role creation and usage for integration tests. I missed porting over the change in service_bootstrap.py because I only ran local Pytest as the end to end integration tests are not working. Thanks for catching this. I fixed it below.

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Create the k8s resource
+    reference = k8s.CustomResourceReference(
+        CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL, resource_name, namespace="default")
+    resource = k8s.create_custom_resource(reference, transformjob)
+    resource = k8s.wait_resource_consumed_by_controller(reference)
+
+    assert resource is not None
+
+    yield (reference, resource)
```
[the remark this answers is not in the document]
jkuruba: Done

test/e2e/sagemaker/replacement_values.py
jkuruba: I moved it to the test file

test/e2e/sagemaker/resources/xgboost_transformjob.yaml
jkuruba: Removed it in tests for transform job for now. These are carried over from test to test when copied, so we need to fix the others too in a different review.

test/e2e/sagemaker/tests/test_transformjob.py
jkuruba: Fixed.

test/e2e/sagemaker/tests/test_transformjob.py
```diff
@@ -0,0 +1,154 @@
...
+    resource_name = random_suffix_name("xgboost-transformjob", 32)
+
+    #Use the model created above
+    replacements = REPLACEMENT_VALUES.copy()
+    replacements["MODEL_NAME"] = model_name
+    replacements["TRANSFORM_JOB_NAME"] = resource_name
+
+    transformjob = load_resource_file(
+        SERVICE_NAME, "xgboost_transformjob", additional_replacements=replacements)
+    #logging.debug(transformjob)
```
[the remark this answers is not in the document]
jkuruba: Fixed

test/e2e/sagemaker/tests/test_transformjob.py
jkuruba: Fixed

test/e2e/sagemaker/tests/test_transformjob.py
jkuruba: Fixed

test/e2e/sagemaker/bootstrap_resources.py
```diff
@@ -20,15 +20,19 @@
...
 
+"""ExecutionRoleARN needs to be created first so the role is ready to use
+before the test starts. Hence it should be the first thing to create in
+bootstrapping
+"""
 @dataclass
 class TestBootstrapResources:
-    DataBucketName: str
     ExecutionRoleARN: str
+    DataBucketName: str
```
RedbackThomson, line 30: Both of these resources should be created before the test starts? This structure purely defines the elements of the bootstrapping resources structure, it shouldn't indicate the precedence by which they're created.

test/e2e/sagemaker/service_bootstrap.py
```diff
@@ -76,10 +76,14 @@ def create_data_bucket() -> str:
...
+before the test starts. Hence it should be the first thing to create in
+bootstrapping
+"""
 def service_bootstrap() -> dict:
     logging.getLogger().setLevel(logging.INFO)
 
     return TestBootstrapResources(
-        create_data_bucket(),
         create_execution_role(),
+        create_data_bucket(),
```
RedbackThomson, line 88: `create_data_bucket()` does not rely on `create_execution_role()` - why does this order matter?

test/e2e/sagemaker/tests/test_transformjob.py
RedbackThomson: Um well it is kind of necessary to get the latest state of the resource. If we did a `get_resource` within the fixture, since the fixture is only loaded once (`scope="module"`), we would not see it updating over time.

That being said, the fixture is already returning a resource with the spec. I think most of the time when we are calling `get_resource`, it's not necessary, since we are just looking at items in the spec fields. We only need to call `get_resource` if we want to see the `status`.

test/e2e/sagemaker/service_bootstrap.py
jkuruba: I noticed is that all tests including training were failing due to role not being available when the tests ran. But they eventually succeed after the role became available. My observation is that it takes some time for the role creation and and the test starts before the role is available. Tests error out saying that role is not present. So, creating the role first gives it time to become available for the test and resolved the issue for me. The error is noticeable in the Batch Transform test because the model creation fails from the boto3 call and with this change, model creation call succeeds.

RedbackThomson: Is it not sufficient to wait for the role to become available as part of the bootstrapping process?

test/e2e/sagemaker/tests/test_transformjob.py
surajkota: Sg

jaypipes (changes requested):
I have the same requests for this PR as I do for the PR for HPO Jobs. Specifically:

https://github.com/aws-controllers-k8s/community/pull/713/files#r591983536
and
https://github.com/aws-controllers-k8s/community/pull/713/files#r591981309

RedbackThomson:
@jaypipes We would like to merge this, just so we can unblock this story from our sprint. We have a follow-up story to come through each of the integration tests and clean them up in the way you suggested.

test/e2e/sagemaker/service_bootstrap.py
jkuruba: I changed the test to use sleep and added a comment indicating that it is to work around this issue. For now, sleep is sufficient for the role to become available

surajkota (approved):

Issue #711, opened 2021-03-10
build-controller.sh script failing as ack-generate binary does not recognize flag "--templates-dir"


echen-98:
**Describe the bug**
Script will fail with message `Error: unknown flag: --templates-dir`

**Steps to reproduce**

1. check out latest code from community, code-generator, and elasticache-controller repos
2. run `make build-ack-generate`
3. run `export SERVICE=elasticache`
4. run `make build-controller`

**Expected outcome**
Expected the controller to build successfully. Currently the script is failing in building the API objects, and is calling the ack-generate binary like this:

```
+ /Users/ercche/go/src/github.com/aws-controllers-k8s/ack-community/scripts/lib/../../../../aws-controllers-k8s/code-generator/bin/ack-generate apis elasticache -o /Users/ercche/go/src/github.com/aws-controllers-k8s/elasticache-controller --templates-dir /Users/ercche/go/src/github.com/aws-controllers-k8s/ack-community/scripts/lib/../../../../aws-controllers-k8s/code-generator/templates --cache-dir '~/.cache/aws-controllers-k8s' --version v1alpha1 --generator-config-path /Users/ercche/go/src/github.com/aws-controllers-k8s/elasticache-controller/generator.yaml
```

**Environment**
Local (Mac)

* Kubernetes version
* Using EKS (yes/no), if so version? 
* AWS service targeted (S3, RDS, etc.): Elasticache (but may not be specific to elasticache)

echen-98:
Root cause: `--templates-dir` flag was renamed `--template-dirs` flag because the latter more accurately describes how you can now specify multiple paths to override the default template directory: https://github.com/aws-controllers-k8s/code-generator/pull/28/files

PR #712, opened 2021-03-10
Fix build-controller script to accommodate recent change to multiple template paths

echen-98:
Issue #, if available:
#711 
Script was broken since ack-generate was [recently changed](https://github.com/aws-controllers-k8s/code-generator/pull/28) to take multiple template paths, so it did not recognize the flag `--templates-dir` since the flag was renamed to `--template-dirs`

Description of changes:
Renamed environment variables and changed arguments passed to ack-generate.


echen-98:
@jaypipes I had a question first, though:

It seems like you already made similar changes to the `build-controller.sh` script in the code-generator repo. I took a quick look through both and it looks like they do pretty much the same thing (correct me if I'm wrong), so is there a reason we need to keep both?

scripts/build-controller.sh
```diff
@@ -141,7 +143,7 @@ config_output_dir="$SERVICE_CONTROLLER_SOURCE_PATH/config/"
 pushd $SERVICE_CONTROLLER_SOURCE_PATH/apis/$ACK_GENERATE_API_VERSION 1>/dev/null
 
 echo "Generating deepcopy code for $SERVICE"
-controller-gen object:headerFile=$TEMPLATES_DIR/boilerplate.txt paths=./...
+controller-gen object:headerFile=$TEMPLATE_DIRS/boilerplate.txt paths=./...
```
echen-98, line 146: I know this won't work for multiple paths but wanted to raise the PR and discuss the question of the two scripts first

jaypipes:
> @jaypipes I had a question first, though:
> 
> It seems like you already made similar changes to the `build-controller.sh` script in the code-generator repo. I took a quick look through both and it looks like they do pretty much the same thing (correct me if I'm wrong), so is there a reason we need to keep both?

We should really just have one script, and ideally that should be in the new dev-tools repo, but we're not quite there yet. In the meantime, probably best to just copy the scripts/build-controller.sh file to the code-generator repo.

echen-98:
> We should really just have one script, and ideally that should be in the new dev-tools repo, but we're not quite there yet. In the meantime, probably best to just copy the scripts/build-controller.sh file to the code-generator repo.

Could we try a workaround like this for the time being (see new commit)? I'm trying to unblock our failing Github workflows so it requires a remote change. I could copy the working script from code-generator into the community repo in this PR, but I feel like keeping two scripts in this manner is unsustainable even in the short term.

For example if I did the above and noticed a mistake in the usage message within code-generator's version of the script (which there is – the usage message still says `TEMPLATES-DIR` instead of `TEMPLATE-DIRS`), then I would have to create 2 additional PRs: one to update the code-generator's version, and one to copy that version into the community repo. Any other subsequent changes would also require 2 PRs for consistency.

echen-98:
Also, some of what I said above was not completely correct. The scripts cannot be copied verbatim as many of the paths set in each script are relative to the location of the script.

jaypipes (approved):
yup, this is a good temporary solution until we have `ackdev`. Thank you Eric!

PR #713, opened 2021-03-10
Integration tests for HyperParameterJob

akartsky:
### Description of changes:
Integration tests for HyperParameterTuning Job in SageMaker 

Related CR:  https://github.com/aws-controllers-k8s/sagemaker-controller/pull/5

### Testing:
Using local kind cluster and pytest

### DIFF
--- a/test/e2e/sagemaker/resources/xgboost_hpojob.yaml
+++ b/test/e2e/sagemaker/resources/xgboost_hpojob.yaml
@@ -1,0 +1,64 @@
+apiVersion: sagemaker.services.k8s.aws/v1alpha1
+kind: HyperParameterTuningJob
+metadata:
+  name: $HPO_JOB_NAME
+spec:
+  hyperParameterTuningJobName: $HPO_JOB_NAME
+  tags:
+    - key: test-key
+      value: test-value
+  hyperParameterTuningJobConfig:
+    strategy: Bayesian
+    hyperParameterTuningJobObjective:
+      type_: Minimize
+      metricName: validation:error
+    resourceLimits:
+      maxNumberOfTrainingJobs: 10
+      maxParallelTrainingJobs: 5
+    parameterRanges:
+      integerParameterRanges:
+      - name: num_round
+        minValue: '10'
+        maxValue: '20'
+        scalingType: Linear
+      continuousParameterRanges: []
+      categoricalParameterRanges: []
+    trainingJobEarlyStoppingType: Auto
+  trainingJobDefinition:
+    staticHyperParameters:
+      base_score: '0.5'
+    algorithmSpecification:
+      trainingImage: $XGBOOST_IMAGE_URI
+      trainingInputMode: File
+    roleARN: $SAGEMAKER_EXECUTION_ROLE_ARN
+    inputDataConfig:
+    - channelName: train
+      dataSource:
+        s3DataSource:
+          s3DataType: S3Prefix
+          s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/train
+          s3DataDistributionType: FullyReplicated
+      contentType: text/csv
+      compressionType: None
+      recordWrapperType: None
+      inputMode: File
+    - channelName: validation
+      dataSource:
+        s3DataSource:
+          s3DataType: S3Prefix
+          s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/validation/
+          s3DataDistributionType: FullyReplicated
+      contentType: text/csv
+      compressionType: None
+      recordWrapperType: None
+      inputMode: File
+    outputDataConfig:
+      s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/hpo_output
+    resourceConfig:
+      instanceType: ml.m4.xlarge
+      instanceCount: 1
+      volumeSizeInGB: 25
+    stoppingCondition:
+      maxRuntimeInSeconds: 3600
+    enableNetworkIsolation: true
+    enableInterContainerTrafficEncryption: false
--- a/test/e2e/sagemaker/resources/xgboost_hpojob.yaml
+++ b/test/e2e/sagemaker/resources/xgboost_hpojob.yaml
@@ -4,17 +4,14 @@ metadata:
   name: $HPO_JOB_NAME
 spec:
   hyperParameterTuningJobName: $HPO_JOB_NAME
-  tags:
-    - key: test-key
-      value: test-value
   hyperParameterTuningJobConfig:
     strategy: Bayesian
     hyperParameterTuningJobObjective:
       type_: Minimize
       metricName: validation:error
     resourceLimits:
-      maxNumberOfTrainingJobs: 10
-      maxParallelTrainingJobs: 5
+      maxNumberOfTrainingJobs: 2
+      maxParallelTrainingJobs: 1
     parameterRanges:
       integerParameterRanges:
       - name: num_round
@@ -53,9 +50,9 @@ spec:
       recordWrapperType: None
       inputMode: File
     outputDataConfig:
-      s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/hpo_output
+      s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/hpo/output
     resourceConfig:
-      instanceType: ml.m4.xlarge
+      instanceType: ml.m5.large
       instanceCount: 1
       volumeSizeInGB: 25
     stoppingCondition:
--- a/test/e2e/sagemaker/tests/test_hpo.py
+++ b/test/e2e/sagemaker/tests/test_hpo.py
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-"""Integration tests for the SageMaker TrainingJob API.
+"""Integration tests for the SageMaker HyperParameterTuning API.
 """
 
 import boto3
--- a/test/e2e/sagemaker/tests/test_hpo.py
+++ b/test/e2e/sagemaker/tests/test_hpo.py
@@ -1,0 +1,151 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""Integration tests for the SageMaker TrainingJob API.
+"""
+
+import boto3
+import pytest
+import logging
+from typing import Dict
+import time
+
+from sagemaker import SERVICE_NAME, service_marker, CRD_GROUP, CRD_VERSION
+from sagemaker.replacement_values import REPLACEMENT_VALUES
+from common.resources import load_resource_file, random_suffix_name
+from common import k8s
+
+RESOURCE_PLURAL = 'hyperparametertuningjobs'
+
+
+@pytest.fixture(scope="module")
+def sagemaker_client():
+    return boto3.client('sagemaker')
+
+
+@pytest.fixture(scope="module")
+def xgboost_hpojob():
+    resource_name = random_suffix_name("xgboost-hpojob", 32)
+
+    replacements = REPLACEMENT_VALUES.copy()
+    replacements["HPO_JOB_NAME"] = resource_name
+
+    hpojob = load_resource_file(
+        SERVICE_NAME, "xgboost_hpojob", additional_replacements=replacements)
+    logging.debug(hpojob)
+
+    # Create the k8s resource
+    reference = k8s.CustomResourceReference(
+        CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL, resource_name, namespace="default")
+    resource = k8s.create_custom_resource(reference, hpojob)
+    resource = k8s.wait_resource_consumed_by_controller(reference)
+
+    assert resource is not None
+
+    yield (reference, resource)    
+
+    # Delete the k8s resource if not already deleted by tests
+    try:
+        k8s.delete_custom_resource(reference)
+    except:
+        pass
+
+
+@service_marker
+@pytest.mark.canary
+class TestHPO:
+    def _get_created_hpo_job_status_list(self):
+        return ["InProgress", "Completed"]
+
+    def _get_stopped_hpo_job_status_list(self):
+        return ["Stopped", "Stopping"]
+
+    def _get_sagemaker_hpo_job_arn(self, sagemaker_client, hpo_job_name: str):
+        try:
+            hpo_desc = sagemaker_client.describe_hyper_parameter_tuning_job(
+                HyperParameterTuningJobName=hpo_job_name
+            )
+            return hpo_desc["HyperParameterTuningJobArn"]
+        except BaseException:
+            logging.error(
+                f"SageMaker could not find an hpo job with the name {hpo_job_name}"
+            )
+            return None
+
+    def _get_sagemaker_hpo_job_status(
+        self, sagemaker_client, hpo_job_name: str
+    ):
+        try:
+            hpo_job = sagemaker_client.describe_hyper_parameter_tuning_job(
+                HyperParameterTuningJobName=hpo_job_name
+            )
+            return hpo_job["HyperParameterTuningJobStatus"]
+        except BaseException:
+            logging.error(
+                f"SageMaker could not find an hpo job with the name {hpo_job_name}"
+            )
+            return None
+
+    def test_create_hpo(self, xgboost_hpojob):
+        (reference, resource) = xgboost_hpojob
+        assert k8s.get_resource_exists(reference)
+
+    def test_hpo_has_correct_arn(self, sagemaker_client, xgboost_hpojob):
+        (reference, _) = xgboost_hpojob
+        resource = k8s.get_resource(reference)
+        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
+
+        assert hpo_job_name is not None
+
+        assert k8s.get_resource_arn(resource) == self._get_sagemaker_hpo_job_arn(
+            sagemaker_client, hpo_job_name
+        )
+
+    def test_hpo_job_has_created_status(
+        self, sagemaker_client, xgboost_hpojob
+    ):
+        (reference, _) = xgboost_hpojob
+        resource = k8s.get_resource(reference)
+        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
+
+        assert hpo_job_name is not None
+
+        current_hpo_job_status = self._get_sagemaker_hpo_job_status(
+            sagemaker_client, hpo_job_name
+        )
+        expected_hpo_job_status_list = (
+            self._get_created_hpo_job_status_list()
+        )
+        assert current_hpo_job_status in expected_hpo_job_status_list
+
+    def test_hpo_job_has_stopped_status(
+        self, sagemaker_client, xgboost_hpojob
+    ):
+        (reference, _) = xgboost_hpojob
+        resource = k8s.get_resource(reference)
+        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
+
+        assert hpo_job_name is not None
+
+        # Delete the k8s resource.
+        _, deleted = k8s.delete_custom_resource(reference)
+        assert deleted is True
+
+        current_hpo_job_status = self._get_sagemaker_hpo_job_status(
+            sagemaker_client, hpo_job_name
+        )
+        expected_hpo_job_status_list = (
+            self._get_stopped_hpo_job_status_list()
+        )
+        assert current_hpo_job_status in expected_hpo_job_status_list
+
+
--- a/test/e2e/sagemaker/tests/test_hpo.py
+++ b/test/e2e/sagemaker/tests/test_hpo.py
@@ -25,127 +25,71 @@ from common.resources import load_resource_file, random_suffix_name
 from common import k8s
 
 RESOURCE_PLURAL = 'hyperparametertuningjobs'
+HPO_JOB_STATUS_CREATED = ("InProgress", "Completed")
+HPO_JOB_STATUS_STOPPED = ("Stopped", "Stopping")
 
-
-@pytest.fixture(scope="module")
-def sagemaker_client():
+def _sagemaker_client():
     return boto3.client('sagemaker')
 
-
-@pytest.fixture(scope="module")
-def xgboost_hpojob():
+def _make_hpojob():
     resource_name = random_suffix_name("xgboost-hpojob", 32)
 
     replacements = REPLACEMENT_VALUES.copy()
     replacements["HPO_JOB_NAME"] = resource_name
 
-    hpojob = load_resource_file(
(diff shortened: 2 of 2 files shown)


test/e2e/sagemaker/resources/xgboost_hpojob.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: sagemaker.services.k8s.aws/v1alpha1
+kind: HyperParameterTuningJob
+metadata:
+  name: $HPO_JOB_NAME
+spec:
+  hyperParameterTuningJobName: $HPO_JOB_NAME
+  tags:
+    - key: test-key
+      value: test-value
```
RedbackThomson, line 9: Please remove - see https://github.com/aws-controllers-k8s/community/pull/706#discussion_r589842022

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
```diff
@@ -0,0 +1,64 @@
...
+  hyperParameterTuningJobConfig:
+    strategy: Bayesian
+    hyperParameterTuningJobObjective:
+      type_: Minimize
+      metricName: validation:error
+    resourceLimits:
+      maxNumberOfTrainingJobs: 10
+      maxParallelTrainingJobs: 5
```
RedbackThomson, line 17: Please reduce down to 2/1 maybe. We want to reduce costs for anyone that may want to run these integration tests

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
```diff
@@ -0,0 +1,64 @@
...
+        s3DataSource:
+          s3DataType: S3Prefix
+          s3URI: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/validation/
+          s3DataDistributionType: FullyReplicated
+      contentType: text/csv
+      compressionType: None
+      recordWrapperType: None
+      inputMode: File
+    outputDataConfig:
+      s3OutputPath: s3://$SAGEMAKER_DATA_BUCKET/sagemaker/training/hpo_output
```
RedbackThomson, line 56: nit: Change to `/sagemaker/hpo/output`. I know HPO is technically running training, but it would help to differentiate the results of each integration test (if we choose to modify the HPO training job spec.

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
```diff
@@ -0,0 +1,64 @@
...
+    resourceConfig:
+      instanceType: ml.m4.xlarge
```
RedbackThomson, line 58: Reduce to smaller instance type: `ml.m5.large`

I see in Training we are still using `m4.xl`, that should probably be changed, too. `m5.large` is the cheapest.

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
```diff
@@ -0,0 +1,64 @@
...
+      instanceCount: 1
+      volumeSizeInGB: 25
+    stoppingCondition:
+      maxRuntimeInSeconds: 3600
+    enableNetworkIsolation: true
+    enableInterContainerTrafficEncryption: false
```
RedbackThomson, line 64: Is there any good reason why these flags should be enabled? I know these are enabled in our existing integration tests, but do we ever test for these features?

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+"""Integration tests for the SageMaker TrainingJob API.
```
RedbackThomson, line 13: nit: `s/TrainingJob/Hyperparameter Tuning Job/`

RedbackThomson (changes requested):

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
akartsky: we stop the job right after submitting it so shouldn't matter much 

but ya making the change

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
akartsky: done

test/e2e/sagemaker/tests/test_hpo.py
akartsky: nice catch

test/e2e/sagemaker/resources/xgboost_hpojob.yaml
akartsky: Nope. Just testing extra parameters i guess

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+        _, deleted = k8s.delete_custom_resource(reference)
+        assert deleted is True
+
+        current_hpo_job_status = self._get_sagemaker_hpo_job_status(
+            sagemaker_client, hpo_job_name
+        )
+        expected_hpo_job_status_list = (
+            self._get_stopped_hpo_job_status_list()
+        )
+        assert current_hpo_job_status in expected_hpo_job_status_list
```
jaypipes, line 149: Seems a bit wasteful to me to have a bunch of test methods and call the SageMaker DescribeHyperParameterTuningJob API call multiple times just to test different attributes are set on the CR's Status struct.

You could also pull the `_get_sagemaker_hpo_job()` method out into a module-level function, decorate it with `@pytest.fixture` like so, and eliminate the need for the caller to pass the sagemaker_client param.

Is there a reason we couldn't combine these test methods into a single test? Something like this?

```python
@pytest.fixture(scope="module")
def get_sagemaker_hpo_job(self, sagemaker_client, hpo_job_name: str):
    try:
        hpo_desc = sagemaker_client.describe_hyper_parameter_tuning_job(
            HyperParameterTuningJobName=hpo_job_name
        )
        return hpo_desc
    except BaseException:
        logging.error(
            f"SageMaker could not find an hpo job with the name {hpo_job_name}"
        )
        return None

@service_marker
@pytest.mark.canary
class TestHPO:
    def test_smoke(self, xgboost_hpojob):
        (reference, resource) = xgboost_hpojob
        assert k8s.get_resource_exists(reference)

        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
        assert hpo_job_name is not None

        hpo_job = self._get_sagemaker_hpo_job(sagemaker_client, hpo_job_name)

        assert k8s.get_resource_arn(resource) == hpo_job["HyperParameterTuningJobArn"]
        assert hpo_job["HyperParameterTuningJobStatus"] in HPO_JOB_STATUS_CREATED

        _, deleted = k8s.delete_custom_resource(reference)
        assert deleted is True

        assert hpo_job["HyperParameterTuningJobStatus"] in HPO_JOB_STATUS_STOPPED
```

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+    resource = k8s.create_custom_resource(reference, hpojob)
+    resource = k8s.wait_resource_consumed_by_controller(reference)
+
+    assert resource is not None
+
+    yield (reference, resource)    
+
+    # Delete the k8s resource if not already deleted by tests
+    try:
+        k8s.delete_custom_resource(reference)
```
jaypipes, line 58: This is deleted in one of the tests below, which actually indicates that this should not actually be a fixture (which is some test dependency that should be set up and torn down automatically by pytest) but rather something that should be created specifically for a test of the create/delete functionality as a whole (see more below)

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+@service_marker
+@pytest.mark.canary
+class TestHPO:
+    def _get_created_hpo_job_status_list(self):
+        return ["InProgress", "Completed"]
+
+    def _get_stopped_hpo_job_status_list(self):
+        return ["Stopped", "Stopping"]
```
jaypipes, line 70: Place these as simple constants in the module scope:

```python
HPO_JOB_STATUS_STOPPED = ("Stopped", "Stopping")
HPO_JOB_STATUS_CREATED = ("InProgress", "Completed")
```
see more below...

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+        k8s.delete_custom_resource(reference)
+    except:
+        pass
+
+
+@service_marker
+@pytest.mark.canary
+class TestHPO:
+    def _get_created_hpo_job_status_list(self):
+        return ["InProgress", "Completed"]
```
surajkota, line 67: https://github.com/aws-controllers-k8s/community/pull/706#discussion_r589906565

jaypipes (changes requested):
bit of cleanup to do on this, @akartsky, please see my comments inline for suggestions. thank you!

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+            return hpo_job["HyperParameterTuningJobStatus"]
+        except BaseException:
+            logging.error(
+                f"SageMaker could not find an hpo job with the name {hpo_job_name}"
+            )
+            return None
+
+    def test_create_hpo(self, xgboost_hpojob):
+        (reference, resource) = xgboost_hpojob
```
surajkota, line 99: nit: unused var

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+        assert k8s.get_resource_exists(reference)
+
+    def test_hpo_has_correct_arn(self, sagemaker_client, xgboost_hpojob):
+        (reference, _) = xgboost_hpojob
+        resource = k8s.get_resource(reference)
```
surajkota, line 104: dont need to get_resource here, see https://github.com/aws-controllers-k8s/community/pull/706#discussion_r591941021

same for other places where you need only spec fields

test/e2e/sagemaker/tests/test_hpo.py
```diff
@@ -0,0 +1,151 @@
...
+    def test_hpo_job_has_created_status(
+        self, sagemaker_client, xgboost_hpojob
+    ):
+        (reference, _) = xgboost_hpojob
+        resource = k8s.get_resource(reference)
+        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
+
+        assert hpo_job_name is not None
+
+        current_hpo_job_status = self._get_sagemaker_hpo_job_status(
```
surajkota, line 122: https://github.com/aws-controllers-k8s/community/pull/706#discussion_r589906115

surajkota (changes requested):

test/e2e/sagemaker/tests/test_hpo.py
surajkota: Intention is, even if the test fails, we clean up all the resources which cost $$. So I think author started using fixture rather than putting try catch blocks in the test method.

is this not good?

jaypipes: > (edit) now that I think about the above statement, script should be throwing error if delete fails rather than ignoring the exception because it means we have a leaking resource

Yes, exactly.

Though, my real point here is that we should be considering what the thing is that we are testing and what are those test dependencies.

What I mean by that is when testing the Create [-> Update] -> Delete code paths for objects of type X that we are not using fixtures that call the Create, Update or Delete code paths, since it is those code paths that are the exact part of the system under test that we wish to assert.

See https://docs.pytest.org/en/latest/fixture.html#what-fixtures-are for more details on this concept. There's nothing wrong with having a fixture that creates, yields and cleans up an object of type X. However, such fixtures should be dependencies of tests that are asserting behaviour and actions against objects NOT of type X.

In other words, you should feel free to make a little helper function like this:

```python
def _make_hpojob():
    resource_name = random_suffix_name("xgboost-hpojob", 32)

    replacements = REPLACEMENT_VALUES.copy()
    replacements["HPO_JOB_NAME"] = resource_name

    data = load_resource_file(
        SERVICE_NAME, "xgboost_hpojob", additional_replacements=replacements)
    logging.debug(data)

    return k8s.CustomResourceReference(
        CRD_GROUP, CRD_VERSION, RESOURCE_PLURAL, resource_name, namespace="default")
```

and call this helper function from a fixture that creates, yields and deletes an HPO Job:

```python
@pytest.fixture(scope="module")
def xgboost_hpojob():
    hpo_job = _make_hpojob()
    resource = k8s.create_custom_resource(hpo_job)
    resource = k8s.wait_resource_consumed_by_controller(hpo_job)

    yield (hpo_job, resource)    

    k8s.delete_custom_resource(hpo_job)
```

(note the above has no error suppression around the delete call since we definitely want to error out if deletion doesn't succeed or the resource has already been deleted, because for fixtures, our expectation is that the *fixture* controls the lifecycle of the object being yielded by the fixture, not the caller.

And then, to actually test the Create and Delete actions for an HPO Job, you'd want a test method that runs, in a serial fashion, the same create/delete calls along with assertions about state transitions:

```python
@service_marker
@pytest.mark.canary
class TestHPO:
    def test_smoke(self):
        hpo_job = _make_hpojob()
        resource = k8s.create_custom_resource(reference, hpo_job)
        resource = k8s.wait_resource_consumed_by_controller(reference)
        assert k8s.get_resource_exists(hpo_job)

        hpo_job_name = resource["spec"].get("hyperParameterTuningJobName", None)
        assert hpo_job_name is not None

        hpo_job = get_sagemaker_hpo_job(hpo_job_name)

        assert k8s.get_resource_arn(resource) == hpo_job["HyperParameterTuningJobArn"]
        assert hpo_job["HyperParameterTuningJobStatus"] in HPO_JOB_STATUS_CREATED

        _, deleted = k8s.delete_custom_resource(reference)
        assert deleted is True

        hpo_job = get_sagemaker_hpo_job(hpo_job_name)
        assert hpo_job["HyperParameterTuningJobStatus"] in HPO_JOB_STATUS_STOPPED
```

Now, let's say you had a test for the create and delete code paths for a **different** object of type Y and objects of type Y required an HPO Job to already exist (because, say, objects of type Y referred to an HPO Job by ID).

This is the case where you'd want to use your `xgboost_hpojob` fixture, because the create and delete code paths of the HPO Job itself are **not** the thing being tested. Instead, the create and delete code paths of Object of type Y is what is being tested. So, you'd use the `xgboost_hpojob` as a fixture in the method that tested the create and delete code paths of Objects of type Y, like so:

```python
@service_marker
@pytest.mark.canary
class TestY:
    def test_smoke(self, xgboost_hpojob):
        hpo_job_ref, _ = xgboost_hpojob
        hpo_job_name = hpo_job_ref["spec"].get("hyperParameterTuningJobName"]

        object_y = _make_object_y(hpo_job_name)
        resource = k8s.create_custom_resource(object_y)
        resource = k8s.wait_resource_consumed_by_controller(object_y)
        assert k8s.get_resource_exists(object_y)

        # bunch of assertions...

        _, deleted = k8s.delete_custom_resource(object_y)
        assert deleted is True

        # bunch of assertions... but do NOT delete the HPO Job fixture,
        # since the fixture owns the lifecycle of the HPO Job and not
        # this `TestY.test_smoke` method.
```

akartsky: oh this is neat 

In the `xgboost_hpojob` fixture we only need to  delete it if it exists else it will throw a "resource not found" error. 
This is required because some tests might delete the resource during testing.

--------

By combining these tests into 1, we lose a few thing :
- While running locally we have to wait for whole thing to pass/fail and then we see the logs. Instead of seeing each test pass/fail as we go.
- Now that we only have 1 test, it can fail because of multiple reasons and we won't be able to track that metric.

--------------

Let me know what you guys think 
Once we merge this PR i can make this same change for all of our tests

test/e2e/sagemaker/tests/test_hpo.py
akartsky: done

test/e2e/sagemaker/tests/test_hpo.py
akartsky: fixed

test/e2e/sagemaker/tests/test_hpo.py
akartsky: fixed

test/e2e/sagemaker/tests/test_hpo.py
akartsky: fixed

test/e2e/sagemaker/tests/test_hpo.py
akartsky: we cannot call `get_sagemaker_hpo_job ()` if we make that a fixture.

if we don't make it a fixture we also cannot use sagemaker_client as parameter.

so making both of these functions just normal functions instead of fixtures

test/e2e/sagemaker/tests/test_hpo.py
akartsky: we can't move this to "stoped_hpo_list" because then we won't be able to verify if delete is working.

test/e2e/sagemaker/tests/test_hpo.py
jaypipes: > oh this is neat
> 
> In the `xgboost_hpojob` fixture we only need to delete it if it exists else it will throw a "resource not found" error.
> This is required because some tests might delete the resource during testing.

There should not be assertions in fixtures. Fixtures are for test dependencies. Fixtures should handle the lifecycle of these test dependencies entirely. If a test deletes an object that is `yield`'d from a fixture and that fixture deletes the object after the `yield`, that is an indication that the **test should not be using that fixture**, because the `yield`'d object is no longer a test *dependency*, rather, the object is the *target* of the test's Actions and Assertions.

> By combining these tests into 1, we lose a few thing :
> 
>     * While running locally we have to wait for whole thing to pass/fail and then we see the logs. Instead of seeing each test pass/fail as we go.

The `assert` statements will fail-fast the test. Once something fails, the test will stop running. This is actually what you want, no? Why would you want to continue running a test about a Status attribute if the `create` call failed?

>     * Now that we only have 1 test, it can fail because of multiple reasons and we won't be able to track that metric.

What metric are you referring to here?

akartsky: In the example code `test_smoke()` 👆🏼
```
_, deleted = k8s.delete_custom_resource(object_y)
assert deleted is True
```
This will delete the resource. So this test should not be using the `xgboost_hpojob` fixture right?

-------------------
Combining the tests : 
- yes we will see the assert failures. I meant, while running tests locally if assertions are passing we will not know what stage the test is at. We will only see the logs once the whole thing is done which can take a while. (Eg: SageMaker Endpoint)
- Metrics : I meant like how long did delete test take ? or how many times update failed over time ?. 
By having create/update/delete in one test we will not be able to track individual metrics.

jaypipes: > In the example code `test_smoke()` 👆🏼
> 
> ```
> _, deleted = k8s.delete_custom_resource(object_y)
> assert deleted is True
> ```
> 
> This will delete the resource. So this test should not be using the `xgboost_hpojob` fixture right?

Correct.

> Combining the tests :
> 
>     * yes we will see the assert failures. I meant, while running tests locally if assertions are passing we will not know what stage the test is at. We will only see the logs once the whole thing is done which can take a while. (Eg: SageMaker Endpoint)

I guess I don't see this as a problem.

>     * Metrics : I meant like how long did delete test take ? or how many times update failed over time ?.

I don't see these tests as the source of metrics for things like how long a delete of a resource takes. A couple points about metrics like this:
1) That is essentially a server-side metric that the Sagemaker service should be responsible for tracking, since the length of time to delete of a resource should not be dependent on the client that calls the delete operation.
2) What you would be tracking here is essentially the *poll time*, not the actual delete time. For the actual delete time, only the server side knows that.

For things like how many times an update failed over time, I presume you are talking about how many times the update code path *in a specific test failed over time* (i.e. a regression test analysis)? If you're talking about how many times a particular Update operation failed on the server side, please see point 1) above.

For regression analysis I generally prefer unit tests which can be an effective way to test individual parts of a functional test's code paths by mocking out test dependencies and isolating just one part of the code base under test. Functional integration or end-to-end tests, however, are meant to verify the full behaviour of the system under test. What these tests are trying to do is validate specific use cases in their entirety. The use case being validated here is a "smoke test" of the Create Resource -> Read Resource -> Delete Resource use case. Other use cases I hope that e2e tests will validate are things like:

* Adopting a pre-existing resource
* Cross-account resource management
* Upgrading the version of a controller and verifying resources under management can still be managed

RedbackThomson: Hey @jaypipes,

So I've just taken this advice listed above for work with the new Application Auto Scaling integration tests I've been working on lately. I think the only thing missing from your proposed solution is a way of correctly handling cleanup of tests. The great thing about the fixtures is the ability to `yield` and delete regardless of the assertions within a test. 

As soon as you handle the CRUD lifecycle of the CR within the test body, if you run into a failed assertion prior to the delete instruction then the test will fail early and never clean up its own resources. This then becomes a problem when attempting to retry tests, as older cluster resources could collide with the new ones.

One such proposed solution would be to have a cleanup fixture, with which you can register all of your custom resource references in each test, that will ensure that the tests at least call `delete` before exiting. 

Do you think I'm missing some obvious solution that otherwise could solve this issue?

jaypipes: Sorry @RedbackThomson I seem to have missed your comment above, apologies!

> As soon as you handle the CRUD lifecycle of the CR within the test body, if you run into a failed assertion prior to the delete instruction then the test will fail early and never clean up its own resources. This then becomes a problem when attempting to retry tests, as older cluster resources could collide with the new ones.

I guess I don't see this as a problem, because if an assertion about a resource under test fails, typically we *want* the resource to stay around so we can diagnose *why* the assertion failed. Note that I am specifically referring to resources *under test* -- meaning, the specific resource we are asserting about. Fixtures (and the yield/teardown functionality) are for resources that are *dependencies* of tests, not resources that are themselves the subject of assertions.

jaypipes (approved):

test/e2e/sagemaker/tests/test_hpo.py
RedbackThomson: I would agree up until your CI system runs the tests 30 times a day, each of them failing because of a poorly written test, and then you start to hit resource quotas and limits in the AWS account. This is what we saw happening in our long-running canary tests for SageMaker and it required a separate script that cleaned up hanging resources. It isn't ideal, because we have to manage that script separately from our tests.

test/e2e/sagemaker/tests/test_hpo.py
surajkota: A job can go from InProgress -> Stopping -> Completed/Stopped depending on when it was stopped. So its fine to expect Completed if job is stopped.

If completed is part of created state then we don't know if its because it has deleted or job actually completed. Either we need a longer job here or we need to poll more frequently

RedbackThomson (approved):
Looks great

test/e2e/sagemaker/tests/test_hpo.py
jaypipes: > I would agree up until your CI system runs the tests 30 times a day, each of them failing because of a poorly written test, and then you start to hit resource quotas and limits in the AWS account. This is what we saw happening in our long-running canary tests for SageMaker and it required a separate script that cleaned up hanging resources. It isn't ideal, because we have to manage that script separately from our tests.

This is actually how upstream Kubernetes CI system works. There is a separate process called Boskos that acts as a janitor for the Prow CI jobs and ensures resources are cleaned up:

https://github.com/kubernetes-sigs/boskos

Eventually, when we move to a Prow-based system, we might consider moving our bootstrap/cleanup for some things into Boskos configurations.

surajkota (approved):

PR #714, opened 2021-03-11
Add option to skip python tests and execute bash tests instead

echen-98:
Description of changes: 

- Add option to skip python tests and execute bash tests instead: the default behavior remains to execute the python tests, but now that can be overridden if `SKIP_PYTHON_TESTS` is set `true`.
- Update usage message
- Remove invocation of python tests from `run-tests.sh` and clean up script

Regarding `run-tests.sh`: I removed the check for an `__init__.py` file since this check already exists in `kind-build-test.sh`; it was pretty confusing that even if `enable_python_tests` was set `false`, `run-tests.sh` would be executed and go on to execute python tests anyway (if they exist for the service). 

In other words, `enable_python_tests="true"` means "run the python tests inside Docker" – there's no change there. However, before, `enable_python_tests="false"` meant "run python tests normally if they exist; otherwise, run bash tests". With these changes, `enable_python_tests="false"` will actually mean "don't run python tests, but run bash tests instead". 

### DIFF
--- a/scripts/kind-build-test.sh
+++ b/scripts/kind-build-test.sh
@@ -19,6 +19,8 @@ AWS_ROLE_ARN=${AWS_ROLE_ARN:-""}
 ACK_ENABLE_DEVELOPMENT_LOGGING="true"
 ENABLE_PROMETHEUS=${ENABLE_PROMETHEUS:-"false"}
 TEST_HELM_CHARTS=${TEST_HELM_CHARTS:-"true"}
+SKIP_PYTHON_TESTS=${SKIP_PYTHON_TESTS:-"false"}
+RUN_PYTEST_LOCALLY=${RUN_PYTEST_LOCALLY:="false"}
 ACK_LOG_LEVEL="debug"
 ACK_RESOURCE_TAGS='services.k8s.aws/managed=true, services.k8s.aws/created=%UTCNOW%, services.k8s.aws/namespace=%KUBERNETES_NAMESPACE%'
 DELETE_CLUSTER_ARGS=""
@@ -93,6 +95,14 @@ Environment variables:
                             Default: $ROOT_DIR/build/tmp-$CLUSTER_NAME
   K8S_VERSION               Kubernetes Version [1.14, 1.15, 1.16, 1.17, and 1.18]
                             Default: 1.16
+  TEST_HELM_CHARTS          Whether to run the test-helm.sh script (<true|false>)
+                            Default: true
+  SKIP_PYTHON_TESTS         Whether to skip python tests and run bash tests instead for
+                            the service controller (<true|false>)
+                            Default: false
+  RUN_PYTEST_LOCALLY        If python tests exist, whether to run them locally instead of
+                            inside Docker (<true|false>)
+                            Default: false
 "
 
 if [ $# -ne 1 ]; then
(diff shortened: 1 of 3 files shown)


scripts/kind-build-test.sh
```diff
@@ -93,6 +94,11 @@ Environment variables:
                             Default: $ROOT_DIR/build/tmp-$CLUSTER_NAME
   K8S_VERSION               Kubernetes Version [1.14, 1.15, 1.16, 1.17, and 1.18]
                             Default: 1.16
+  TEST_HELM_CHARTS          Whether to run the test-helm.sh script (<true|false>)
```
jaypipes, line 97: :+1: thanks for adding this one.

test/e2e/run-tests.sh
```diff
@@ -21,41 +27,21 @@ if [ $# -ne 1 ]; then
...
-  PYTEST_LOG_LEVEL="${PYTEST_LOG_LEVEL:-"INFO"}"
-  python bootstrap.py "${SERVICE}"
-
-  set +e
-
-  PYTHONPATH=. pytest -n auto --dist loadfile --log-cli-level "${PYTEST_LOG_LEVEL}" "${SERVICE}"
-  python cleanup.py "${SERVICE}"
-
-  set -eo pipefail
-fi
```
jaypipes, line 61: hmmm. by removing the above, you are removing the ability of the contributor to easily run the python tests on their local system instead of inside the Docker python testing image (which is how $TEST_E2E_DIR/build-run-test-dockerfile.sh works).

I suppose I don't mind too much being forced to use the Docker approach to running python tests, but wondering what 
@RedbackThomson thoughts are on this?

RedbackThomson: Yeah perhaps I don't understand why this block is being removed? It seems to just be removing the ability to run these tests locally?
I personally run tests locally because waiting for Docker and debugging inside a container is a pain.

echen-98: Yeah I thought about this a bit more, I agree keeping this would be useful. 

Since the bash tests are still useful to us (before we migrate everything to python), I found it inconvenient that `run-tests.sh` automatically ran python tests if they existed, even though the script was invoked because `enable_python_tests` was false in `kind-build-test.sh`.

I suppose there are currently 3 ways to run these tests (bash, python locally, python inside Docker), I'll try to come up with a way to allow the user to clearly specify one way.

echen-98:
Ok hopefully this should be an improvement. The way to run tests is determined by `SKIP_PYTHON_TESTS` (default false, as before) and `RUN_PYTEST_LOCALLY` (default false but can change if needed), and the logic is abstracted to `run-tests.sh`.

scripts/kind-build-test.sh
```diff
@@ -238,21 +248,7 @@ if [[ "$TEST_HELM_CHARTS" == true ]]; then
...
-if [[ "$enable_python_tests" == "false" ]]; then
-    # Run tests directly
-    $TEST_E2E_DIR/run-tests.sh $AWS_SERVICE
-else
-    # Run tests inside Dockerfile
-    $TEST_E2E_DIR/build-run-test-dockerfile.sh $AWS_SERVICE
-fi
+# run e2e tests
+export SKIP_PYTHON_TESTS="$SKIP_PYTHON_TESTS"
+export RUN_PYTEST_LOCALLY="$RUN_PYTEST_LOCALLY"
```
RedbackThomson, line 253: nit: These can be simplified down to just `export SKIP_PYTHON_TESTS` and `export RUN_PYTEST_LOCALLY`?

RedbackThomson (approved):
This is great! Thank you @echen-98 !

scripts/kind-build-test.sh
echen-98: Done, thanks for the feedback!

jaypipes (approved):
Nice work, Eric. ++

PR #725, opened 2021-03-23
Elasticache: replication group large cluster test and support for skipping slow tests by default

echen-98:
Description of changes:

- large cluster load test
- add support for a `pytest.mark.slow` marker, and skip these tests by default (we don't want tests like this to run with the rest of the tests)

### DIFF
--- a/test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
+++ b/test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
@@ -1,0 +1,73 @@
+# Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may
+# not use this file except in compliance with the License. A copy of the
+# License is located at
+#
+#	 http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+"""Large cluster test for replication group resource
+"""
+
+import pytest
+
+from time import sleep
+from common import k8s
+from elasticache.tests.test_replicationgroup import make_replication_group, rg_deletion_waiter, make_rg_name, DEFAULT_WAIT_SECS
+from elasticache.util import provide_node_group_configuration
+
+@pytest.fixture(scope="module")
+def rg_largecluster_input(make_rg_name):
+    return {
+        "RG_ID": make_rg_name("rg-large-cluster"),
+        "NUM_NODE_GROUPS": "125",
+        "REPLICAS_PER_NODE_GROUP": "3"
+    }
+
+@pytest.fixture(scope="module")
+def rg_largecluster(rg_largecluster_input, make_replication_group, rg_deletion_waiter):
+    input_dict = rg_largecluster_input
+
+    (reference, resource) = make_replication_group("replicationgroup_largecluster", input_dict, input_dict["RG_ID"])
+    yield (reference, resource)
+
+    # teardown
+    k8s.delete_custom_resource(reference)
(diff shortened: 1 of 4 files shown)


test/e2e/elasticache/resources/replicationgroup_loadtest.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: elasticache.services.k8s.aws/v1alpha1
+kind: ReplicationGroup
+metadata:
+  name: $RG_ID
+spec:
+  cacheNodeType: cache.t3.micro
+  engine: redis
+  numNodeGroups: $NUM_NODE_GROUPS
+  replicasPerNodeGroup: $REPLICAS_PER_NODE_GROUP
+  replicationGroupDescription: large CME RG for load test
```
jaypipes, line 10: CME == Cluster Mode Enabled, yes?

test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
```diff
@@ -0,0 +1,75 @@
...
+# permissions and limitations under the License.
+
+"""Large cluster load test for replication group resource
+"""
+
+import pytest
+
+from time import sleep
+from common import k8s
+from elasticache.tests.test_replicationgroup import make_replication_group, rg_deletion_waiter, make_rg_name, DEFAULT_WAIT_SECS
```
jaypipes, line 21: you import DEFAULT_WAIT_SECS from test_replicationgroup but override this variable name below on line 24 :)

test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
```diff
@@ -0,0 +1,75 @@
...
+                          "nodeGroupConfiguration": provide_node_group_configuration(desired_node_groups)}}
+        _ = k8s.patch_custom_resource(reference, patch)
+        sleep(DEFAULT_WAIT_SECS) # required as controller has likely not placed the resource in modifying
+        assert k8s.wait_on_condition(reference, "ACK.ResourceSynced", "True", wait_periods=90)
+
+        # assert new state after scaling in
+        resource = k8s.get_resource(reference)
+        assert resource['status']['status'] == "available"
+        assert len(resource['status']['nodeGroups']) == desired_node_groups
+        assert len(resource['status']['memberClusters']) == desired_total_nodes
```
jaypipes, line 75: I thought from the name of this test that it would be testing the load on the ACK controller, but that isn't actually the case here; instead this is simply testing that the ACK controller waits for a long-ish period of time for a large ElastiCache cluster to get into a available state. This is fine, but it's not a load test in any sense of the word. Perhaps consider changing any language in the PR commit message and files like `test/e2e/elasticache/resources/replicationgroup_loadtest.yaml` to remove the word "load".

test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
echen-98: will delete the override, I was playing around with different times and forgot I had imported it after I settled on 30

test/e2e/elasticache/resources/replicationgroup_loadtest.yaml
echen-98: correct, I can expand it

test/e2e/elasticache/tests/test_replicationgroup_largecluster.py
echen-98: I suppose that's true, could you clarify what a load test might look like? Something like making a large volume of requests to the controller and measuring some metric?

I think this test was just to check that the controller doesn't have any issues managing the largest cluster an Elasticache customer can create (with respect to pagination or any other issues that might arise). The long wait times were more so a side effect of creating and modifying a large cluster rather than the main purpose of the test, but I agree that it doesn't necessarily "load" the controller much more than any other regular test.

echen-98:
Ok, removed any "load test" wording (including in the commit message)

jaypipes (approved):
Nice work @echen-98 :)

Issue #735, opened 2021-04-08
Service team ownership of e2e test resources


RedbackThomson:
**Is your feature request related to a problem?**
Currently all of the e2e tests are in the https://github.com/aws-controllers-k8s/community repository. It requires the approval of any of the ACK core contributors to merge and maintain this code over time.

**Describe the solution you'd like**
- Service controller repositories house any service-specific testing code
- https://github.com/aws-controllers-k8s/test-infra contains any common test code and invocation scripts
- Service teams are responsible for maintenance and merging of their tests

**Describe alternatives you've considered**
N/A

**Progress**

- [x] Common scripts moved into `test-infra` - https://github.com/aws-controllers-k8s/test-infra/pull/6
- [x] https://github.com/aws-controllers-k8s/community make targets and scripts updated to reference new repository structures - https://github.com/aws-controllers-k8s/community/pull/736
- [x] Service teams refactored and moved their tests (minor changes) 
  - [x] apigatewayv2 - https://github.com/aws-controllers-k8s/apigatewayv2-controller/pull/1
  - [x] applicationautoscaling - https://github.com/aws-controllers-k8s/applicationautoscaling-controller/pull/2
  - [x] elasticache - https://github.com/aws-controllers-k8s/elasticache-controller/pull/19
  - [x] mq - https://github.com/aws-controllers-k8s/mq-controller/pull/3
  - [x] rds - https://github.com/aws-controllers-k8s/rds-controller/pull/4
  - [x] sagemaker - https://github.com/aws-controllers-k8s/sagemaker-controller/pull/13
  - [x] s3 - https://github.com/aws-controllers-k8s/s3-controller/pull/11

CC'ing:
@aws-controllers-k8s/elasticache-maintainer 
@aws-controllers-k8s/rds-maintainer 
@aws-controllers-k8s/applicationautoscaling-maintainer 
@aws-controllers-k8s/sagemaker-maintainer
@aws-controllers-k8s/apigatewayv2-maintainer 
@aws-controllers-k8s/mq-maintainer

RedbackThomson:
All current Python e2e tests have been moved into their respective repositories. The only e2e tests left in `community` are the bash ones. These are still unofficially supported, but support for them may be broken without warning. They will probably be deleted from this repo fairly soon, anyway.

PR #736, opened 2021-04-08
Point E2E test scripts to `test-infra`

RedbackThomson:
Issue #, if available: #735

Description of changes:
- Points the e2e Make target to the `test-infra` repository scripts
- Creates an alternate KIND cluster config without Prometheus support
  - You cannot run two tests on the same machine if both bind to the `:9090` host port defined in the prometheus worker node group
- Removed all e2e test folders for services that use Python
- nit: Removed `--quiet` from `kind create-cluster` for the sake of debugging in automation


scripts/kind-two-node-prometheus-cluster.yaml
```diff
@@ -0,0 +1,16 @@
...
+        apiVersion: kubeadm.k8s.io/v1beta2
+        kind: ClusterConfiguration
+        metadata:
+          name: config
+  - role: worker
+    extraPortMappings: # Host machine to Prometheus service reachability
+      - containerPort: 30900
+        hostPort: 9090
+        listenAddress: "0.0.0.0" # Optional, defaults to "0.0.0.0"
+        protocol: tcp # Optional, defaults to tcp
```
jaypipes, line 16: where did this file come from?

RedbackThomson: We were enabling prometheus port-forwarding, by default on all KIND clusters. This caused port collision issues when I tried to run multiple e2e tests on the same machine. I don't see this being an issue for our CI, but it could be an issue for local development. I had to create a separate KIND config, one with Prometheus port forwarding, and remove the port forwarding from the default one. Then I use the Prometheus specific one only when the `ENABLE_PROMETHEUS` flag is set `true`.

jaypipes (approved):
k, lgtm. going to squash.

scripts/provision-kind-cluster.sh
```diff
@@ -66,7 +74,7 @@ debug_msg "kind: using Kubernetes $K8_VERSION"
 echo -n "creating kind cluster $cluster_name ... "
 for i in $(seq 0 5); do
   if [[ -z $(kind get clusters 2>/dev/null | grep $cluster_name) ]]; then
-      kind create cluster -q --name "$cluster_name" --image $K8_VERSION --config "$KIND_CONFIG_FILE" --kubeconfig $TMP_DIR/kubeconfig 1>&2 || :
```
jaypipes, line 69: @RedbackThomson was the above change on purpose or did you just forget to re-enable quiet-mode after debugging?

RedbackThomson: I was initially going to put it back to quiet, but I have had lots of silent errors with it that I feel for the sake of debugging it would be better to leave it verbose. For instance, if it fails (silently) then "loading images into cluster" exits with code 2. That red herring led me on a trail of Google-fu until I finally checked that the cluster was actually standing.

PR #756, opened 2021-05-03
Explicitly specify a generator config path for ack-generate olm

komish:
Signed-off-by: Jose R. Gonzalez <josegonzalez89@gmail.com>

**Issue #, if available:**

Unblocks me on the following, but this is not the resolution of that issue.
https://github.com/aws-controllers-k8s/community/issues/744 (specifically: https://github.com/aws-controllers-k8s/community/issues/744#issuecomment-830428118)

**Description of changes:**

This change will update the OLM component of scripts/build-controller-release.sh to explicitly set the path to the generator.yaml. The default value of this option (if unspecified) is to be unset (https://github.com/aws-controllers-k8s/code-generator/blob/main/cmd/ack-generate/command/root.go#L115-L117) which causes the OLM bundle generation logic to incorrectly scaffold things for ignored resources.

👍


jaypipes (approved):
:+1:

Issue #744, opened 2021-04-26
Add OLM-specific generator configs for each service and the service template


komish:
**Is your feature request related to a problem?**
No

**Describe the solution you'd like**
To simplify the automated creation of Operator Lifecycle Manager ("OLM") bundles for a given controller release, the controller repos need a base configuration added.

As an initial goal, I figure it makes sense to contribute one for each repo for which an e2e test exists (referring to: https://github.com/aws-controllers-k8s/community/tree/main/test/e2e). That said, I see S3 is missing here and I've done most testing against S3 so I'll throw that in as well:

* dynamodb https://github.com/aws-controllers-k8s/dynamodb-controller
* ecr https://github.com/aws-controllers-k8s/ecr-controller
* kms https://github.com/aws-controllers-k8s/kms-controller
* mq https://github.com/aws-controllers-k8s/mq-controller
* sfn https://github.com/aws-controllers-k8s/sfn-controller
* sns https://github.com/aws-controllers-k8s/sns-controller
* sqs https://github.com/aws-controllers-k8s/sqs-controller
* sqs https://github.com/aws-controllers-k8s/s3-controller

We can also add a generic config for the template

* template https://github.com/aws-controllers-k8s/template-controller

I expect these configs to live at this path: `{SERVICE}-controller/olm/olmconfig.yaml` to align with the default location being used by build scripts.

Alternatively, what services would you prefer to see? I can add those as well. 

**Describe alternatives you've considered**
The alternative is to have these configs hosted elsewhere as a relatively-static asset that can be pulled down at build-time. This alternate would require an additional step either in build automation or in the scripts that call `ack-generate olm`.

RedbackThomson:
Actually each of the repositories you've listed in this are using the old-style bash e2e resources. All of the other repositories have transferred to the new Python e2e resources, and I'd say those are more suitable for this enhancement. A full list of those repositories is in this issue: https://github.com/aws-controllers-k8s/community/issues/735

komish:
Ah, thanks @RedbackThomson! I'll check out the list and target that instead.

komish:
With that in mind, here are the new targets:

- [ ] apigatewayv2 - [aws-controllers-k8s/apigatewayv2-controller](https://github.com/aws-controllers-k8s/apigatewayv2-controller)
- [ ] applicationautoscaling - [aws-controllers-k8s/applicationautoscaling-controller](https://github.com/aws-controllers-k8s/applicationautoscaling-controller)
[...]

PR #760, opened 2021-05-05
Add an environment variable to set aws-sdk-go version in `build-controller-release.sh`

A-Hilaly:
Part of #744 [discussion](https://github.com/aws-controllers-k8s/community/issues/744#issuecomment-832260936)

Description of changes:
Currently the aws-sdk-go version is hard coded in the `scripts/build-controller-release.sh`. This commit adds an envirnoment variable `AWS_SDK_GO_VERSION` to be able to set SDK version used to generate the release files.

### DIFF
--- a/scripts/build-controller-release.sh
+++ b/scripts/build-controller-release.sh
@@ -36,6 +36,7 @@ ACK_GENERATE_BIN_PATH=${ACK_GENERATE_BIN_PATH:-$DEFAULT_ACK_GENERATE_BIN_PATH}
 ACK_GENERATE_API_VERSION=${ACK_GENERATE_API_VERSION:-"v1alpha1"}
 ACK_GENERATE_CONFIG_PATH=${ACK_GENERATE_CONFIG_PATH:-""}
 ACK_GENERATE_IMAGE_REPOSITORY=${ACK_GENERATE_IMAGE_REPOSITORY:-"$DEFAULT_IMAGE_REPOSITORY"}
+AWS_SDK_GO_VERSION=${AWS_SDK_GO_VERSION:-"v1.35.5"}
 
 DEFAULT_TEMPLATES_DIR="$ROOT_DIR/../../aws-controllers-k8s/code-generator/templates"
 TEMPLATES_DIR=${TEMPLATES_DIR:-$DEFAULT_TEMPLATES_DIR}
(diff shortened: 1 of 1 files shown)


scripts/build-controller-release.sh
```diff
@@ -36,6 +36,7 @@ ACK_GENERATE_BIN_PATH=${ACK_GENERATE_BIN_PATH:-$DEFAULT_ACK_GENERATE_BIN_PATH}
 ACK_GENERATE_API_VERSION=${ACK_GENERATE_API_VERSION:-"v1alpha1"}
 ACK_GENERATE_CONFIG_PATH=${ACK_GENERATE_CONFIG_PATH:-""}
 ACK_GENERATE_IMAGE_REPOSITORY=${ACK_GENERATE_IMAGE_REPOSITORY:-"$DEFAULT_IMAGE_REPOSITORY"}
+AWS_SDK_GO_VERSION=${AWS_SDK_GO_VERSION:-"v1.35.5"}
```
RedbackThomson, line 39: Latest is `1.38.33`. Should we move the needle or would this introduce bugs?

RedbackThomson (changes requested):
Also need to update `ag_olm_args`, which has it hardcoded as well.

scripts/build-controller-release.sh
A-Hilaly: Good point. +1 for using the latest - worst case we can set the older version..

A-Hilaly: I'll leave the version as `v1.35.5` for now, it's better to do it separately since other files also use 1.35.5 as default version

RedbackThomson: Do we know how these are coupled? Is it a good idea to make it into an ENV variable if we aren't comfortable changing it?

A-Hilaly: The only scripts using `AWS_SDK_GO_VERSION` variable are `build-controller.sh`  and `build-controller-release.sh`, and AFAIK they not really coupled, both scripts call `ack-generate` for different purposes. The only scenario that i'm concerned about (if the defaults values are not synced) is if a user calls both scripts without setting `AWS_SDK_GO_VERSION`. 

> Is it a good idea to make it into an ENV variable if we aren't comfortable changing it?

IMO It's not super dangerous, it's just safer to keep them in sync. Is there a different approach if we don't make it into an ENV variable?

RedbackThomson: > both scripts call `ack-generate` for different purposes

I suppose this is the confusing part to me. Is this something we should address with a clearer configuration? 

Nevertheless I'm happy to merge this PR for now to unblock @komish . But I think I'll bring this up in our next meeting :)

RedbackThomson (approved):

scripts/build-controller-release.sh
komish: @RedbackThomson For what it's worth - at this point it isn't a hard block because we're not actively publishing/releasing OpenShift/OLM related components. I'll be able to launch my PRs soon without this being merged with the understanding that when the Go SDK version discrepancy gets sorted, things will fall into place. 👍

A-Hilaly: > Is this something we should address with a clearer configuration?

Definitely! I think this is an area we still can improve ++

PR #801, opened 2021-05-26
increase role session duration

surajkota:
Description of changes:
Test parallelism in prow CI is low and it runs only 2-3 tests. sagemaker-controller tests do not finish in one hour as a result. see - https://github.com/aws-controllers-k8s/sagemaker-controller/pull/34. the tests won't be able to clean up the resources as well since the credentials expire which is a big issue

Can we increase the number of threads in addition to this change? CI experience is bad if the developer needs to wait for 2 hours to know their test results

Another solution is to make the session duration and num threads configurable parameter per controller repo if we configure a file that has these env variables in each controller repo

### DIFF
--- a/scripts/lib/aws.sh
+++ b/scripts/lib/aws.sh
@@ -50,7 +50,7 @@ aws_generate_temp_creds() {
     JSON=$(daws sts assume-role \
            --role-arn "$ACK_ROLE_ARN"  \
            --role-session-name tmp-role-"$__uuid" \
-           --duration-seconds 3600 \
+           --duration-seconds 7200 \
            --output json || exit 1)
 
     AWS_ACCESS_KEY_ID=$(echo "${JSON}" | jq --raw-output ".Credentials[\"AccessKeyId\"]")


RedbackThomson:
Currently pytest detects number of usable cores (virtualised or otherwise) and runs one threaded test per core. I think that this should be the default behaviour, but from inside a container (and that container inside a DIND Prow job) it doesn't make much sense. We'll have to play around with CPU and memory limits on the pods to be sure we don't cap out when multithreading, but I can look into it.

scripts/lib/aws.sh
```diff
@@ -50,7 +50,7 @@ aws_generate_temp_creds() {
     JSON=$(daws sts assume-role \
            --role-arn "$ACK_ROLE_ARN"  \
            --role-session-name tmp-role-"$__uuid" \
-           --duration-seconds 3600 \
+           --duration-seconds 7200 \
```
RedbackThomson, line 53: I'd like to try this out in a Prow environment before merging. I worry that IAM role chaining, this might be denied by the STS service.

jaypipes:
@surajkota some questions for you:

1) How long does the slowest SageMaker e2e test take?
2) What kind of parallelism are you looking for? pytest automatically spawns a number of workers related to the Pod's number of CPUs, so we could increase the size of the ProwJob CPU requests for some of the service controller e2e jobs.
3) When you run the SageMaker e2e tests "locally" (against a pre-setup EKS cluster with your controller installed already), how long do the e2e tests take to run?
4) We could look into [a fail-fast option](https://docs.pytest.org/en/6.2.x/usage.html#stopping-after-the-first-or-n-failures) so that a test failure can prematurely kill other running tests.

scripts/lib/aws.sh
surajkota: Cool, let us know the outcome

surajkota:
> How long does the slowest SageMaker e2e test take?

about 22-25 mins

> What kind of parallelism are you looking for? pytest automatically spawns a number of workers related to the Pod's number of CPUs, so we could increase the size of the ProwJob CPU requests for some of the service controller e2e jobs.

Yes, I am looking for more cpus so pytest xdist can spawn more threads. I used the incorrect term in the issue, I meant to say `Can we increase the number of cpus/vcpus`

> When you run the SageMaker e2e tests "locally" (against a pre-setup EKS cluster with your controller installed already), how long do the e2e tests take to run?

Codebuild size - 15 GB memory, 8 vCPUs
Pytest num threads - 10,
~30 minutes only for the pytest command excluding bootstrap, cleanup, building container
~40 minutes end to end

> We could look into a fail-fast option so that a test failure can prematurely kill other running tests.

Even if we add this configuration, resources will leak from tests and service_cleanup script if credentials expire

scripts/lib/aws.sh
RedbackThomson: As I suspected, this violates the 1 hour IAM chaining rule. 
```
An error occurred (ValidationError) when calling the AssumeRole operation: The requested DurationSeconds exceeds the 1 hour session limit for roles assumed by role chaining.
```

I think that an hour is more than enough time to achieve an e2e test result, though. So instead we should look toward parallelising the tests.

Issue #763, opened 2021-05-06
Add support for optionally building service controller images that meet Red Hat's Container Certification Requirements


komish:
**Is your feature request related to a problem?**
No.

**Describe the solution you'd like**

The ability to build the service controller images in a way that meets Red Hat's Container Certification requirements. This amounts to

* A base image built on Red Hat UBI ("minimal" should suffice in this case)
* Additional labels added 

For starters, I would target the latest UBI minimal image ([8.3](https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?gti-tabs=unauthenticated)) until a Micro image is made available. When a Micro image is made available, it may be worth considering a shift to that if everything works out of the box.

This image will need to be bumped periodically as the minor release increments. Also note that I'll need to perform a test to make sure the controller functions with the UBI image.

**Potential challenges to address**

1) In the build-controller-image.sh script, you cannot change the target docker registry without replacing the entire repository AND tag. If possible, add the ability to change the registry and repository independent of the tag logic. [publish-controller-image.sh](https://github.com/aws-controllers-k8s/community/blob/main/scripts/publish-controller-image.sh#L29-L34) and [olm-publish-bundle-image.sh](https://github.com/aws-controllers-k8s/community/blob/main/scripts/olm-build-bundle-image.sh#L67-L70) both decouple the image tag from the registry and introduce a precedence based on specificity to determine what the resulting value would be.

2) The final controller image that would be submitted to Red Hat certification pipelines has a few additional pieces of metadata (read: labels) that need to access on the built container image. Those amount to: Reference: [Red Hat Certified Operator Guide - Docker](https://redhat-connect.gitbook.io/certified-operator-guide/ansible-operators/building-an-ansible-operator/dockerfile-requirements-ansible)

   ```Dockerfile
   LABEL name="ACK - Amazon SageMaker" \
      vendor="Amazon, Inc." \
      version="v0.0.1" \
      release="1" \
      summary="AWS Controllers for Kubernetes - Amazon SageMaker Controller" \
      description="This operator will manage Amazon SageMaker services from within your cluster."
    ```

3) The final controller image needs to have its license accessible at `/licenses`. This is probably just an extra `COPY` statement in the docker file. https://github.com/aws-controllers-k8s/community/blob/main/Dockerfile#L47

[...]

PR #820, opened 2021-06-07
Parameterize controller build image

komish:
**Issue #, if available:**
Related to #763, but does not close this issue.

**Description of changes:**

This change adds a new `base_image` argument to all Dockerfiles, and sets the default value of this parameter to the original value of `amazonlinux:2`. This should cause no change for existing build workflows.

According to documentation, this new argument must exist prior to the very first `FROM` in order to be used in a `FROM` action call. https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

This enables developers to build a container from this Dockerfile with an alternate target base image for the final stage.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. 👍 

Signed-off-by: Jose R. Gonzalez <josegonzalez89@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# Base image to use for the final stage
+ARG base_image=amazonlinux:2
 # Build the manager binary
 FROM golang:1.14.1 as builder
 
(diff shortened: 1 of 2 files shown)


A-Hilaly (approved):
👍

Issue #830, opened 2021-06-10
API Reference Documentation


RedbackThomson:
**Is your feature request related to a problem?**
There is currently no way to see the API spec in a easily-readable way without looking directly at the source code.

**Describe the solution you'd like**
An auto-generated documentation website (connected to the `community` docs) that serves each service and each version's spec and status fields

**Describe alternatives you've considered**
N/A

RedbackThomson:
Documentation is now generated for all services upon pushing a new tag and is regenerated automatically when pushing changes to the `docs/` folder in community

PR #834, opened 2021-06-15
Create API reference documentation generator

RedbackThomson:
Issue #, if available: #830 

Description of changes:
Create a Python script to traverse the service controller repositories for all CRD bases. From those:
1. Generate a YAML header in markdown for each resource
2. Generate an Overview page of all resources and their versions


docs/README.md
```diff
@@ -59,8 +59,16 @@ The [MkDocs reference](https://www.mkdocs.org/user-guide/writing-your-docs/) inc
...
+The `scripts/gen_reference.py` file will attempt to navigate every service controller repository in the `$GOPATH` and build a full model of every custom resource and version based off the CRD base files. It will then output these models into YAML headers placed into markdown files under `contents/reference/<service>/<version>/<resource>.md`, which are then rendered by `mkdocs-macros` snippets in `contents/snippets/`.
+
+To run `scripts/gen_reference.py` navigate to `community/docs` and then run `python3 ./scripts/gen_reference.py`.
+
 ### Publish
 
 To publish updated docs, commit changes to the markdown files and open a pull request. When your commits are merged, a github action will automatically build and deploy the site.
 
-[View the status of github action runs.](https://github.com/aws/aws-controllers-k8s/actions)
+[View the status of github action runs.](https://github.com/aws-controllers-k8s/community/actions)
```
jaypipes, line 74: thanks :)

docs/scripts/gen_reference.py
```diff
@@ -0,0 +1,217 @@
...
+    group: str
+    apiVersion: str
+    description: str
+    scope: str
+    names: ResourceNames
+    spec: List[Field]
+    status: List[Field]
+
+    def output_path(self) -> Path:
+        return Path(f"{self.service}/{self.apiVersion}/{self.name}.md")
```
jaypipes, line 43: might consider lowercasing the `{self.name}`...

docs/scripts/gen_reference.py
```diff
@@ -0,0 +1,217 @@
...
+            controller_repo.stem.startswith("template"):
+            continue
+
+        service_bases_path = controller_repo / bases_path
+
+        if not service_bases_path.exists():
+            raise ValueError(f"Service base path {service_bases_path} does not exist")
+
+        # Get service name from repository
+        # TODO(RedbackThomson): Find an elegant way to get name-cased
```
jaypipes, line 190: Why? I actually prefer the lowercased alias because that is what will appear in both the name of the controller image/chart tags as well as the CRD's APIGroup...

docs/README.md
```diff
@@ -59,8 +59,16 @@ The [MkDocs reference](https://www.mkdocs.org/user-guide/writing-your-docs/) inc
 
 Review the [Material for MkDocs reference](https://squidfunk.github.io/mkdocs-material/reference/formatting/) for information on the theme and formatting.
 
+### Generating the reference
+
+Our documentation contains an API reference section that is autogenerated from the custom resource definitions of each ACK service. These reference documents need to be re-generated each time there is a change to any of the fields or versions of any of these CRDs.
+
+The `scripts/gen_reference.py` file will attempt to navigate every service controller repository in the `$GOPATH` and build a full model of every custom resource and version based off the CRD base files. It will then output these models into YAML headers placed into markdown files under `contents/reference/<service>/<version>/<resource>.md`, which are then rendered by `mkdocs-macros` snippets in `contents/snippets/`.
```
jaypipes, line 66: if we are going to integrate this into our postsubmit Prow jobs, is the ProwJob's working tree going to contain all the controller repositories?

docs/scripts/gen_reference.py
RedbackThomson: The `self.name` is the only part of the CRD that is properly cased - and that means it shows up properly cased in the sidebar. Usually I'd agree with the filename being lowercased, but in this case mkdocs is using it for formatting so I think it's preferred.

docs/README.md
RedbackThomson: That's the plan. Attach every service controller to every post submit prow job. It's easy enough to do this programmatically.

docs/scripts/gen_reference.py
```diff
@@ -0,0 +1,217 @@
...
+
+    status = schema['properties']['status']
+    status_properties = status['properties']
+    for field_name, field in status_properties.items():
+        required  = field_name in status['required'] if 'required' in status else False
+        res_status.append(convert_field(field_name, field, required))
+
+    return Resource(
+        name=crd['spec']['names']['kind'],
+        service=service, #TODO(RedbackThomson): Load from file path?
```
RedbackThomson, line 130: nit: Already done this @RedbackThomson

docs/scripts/gen_reference.py
RedbackThomson: I think we'll need both. One for prettiness and one for usability. But it's always easier to go from proper casing to lowercasing?

docs/README.md
jaypipes: ack

docs/scripts/gen_reference.py
jaypipes: k, makes sense

docs/scripts/gen_reference.py
jaypipes: well, either way it's not urgent :)

jaypipes (approved):

Issue #812, opened 2021-06-03
ability to deal with server side defaults/initializers


surajkota:
**Is your feature request related to a problem?**

Ability to specify fields initialized by service post create. Following are the patterns observed:
  1. Generated fields. Example:
     1. roleARN field in ScalableTarget resource is generated by service. See https://github.com/aws-controllers-k8s/applicationautoscaling-controller/pull/12#discussion_r640097037 (service - Applicationautoscaling)
     1. staticHyperparameter field in Hyperparameter tuning resource. Slice has some extra fields returned from the service which should be ignored for comparison. See [customSetDefaults in HPJob](https://github.com/aws-controllers-k8s/sagemaker-controller/blob/main/pkg/resource/hyper_parameter_tuning_job/custom_delta.go#L28-L32) (service - SageMaker)
  2. Default value for fields
      1. Boolean fields returned as False.
         - Scalable target struct in ScalableTarget resource - `SuspendedState.<*>` (all fields) returned as False by default. See [setDefaults in autoscaling](https://github.com/aws-controllers-k8s/applicationautoscaling-controller/blob/main/pkg/resource/scalable_target/custom_delta.go#L25-L39) (service - Applicationautoscaling)
      1. String fields return default values. e.g. Model resource - `PrimaryContainer.Mode` retuned as  `SingleMode` by default (service - SageMaker)

**Describe the solution you'd like**
1.a, 2.a, 2.b - if DesiredResource.Spec.field = nil set these values from latestResource
1.b -  a hook to write custom algorithm for comparison and remove service generated fields, ignore setting values from latestResource to desired

**Note**: In `2.a`, 
- whole desired.spec.SuspendedState == nil and service still returns false values for all fields in the structs (`DynamicScalingInSuspended, DynamicScalingInSuspended, ScheduledScalingSuspended`)
-  `desired.spec.SuspendedState` != nil but `desired.spec.SuspendedState.DynamicScalingInSuspended` == nil and service still returns False for `desired.spec.SuspendedState.DynamicScalingInSuspended`

**Describe alternatives you've considered**
1.b - a hook to write custom logic to insert only server generated <key, value> pairs into the slice only if diff is False
1.b - a hook to write custom algorithm for comparison and insert service generated fields, set values from latestResource into desired only for comparison

echen-98:
>  1.b - a hook to write custom algorithm for comparison and remove service generated fields, ignore setting values from latestResource to desired

[...]

PR #849, opened 2021-06-28
Proposal for handling late-initialization in ACK.

vijtrip2:
Issue #812 , if available:

Description of changes:
* Adding the proposal on how to handle server-side defaults/late-initialization in ACK.
* You can find the rendered proposal document [here](https://github.com/vijtrip2/aws-controllers-k8s/blob/late-initializers/docs/design/proposals/late-initialization/late-initialization.md)

### DIFF
--- a/docs/design/proposals/late-initialization/late-initialization.md
+++ b/docs/design/proposals/late-initialization/late-initialization.md
@@ -1,0 +1,532 @@
+## Problem Overview
+
+[K8s convention](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#late-initialization) describes **Late Initialization** as “When resource fields are set by a system controller after an object is created/updated”
+
+In the context of an ACK service controller, when an AWS resource is created or updated by the controller, the AWS service may set optional fields for a resource to a default value and return them as part of Create/Update output or even as Get output in some cases. 
+
+This **late initialization** of optional fields in the AWS resource are not currently reflected in the desired state of the resource. This causes a difference to be detected between the stored desired state of the resource and the latest state fetched from the AWS service at the beginning of the reconciliation loop. However, this difference is erroneous since the controller does not need to call the resource manager's Update method in order for the actual state to match the desired state of the resource.
+
+This document will propose the solution for above problem and provide the details of implementing it.
+
+## Solution Requirements
+
+A solution for this problem should :
+
+1. Update the desired state of the K8s custom resource with late initialized fields which are optional and not provided by the K8s user. 
+
+2. Provide configuration for AWS service teams to indicate which operation output returns the late initialized fields. i.e. Get/List/Create/Update output.
+
+    > AWS APIs have different behavior for returning the late initialized fields to the user. Some services return these fields as part of Create output, while some services will return the late initialized fields as part of  ReadOne output after the resource creation.
+
+3. Wait for asynchronous resource creation to populate late initialized fields.
+
+    > For some services/resources the late initialized fields do not appear immediately.  This happens when there is asynchronous processing involved in resource creation. The proposed solution should allow waiting and retry until the late initialized fields appear in the mentioned API output.
+
+4. Allow the AWS service teams to provide custom code using hooks for handing of late initialized fields.
+
+## Proposed Solution
+In this solution, There will be updates in 
+1) **ACK Code Generator** to capture the late initialized fields from AWS operation output and use them to late initialize k8s custom resource object. Code generator changes will support a) configuration for service teams to tell ACK code generator which fields will be late initialized and b) configuration to provide custom code for late initialization.
+2) **ACK Runtime** to a) update the custom resource stored in etcd with late initalized optional spec fields And b) handle the late initalized fields for resources that have asynchronous creation.
+    > For the resources with Async creation, the late initalized fields are not readily available after create operation. Therefore the reconciler loop needs to be repeated for these resources until the late initalized fields are available.
+
+### ACK Code Generator Changes
+1. Fix the latent bug where Spec fields are not set from the Create and Update response. Once this bug is fixed the late initialized fields which are present in Create and Update response will be set.
+    ```go
+    // Current Code
+    // sdk.go.tpl    
+    func (rm *resourceManager) sdkCreate(
+	ctx context.Context,
+	desired *resource,
+    ) (created *resource, err error) {
+        ...
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}
+        ...
+    }
+    ```
+
+    ```go
+    // Proposed Code
+    // sdk.go.tpl
+    func (rm *resourceManager) sdkCreate(
+	ctx context.Context,
+	desired *resource,
+    ) (created *resource, err error) {
+        ...
+        // NOTE: The performSpecUpdate boolean param will be removed
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 }}
+        ...
+    }
+    ```
+
+2. **Provide a new "LateInitializationConfig" struct inside generator config for AWS service teams to provide which spec fields will be late initialized.** These fields will be the ones which are late initialized by the server but not present in the Create and Update response.
+
+    These fields will be populated by making the readOne call. AWS service teams can also provide a delay after which these fields will be late initialized.
+    LateInitializationConfig will be added as part of FieldConfig. 
+
+    ```go
+        // LateInitializeConfig contains instructions for how to handle the
+        // retrieval and setting of server-side defaulted fields.
+        type LateInitializationConfig struct {
+            // DelaySeconds provides the number of seconds to wait to set
+            // late initialized fields after successful Create/Update call
+            DelaySeconds int `json:"delay_seconds,omitempty"`
+        }
+
+        type FieldConfig struct {
+            LateInitialize *LateInitializeConfig `json:"late_initialize,omitempty"`
+        }
+    ```
+
+3. **Provide 2 new hooks to AWS service teams to customize the code generation in step 4**
+    
+    These 2 hooks will be:
+
+    a) **late_initialize_pre_read_one**:- This hook will provide the custom code to be inserted **before** setting server side defaults for missing optional spec fields.
+
+    b) **late_initialize_post_read_one**: This hook will provide the custom code to be inserted **after** setting server side defaults for missing optional spec fields.
+
+4. Add `AWSResourceManager.LateInitialize()` method to handle server side defaults for fields mentioned in LateInitializationConfig.
+
+* Proposed Pseudo Code
+    ```go
+    // manager.go.tpl
+    func (rm *resourceManager) LateInitialize(
+	ctx context.Context,
+    latest acktypes.AWSResource,
+    ) (acktypes.AWSResource, error) {
+        {{- if $hookCode := Hook .CRD "late_initialize_pre_read_one" }}
+            {{ $hookCode }}
+        {{- end }}
+
+        // 1. Filter all the fields which have LateInitializationConfig.
+        // 2. Filter out the fields from step1 with nil value in latest object.
+        // 3. If number of fields after step2 are nonzero, calculate the delay
+        // for late initialization.
+        // 4. If the latest object has 'LateInitialized=False' condition,
+        // no need to delay the late initialization. Otherwise calculate
+        // net delay for late initialization.(why? see step 7).
+        // 5. Net delay is maximum delay among all the Fields from step2.
+        // 6. If net delay is more than zero, add 'LateInitialized=False'
+        // condition to the resource status indicating that resource is still
+        // being late initalized and return with requeue delay as net delay.
+        // 7. Make the ReadOne call.
+        // 8. Filter out the fields from step2, which are not-nil in ReadOne output.
+        // 9. For all the fields in step 9, set the lateInitialized value.
+        // 10. Upsert resource condition 'LateInitialized=True'.
+
+        {{- if $hookCode := Hook .CRD "late_initialize_post_read_one" }}
+            {{ $hookCode }}
+        {{- end }}
+        // return latest with late initialized fields updates 
+    }
+    ```
+
+### ACK Runtime Changes
+
+* Current reconciler source code
+    ```go
+    func (r *reconciler) Sync() {
+        var latest acktypes.AWSResource
+        latest, err := rm.ReadOne()
+
+        if err!= nil {
+            if err != NotFound {
+                patchResourceStatus()
+                return err
+            }
+            latest,err := rm.Create()
+            if err != nil {
+                patchResourceStatus()
+                return err
+            }
+        } else {
+            delta := r.rd.Delta(desired, latest)
+            If delta.DifferentAt("Spec") {
+                latest,err := rm.Update()
+                if err != nil {
+                    patchResourceStatus()
+                    return err
+                }
+            }
+        }
+
+        r.patchResource(ctx, desired, latest)
+
+        // iterate through latest.conditions for requeue
+
+        return nil
+    }
+    ```
+
+* Proposed reconciler source code
+
+    ```go
+    func (r *reconciler) Sync() {
+        observed, err := rm.ReadOne()
+        var latest acktypes.AWSResource
+
+        if err!= nil {
+            if err != NotFound {
+                patchResourceStatus()
+                return err
+            }
+            // rm.Create() also updates the server side defaults from CreateResponse
+            latest,err := rm.Create()
+            if err != nil {
+                patchResourceStatus()
+                return err
+            }
+        } else {
+            delta := r.rd.Delta(desired, observed)
+            If delta.DifferentAt("Spec") {
+                // rm.Update() also updates the server side defaults from UpdateResponse
+                latest,err := rm.Update()
+                if err != nil {
+                    patchResourceStatus()
+                    return err
+                }
+            }
+        }
+
+        // updates the latest object with server side defaults.
+        // This operation supports hooks for service team to use.
+        // In case of no late initialization, lateInitializedLatest and 
+        // latest will be same.
+        lateInitializedLatest, lateInitializeRequeueErr := rm.LateInitialize(ctx, latest)
+
+        // Always persist the object state before exiting.
+        // patchResource will only Patch, if there is a diff.
+        if lateInitializedLatest != nil {
+            r.patchResource(ctx, desired, lateInitializedLatest) // patches metadata, spec and status if needed
+            if lateInitializeRequeueErr != nil {
+                return lateIntializedRequeueErr
+            }
+            // iterate through lateInitializedLatest.conditions for requeue
+        }
+        return nil
+    }
+    ```
+
+### Handling Async Resource Creation
+For handling the late initialization of resources, where server side defaults are not returned as part of CreateOutput, but are either returned in subsequent read calls or custom API calls, AWS service teams can make use of server-side-default hooks and return error from the `rm.SetLateInitializedDefaults()` method until late initialization fields are present. This error will trigger the requeue of reconciliation task and then succeed when SetLateInitializedDefaults() does not error out.
+
+There is also an option of mentioning static Delay for late initialized fields to get populated.
+
+This error will not impact patching the resource's metadata/spec/status. The updates will take place before requeuing due to unavailability of default fields. 
+
+-----------------------------
+
+
+## Alternate Solution
+
+### Solution 1
+In this solution, There will be updates in 
+1) **ACK Code Generator** to capture the late initialized fields from AWS operation output and use them to late initialize k8s custom resource object. Code generator changes will support a) configuration for service teams to tell ACK code generator which fields will be late initialized and b) configuration to provide custom code for late initialization.
+2) **ACK Runtime** to a) update the custom resource stored in etcd with late initalized optional spec fields And b) handle the late initalized fields for resources that have asynchronous creation.
+    > For the resources with Async creation, the late initalized fields are not readily available after create operation. Therefore the reconciler loop needs to be repeated for these resources until the late initalized fields are available.
+
+### ACK Code Generator Changes
+1. **Provide a new "LateInitializationConfig" struct for AWS service teams to provide which spec fields(spec.A) or spec field members(spec.A.X) will be late initialized.**
+
+    Only the fields/members mentioned as part of 'LateInitializationConfig' will be considered for late initialization. (The reason is explained inside step 3 "NOTE" )
+
+    `LateInitializationConfig` struct will be the member of `ResourceConfig` and it will look like following:
+
+    ```go
+    type LateIntializationConfig struct { 
+    
+    // valid values: 'readOne, readMany, create, update'
+    DefaultSourceMethod *string `json:"default_source_method"`
+    
+    DefaultedSpecFields []DefaultedSpecFieldConfig `json:"defaulted_spec_fields"`
+    }
+
+
+    type DefaultedSpecFieldConfig struct {
+    
+    // valid values: 'readOne, readMany, create, update'
+    // If this is not mentioned, then default_source_method is used
+    SourceMethod *string `json:"source_method,omitempty"`
+    
+    // Path to the field relative to Spec, which will be late initialized
+    // Example for primitives: "a" to replace "desired.ko.Spec.a" with "<source_method>Latest.ko.spec.a"
+    // Example for struct: "a.b" to replace "desired.ko.Spec.a.b" with "<op_name>Latest.ko.spec.a.b"
+    // Example for map: "a.b.[c]" to replace "desired.ko.Spec.a.b[c]" with "<op_name>Latest.ko.spec.a.b[c]"
+    // TODO: support for List
+    
+    Path *string `json:"path"` 
+    
+    // TODO: possible support for source path and target path, which do not have to be exactly similar as they are at present.
+    }
+    ```
+
+2. **Provide 3 new hooks to AWS service teams to customize the code generation in step 5**
+    
+    These 3 hooks will be:
+    
+    a) **set_late_initialized_defaults_override**: This hook will provide the **complete implementation** for setting default values of optional spec fields which are not provided by the k8s user and defaulted by AWS service.
+    This hook is mutually exclusive with "pre_set_late_initialized_defaults" & "post_set_late_initialized_defaults" hooks.
+
+    b) **pre_set_late_initialized_defaults**:- This hook will provide the custom code to be inserted **before** setting server side defaults for missing optional spec fields.
+
+    c) **post_set_late_initialized_defaults**: This hook will provide the custom code to be inserted **after** setting server side defaults for missing optional spec fields.
+
+3. **Updating Spec fields in sdkCreate and sdkUpdate methods**
+
+    Currently sdkCreate method prepares Create reques from desired object's Spec fields and updates the Status fields of latest object with Create response.
+    
+    The missing part here is that some late initialized fields are returned as part of create Response should get populated in the spec of latest object.
+
+    > NOTE: All spec fields cannot be updated from Create Response because spec fields types are created from Create Input and there can be type mismatch even with same field names.
+
+    Based on the spec fields mentioned in `LateInitializationConfig` by AWS service teams, the Create response will be used to late initalize these spec fields in the latest object.
+
+* Current Code:
+
+    ```go
+    // sdk.go.tpl    
+    func (rm *resourceManager) sdkCreate(
+	ctx context.Context,
+	desired *resource,
+    ) (created *resource, err error) {
+        ...
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}
+        ...
+    }
+    ```
+
+* Proposed Code Changes:
+
+    ```go
+    // sdk.go.tpl  
+
+    func (rm *resourceManager) sdkCreate(
+	ctx context.Context,
+	desired *resource,
+    ) (created *resource, err error) {
+        ...
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 false }}
+        {{GoCodeSetDefaultsOutput .CRD "resp" "ko" OpTypeCreate 1 }}
+        ...
+    }
+
+    //controller.go
+    
+    "GoCodeSetDefaultsOutput": func(r *ackmodel.CRD, sourceVarName string, targetVarName string,
+    opType model.OpType,
+    indentLevel int) string {
+			return code.SetSpecDefaults(r.Config(), r, opType, sourceVarName, targetVarName, indentLevel)
+		}
+
+    //set_resource.go
+    
+    func SetSpecDefaults(
+        cfg *ackgenconfig.Config,
+        r *model.CRD,
+        opType model.OpType,
+        sourceVarName string,
+        targetVarName string,
+        indentLevel int
+    ) {
+        // 1. read the late initialization config
+        // 2. filter and only consider fields with source={{opType}}
+        // 3. for each field from #2
+            
+            // 3a. Split the field string on '.'
+            
+            // 3b. if the first  split string exists is CRD Spec fields, start adding null checks for each split string. These null checks would make sure that only missing optional fields get initialized in #3c
+            
+            // 3c. For the last split string, find the member shape and use the existing code to generate assignment from "resp.{{path}}(source)" to "ko.{{path}}(target)" 
+
+        // 4. return output created from #3.
+    }
+    ```
+
+4. **sdkUpdate will be modified similarly as sdkCreate.** 
+    > NOTE: For late initialized field with source 'create' and 'update', it is important to update the latest spec inside sdkCreate or sdkUpdate method because these operations will be called only once unlike readOne or readMany which can be invoked multiple times successfully to read late initialized fields.
+
+5. Add `AWSResourceManager.SetLateInitializedDefaults()` method to handle server side defaults with source `readOne`, `readMany` or custom code.
+
+* Proposed Pseudo Code
+    ```go
+    // manager.go.tpl
+    func (rm *resourceManager) SetLateInitializedDefaults(
+	ctx context.Context,
+    latest acktypes.AWSResource,
+    ) (acktypes.AWSResource, error) {
+        {{- if $hookCode := Hook .CRD "set_late_initialized_defaults_override" }}
+            {{ $hookCode }}
+        {{- else}}
+            {{- if $hookCode := Hook .CRD "pre_set_late_initialized_defaults" }}
+                {{ $hookCode }}
+            {{- end }}
+
+            // if lateInitializationConfig has readOne source path 
+            // resp := readOne 
+            {{GoCodeSetDefaultsOutput .CRD "resp" "ko" OpTypeReadOne 1 }}
+
+            // if lateInitializationConfig has readMany source path 
+            // resp := readMany 
+            {{GoCodeSetDefaultsOutput .CRD "resp" "ko" OpTypeReadMany 1 }}
+
+            {{- if $hookCode := Hook .CRD "post_set_late_initialized_defaults" }}
+                {{ $hookCode }}
+            {{- end }}
+            // return latest with server side default spec updates 
+        {{- end }}
+    }
+    ```
+
+> NOTE: Currently hooks are not present for Create and Update output responses. Hooks will only be present for `AWSResourceManager.SetLateInitializedDefaults()` method. Using CreateOutput and UpdateOutput for setting server side defaults seems straight forward without needing any custom code.
+
+
+### ACK Runtime Changes
+
+* Current reconciler source code
+    ```go
+    func (r *reconciler) Sync() {
+        var latest acktypes.AWSResource
+        latest, err := rm.ReadOne()
+
+        if err!= nil {
+            if err != NotFound {
+                patchResource() // only status
+                return err
+            }
+            latest,err := rm.Create()
+            if err != nil {
+                patchResource() // only status
+                return err
+            }
+        } else {
+            delta := r.rd.Delta(desired, latest)
+            If delta.DifferentAt("Spec") {
+                latest,err := rm.Update()
+                if err != nil {
+                    patchResource() // only status
+                    return err
+                }
+            }
+        }
+
+        r.patchResource(ctx, desired, latest) // only status
+
+        // iterate through latest.conditions for requeue
+
+        return nil
+    }
+    ```
+
+* Proposed reconciler source code
+
+    ```go
+    func (r *reconciler) Sync() {
+        observed, err := rm.ReadOne()
+        var latest acktypes.AWSResource
+
+        if err!= nil {
+            if err != NotFound {
+                patchResourceStatus()
+                return err
+            }
+            // rm.Create() also updates the server side defaults from CreateResponse
+            latest,err := rm.Create()
+            if err != nil {
+                patchResourceStatus()
+                return err
+            }
+        } else {
+            delta := r.rd.Delta(desired, observed)
+            If delta.DifferentAt("Spec") {
+                // rm.Update() also updates the server side defaults from UpdateResponse
+                latest,err := rm.Update()
+                if err != nil {
+                    patchResourceStatus()
+                    return err
+                }
+            }
+        }
+
+        // updates the latest object with server side defaults
+        // this operation will support hooks for service team to use.
+        setDefaultErr := rm.SetLateIntializedDefaults(ctx, latest)
+
+        // First patch the latest object's Spec+Metadata, in case there was any update in
+        // metadata.Annotations OR if Create/Update operation added any server side defaults.
+        // Also patch the latest object's Status.
+        // Once the resource is patched then requeue if "setDefaultErr" is not nil
+
+        r.patchResource(ctx, desired, latest) // patches metadata, spec and status if needed
+
+        if setDefaultErr != nil {
+            return requeue.NeededAfter(setDefaultErr, TBD)
+        }
+
+        // iterate through latest.conditions for requeue
+
+        return nil
+    }
+
+    func (r *reconciler) patchResource(ctx, desired, latest) err {
+        err := r.patchResourceSpecAndMetadata(ctx, desired, latest)
+        if err != nil {
+            return err
+        }
+
+        err = r.patchResourceStatus(ctx, desired, latest)
+        if err != nil {
+            return err
+        }
+    }
+
+    func (r *reconciler) patchResourceSpecAndMetadata(ctx, desired, latest) err {
+        // if there is diff in spec or metadata between desired and latest,
+        // call kc.Patch()
+    }
+
+    func (r *reconciler) patchResourceStatus(ctx, desired, latest) err {
+        // if r.rd.UpdateCRStatus(latest) returns true, call kc.Status().Patch()
+    }
+    ```
+
+### Handling Async Resource Creation
+For handling the late initialization of resources, where server side defaults are not returned as part of CreateOutput, but are either returned in subsequent read calls or custom API calls, AWS service teams can make use of server-side-default hooks and return error from the `rm.SetLateInitializedDefaults()` method until late initialization fields are present. This error will trigger the requeue of reconciliation task and then succeed when SetLateInitializedDefaults() does not error out.
+
+This error will not impact patching the existing resource's metadata/spec/status. The updates will take place before requeuing due to unavailability of default fields. 
+
+### Solution 2
+
+#### TL;DR: 
+
+There will be a new `r.rd.AddLateInitializedFields()` method call added in the reconciliation loop, which will return `true` if the desired spec was updated with late initialized field(s) from the latest spec. If so, then reconciliation loop will patch the desired spec with new updates. 
+
+See the following sections for detailed explanation of this solution: 
+
+#### Reconciler Updates
+
+* Create separate variables for `readOneLatest`, `createLatest` and `updateLatest` instead of a single `latest` variable which gets overwritten after readOne/Create/Update operation in the reconciliation loop.
+* Create a `desiredBase` which is copy of `desired` object after the status is updated(`patchResource()`) in etcd by the reconciler. We will use this copy as base to patch the desired resource spec in etcd.
+* Add `specUpdated, err := r.rd.AddLateInitializedFields(desired, readOneLatest, createLatest, updateLatest)`
+    > If the service teams have provided the configuration for populating late initialized field into `generator.yaml`, this method will update the desired spec accordingly and return `true, nil`. Otherwise it will return `false, nil`.
+* If `specUpdated` is `true`, reconciler will patch the resource in etcd using `desiredBase` as base object, otherwise no patching of K8s custom resource will take place and reconciliation will finish. 
+
+#### Sdk.go Updates
+
+* Allow `sdkCreate` and `sdkUpdate` to populate the `Spec` of `latest` object and not just the `Status`.
+* This will allow capturing the late initialized fields from Create and Update API output.
+
+#### ResourceDescriptor Updates
+
+* A  new method will be added in `ResourceDescriptor`, which will look like
+    `AddLateInitializedFields(desired, readOneLatest, createLatest, updateLatest)`
+* The default implementation of this method will be to return `false, nil` if there is no entry in `generator.yaml` for handling late initialized fields.
+* If AWS service team has mentioned some fields that are supposed to be late initialized, this method will perform
+    `desired.ko.Spec.{{path_from_generator.yaml}} = {{operation_from_generator.yaml}}Latest.ko.Spec.{{path_from_ganarator.yaml}}`
+* The above statement will be executed with proper non-nil check on latest object, proper non-nil check on desired object and making sure the existing entry does not exist in the desired spec (i.e. Only perform assignment if the value is absent in desired spec).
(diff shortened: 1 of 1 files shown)


docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+
+❌ → not achievable
+
+### Solution 1
+
+In this solution, the status fields will include all the fields which are CreateOutput shape members. Current status 
+fields are the CreateOutput shape members which are not present in CreateInput shape (spec fields).
+
+During reconciliation, the Read(One|Many),Update request inputs will be created from desired state but giving precedence
+to status fields over spec in case of conflict.
```
RedbackThomson, line 49: This case seems to imply that there could be fields in `status` with the same name as ones in `spec`? The ones in `status` are then given precedence, which would mean we never use the values in `spec` again. Therefore the `spec` fields are only referenced for the initial `create` and never again? 

Or are you implying here that the `status` fields are simply used by the `diff` and for UX purposes, and that they're not being used as the source of truth for update operations?

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+    describe their custom resources.
+2. *Consistency across ACK service controllers* ✅
+    1. There should be no complexity in following this approach across all ACK service controllers.
+3. *GitOps compatibility* ✅
+    1. There should be no complexity in implementing GitOps for ACK custom resources. 
+    2. Approach mentioned in low level design below should work.
+4. *User experience for custom resource lifecycle* ✅
+    1. Applying updates to custom resource can become confusing for the user.
+    2. To update custom resource spec, K8s user will have to check resource status because status will have most updated
+     value of all the fields.
```
RedbackThomson, line 63: This is seen as `doable`? These subpoints seem like detractors that make this solution impossible.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+* The number of status fields will be very large and there will be a lot of duplication in custom resource description
+
+
+### Solution 2
+
+This solution builds on top of “Solution 1”, with the assumption that “Problem” mentioned above is not very common 
+amongst AWS APIs.
+
+In this solution, service teams can specify the spec fields which will be duplicated in both spec and status of custom
+resource using generator.yaml
```
RedbackThomson, line 76: Unfortunately I don't think this provides any value over solution 1, except for reducing duplication. Understand the reasoning for proposing this, though.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+4. *User experience for custom resource lifecycle* ✅
+    1. Applying updates to custom resource can become confusing for the user.
+    2. To update custom resource spec, K8s user will have to check resource status because status will have most updated
+     value for some of the fields.
+
+
+
+### Solution 3 (Preferred Solution)
+
+In this solution, If a default value is added or existing value is updated as part of Create/Update API output, ACK service controller will update the spec/status of the custom resource accordingly. These updates will be persisted in custom resource’s Annotation to refer in future reconciliations.
```
RedbackThomson, line 96: By `annotation` are you referring to `metadata.annotations` or another list in `status`, like `conditions`?

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+* A basic “Guestbook” [operator](https://github.com/vijtrip2/guestbook/blob/main/controllers/guestbook_controller.go) from
+ kubebuilder getting started guide. 
+* “Guestbook” [manifest](https://github.com/vijtrip2/flux-get-started/blob/master/guestbook/guestbook-sample.yaml) being
+ deployed using fluxcd. 
+* Controller adds a defaultValue in spec on each run, updates an existing value in spec and updates the annotation on each
+ reconcile loop.
+
+Observations:
+
+* During the start of each reconciliation, the spec and status are reset to values present in the git repository.
```
RedbackThomson, line 141: This will cause an extra `update` operation for every CR every time Flux syncs a new commit?

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+* Annotations are not reset, if a value is updated in annotations, it is present for future reconciliations.
+* Changes made after reconciliation are not reflected back into the git repository. “Fluxcd” does not support this.
```
RedbackThomson, line 143: What do you mean by `reflected back into the git repository`? A new commit modifying the GitOps repository? I'd sure hope no K8s CD system modifies the repo haha

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+           \“overrideFieldC\”:
+               {
+                   \“userInput\”:\“valueC\”,
+                   \“serverOverride\”:\“valueC-xyz\”
+               }
+       }”
+       Spec:
+        fieldA: valueA
+        defaultFieldB: valueB
+        overrideFieldC: valueC-xyz
```
RedbackThomson, line 202: I disagree that we should persistent these overrides back into etcd. This would appear to the user that the controller is modifying their GitOps deployment. GitOps tools like flux or ArgoCD have options in place to "heal" resources that are modified. That is, if we modify the spec, GitOps tools will modify them back.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,277 @@
...
+
+3. During the next reconciliation of manifest mentioned in #1, the desired state will be constructed not just from spec 
+   in etcd but also the Annotations. and the desired object will look like the object persisted in etcd in #2
+
+   After reading the desired state from etcd, the reconciler code will go through all the spec fields and see whether 
+   they exist in the spec-override annotation map. If a corresponding entry exists, reconciler will marshall current 
+   desired value of spec field and compare with “userInput” in annotation override, 
+
+   * If there is no difference between “userInput” and marshalled desiredValue, reconciler will update the desired state
+     with serverOverride from spec-override annotation.
```
RedbackThomson, line 223: Just throwing this crazy idea out there - I hope that no AWS service has implemented this:
What if the server side value is non-deterministic? That is, we give it the same value on the `update` but it returns a different `serverOverride`?

RedbackThomson (commented):
Wow, really good stuff, here. Your solution is actually a lot more elegant than I had initially imagined. Great detective work with flux. Left some comments nitpicking specifics

docs/design/proposals/late-initialization/late-initialization.md
jaypipes: @RedbackThomson ElastiCache's `User.AccessString` field is an example of this. The value that the k8s user places in the `User.AccessString` field is *always* overridden by the ElastiCache service and returned as the `AccessString` in the `DescribeUser` API call.

jaypipes (changes requested):
@vijtrip2 I would like you to break this PR into two separate proposals, one that specifically addresses the late initialization of fields that the user does *not* specify and the server sets to a default value and another proposal for handling fields where the k8s user *does* specific a value for the field, but the backend service overrides that specified value.

The two use cases are substantially different and I'd like to tackle the former use case (defaulting field values that the k8s user did *not* specify) ASAP before moving on to tacklling the server-override field use case.

vijtrip2:
> @vijtrip2 I would like you to break this PR into two separate proposals, one that specifically addresses the late initialization of fields that the user does _not_ specify and the server sets to a default value and another proposal for handling fields where the k8s user _does_ specific a value for the field, but the backend service overrides that specified value.
> 
> The two use cases are substantially different and I'd like to tackle the former use case (defaulting field values that the k8s user did _not_ specify) ASAP before moving on to tackling the server-override field use case.

Sounds good. I actually started implementing server-override, that's why this PR went stale. 
I will prioritize work on server-side defaults instead.

docs/design/proposals/late-initialization/late-initialization.md
echen-98: @RedbackThomson I'm not sure that it's nondeterministic – although we can't necessarily translate between the user-specified `AccessString` and the server-returned `AccessString`, you should get the same thing every time. However, we should probably assume that the way the expansion/translation is done on the server-side is subject to change.

more info here: https://github.com/aws-controllers-k8s/elasticache-controller/pull/32

echen-98:
> @vijtrip2 I would like you to break this PR into two separate proposals, one that specifically addresses the late initialization of fields that the user does not specify and the server sets to a default value and another proposal for handling fields where the k8s user does specific a value for the field, but the backend service overrides that specified value.

> The two use cases are substantially different and I'd like to tackle the former use case (defaulting field values that the k8s user did not specify) ASAP before moving on to tacklling the server-override field use case.

+1, I think "default fields" and "override fields" need different solutions.

I would request that [this comment](https://github.com/aws-controllers-k8s/community/issues/812#issuecomment-866421105) be considered when coming up with solutions for default fields.

docs/design/proposals/late-initialization/late-initialization.md
echen-98: +1. 

We can use this annotations solution to override values when generating the delta so the original `Spec` will be preserved.

vijtrip2:
> > @vijtrip2 I would like you to break this PR into two separate proposals, one that specifically addresses the late initialization of fields that the user does not specify and the server sets to a default value and another proposal for handling fields where the k8s user does specific a value for the field, but the backend service overrides that specified value.
> 
> > The two use cases are substantially different and I'd like to tackle the former use case (defaulting field values that the k8s user did not specify) ASAP before moving on to tacklling the server-override field use case.
> 
> +1, I think "default fields" and "override fields" need different solutions.
> 
> I would request that [this comment](https://github.com/aws-controllers-k8s/community/issues/812#issuecomment-866421105) be considered when coming up with solutions for default fields.

@echen-98 We decided in a previous meeting that late initialized fields will be populated back in resource spec. And the new solution will be based on that for late initialized(i.e.) default fields. More details will be there in future PR. 

Going that path, "the comment" regarding nil difference will not be required for handling unnecessary updates.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: They were supposed to be used as source of truth in this solution. But thankfully we are not implementing this solution.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: By the :heavy_check_mark: , I meant it is possible to have an implementation like this, but we are certainly not going to implement because of subpoints.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: I checked again and flux does not override the late initialized values. For server-side-overrides, we decided not to update the spec(only capture the diffs in annotation).
So the new update will not be triggered by flux, if there is no change in git repo.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Yeah, that's what I meant.
Fluxcd does update the git repo back but only if you make changes to k8s object using flux api.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Yes, we went with the approach not to update the resource spec for server-side-overrides. Late initialized fields will be written to the resource spec.

vijtrip2:
* I have updated this PR to only deal with `late initialized fields`
* I have created a separate PR [#856](https://github.com/aws-controllers-k8s/community/pull/856) to propose a solution for server-side-overrides.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+See the following sections for detailed explanation of this solution: 
+
+### Reconciler Updates
+
+* Create separate variables for `readOneLatest`, `createLatest` and `updateLatest` instead of a single `latest` variable which gets overwritten after readOne/Create/Update operation in the reconciliation loop.
+* Create a `desiredBase` which is copy of `desired` object after the status is updated(`patchResource()`) in etcd by the reconciler. We will use this copy as base to patch the desired resource spec in etcd.
+* Add `specUpdated, err := r.rd.AddLateInitializedFields(desired, readOneLatest, createLatest, updateLatest)`
+    > If the service teams have provided the configuration for populating late initialized field into `generator.yaml`, this method will update the desired spec accordingly and return `true, nil`. Otherwise it will return `false, nil`.
+* If `specUpdated` is `true`, reconciler will patch the resource in etcd using `desiredBase` as base object, otherwise no patching of K8s custom resource will take place and reconciliation will finish. 
+
```
jaypipes, line 42: I'm not sure I'm no board with the above proposed solution. I was thinking that we would add a new `pkg/types.AWSResourceManager.SetOptionalFieldDefaults()` interface method and call that immediately after a successful call to `AWSResourceManager.Create()`.

```go
type AWSResourceManager interface {
...
    // SetDefaults determines if any optional fields in the supplied resource
    // have not been set by the user OR set by the resource manager in the response
    // from the Create output shape. If any such fields exist, SetDefaults calls
    // one or more AWS service API calls to retrieve the service-side default value of those
    // fields and returns a copy of the supplied resource with those default values set on the
    // resource's `Spec`.
    SetDefaults(AWSResource) (AWSResource, error)
}
```

In other words, instead of [this](https://github.com/aws-controllers-k8s/runtime/blob/089321f8f32690fab024fd0e633a0f70e9726262/pkg/runtime/reconciler.go#L232-L246):

```go
		rlog.Enter("rm.Create")
		latest, err = rm.Create(ctx, desired)
		rlog.Exit("rm.Create", err)
		if err != nil {
			if latest != nil {
				// this indicates, that even though Create failed
				// there is some changes available in the latest.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latest)
			}
			return err
		}
		rlog.Info("created new resource")
```

we would do something like this:

```go
		rlog.Enter("rm.Create")
		latest, err = rm.Create(ctx, desired)
		rlog.Exit("rm.Create", err)
		if err != nil {
			if latest != nil {
				// this indicates, that even though Create failed
				// there is some changes available in the latest.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latest)
			}
			return err
		}
		rlog.Enter("rm.SetDefaults")
		latestWithDefaults, err = rm.SetDefaults(ctx, latest)
		rlog.Exit("rm.SetDefaults", err)
		if err != nil {
			if latestWithDefaults != nil {
				// this indicates, that even though SetDefaults failed
				// there is some changes available in the latestWithDefaults.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latestWithDefaults)
			}
			return err
		}
		if rlog.DebugEnabled() {
			delta := r.rd.Delta(latest, latestWithDefaults)
			if delta.DifferentAt("Spec") {
				rlog.Debug("set default values on resource", "diff", delta.Differences)
			}
		}
		rlog.Info("created new resource")
```

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+This **late initialization** of AWS resource creates a difference between desired and latest state of k8s custom resource because currently the desired manifest does not get updated with these late initialized properties.
+
+This document will propose the solution for above problem and provide the details of implementing it.
+
+## Requirements
+
+A solution for this problem should :
+
+1. Update the desired state of the K8s custom resource with late initialized fields.
```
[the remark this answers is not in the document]
vijtrip2: Yes, That is the intention. Only update the fields which were originally not present in the desired manifest.
Updating the language.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Requirements
+
+A solution for this problem should :
+
+1. Update the desired state of the K8s custom resource with late initialized fields. 
+
+2. Provide configuration for AWS service teams to indicate which operation output returns the late initialized fields. i.e. Get/List/Create/Update output.
+
+    > AWS APIs have different behavior for returning the late initialized to the user. Some services return these fields as part of Create output, while some services will return the late initialized fields as part of  ReadOne output after the resource creation.
```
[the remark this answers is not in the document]
vijtrip2: Yes, There is a mechanism of 3 `hooks` for provide custom-code for late initialized fields.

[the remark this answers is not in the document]
RedbackThomson: Should this possibly be changed to `indicate which field may be late initialized`? This leaves the mechanism by which it is updated (i.e. the operation to fetch it) as an implementation detail.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: I see one concerns with the solution in comment:
**"SettingDefault" coupled with resource creation**
* This coupling can be problematic because I believe some APIs do not return late initialized fields as part of create response. Ex: `elasticache`, which does the asynchronous creation and the late initialized fields eventually showup as result of Get call after asynchronous creation has finished. 
* Another problem in the same vein is that `Create` resource call is only made once, if the default values are set as part of an `Update operation`, the above solution will not capture them. I can think of an example where customer just enables logging on their `apigateway stage` using `Update` operation, which sets the default value of `log level` to `Info`. If the logging is not enabled this value is not set.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+* If `specUpdated` is `true`, reconciler will patch the resource in etcd using `desiredBase` as base object, otherwise no patching of K8s custom resource will take place and reconciliation will finish. 
+
+### Sdk.go Updates
+
+* Allow `sdkCreate` and `sdkUpdate` to populate the `Spec` of `latest` object and not just the `Status`.
+* This will allow capturing the late initialized fields from Create and Update API output.
+
+### ResourceDescriptor Updates
+
+* A  new method will be added in `ResourceDescriptor`, which will look like
```
RedbackThomson, line 50: I think this functionality will be more broadly implemented than just in `ResourceDescriptor`. I think it could be extended to be a separate composition as part of `AWSResourceManager`. Further, I think it could be split into separate handlers for each operation type (`ReadOne`, `Update`, `Create`, `ReadMany`, etc.)

```golang
type AWSResourceManager interface {
    LateInitializationManager
    ...
}

type LateInitializationManager {
    SetCreateDefaults(AWSResource) (AWSResource, error)
    SetUpdateDefaults(AWSResource) (AWSResource, error)
    SetReadDefaults(AWSResource) (AWSResource, error)
    ...
}
```

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. Update the desired state of the K8s custom resource with late initialized fields.  ✅
+
+2. Provide configuration for AWS service teams to indicate which operation output returns the late initialized fields. i.e. Get/List/Create/Update output. ✅
+
+3. Wait for asynchronous resource creation. ✅
+
+    > Service teams can add a hook which will look at the state of latest object and determine if the reconciliation loop should run again. If so, they can return an error from the “AddLateInitializedFields()” method and the reconciliation loop will be triggered again.
+
+4. Allow the service teams to provide custom code using hooks for handing of late initialized fields. ✅
```
RedbackThomson, line 113: Did you evaluate other possible solutions? If not, lets at least add the discussions from this Github PR into the document for posterity.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,113 @@
...
+
+    > AWS APIs have different behavior for returning the late initialized to the user. Some services return these fields as part of Create output, while some services will return the late initialized fields as part of  ReadOne output after the resource creation.
+
+3. Wait for asynchronous resource creation to populate late initialized fields.
+
+    > For some services/resources the late initialized fields do not appear immediately.  This happens when there is asynchronous processing involved in resource creation. The proposed solution should allow waiting and retry until the late initialized fields appear in the mentioned API output.
+
+4. Allow the AWS service teams to provide custom code using hooks for handing of late initialized fields.
+
+## Solution
```
ellistarn, line 27: I'm not seeing any discussion around webhooks in this doc. The typical pattern for CRD initialization logic (defaulting) relies on admission webhooks. Has this been considered?

vijtrip2: The concerned default values in this problem are set by AWS services, and those will only be present during reconciliation.

Admission webhooks will be great if we wanted to add any default values to the manifest before controller reconciliation, but that is not the problem here.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: ACK Core decided to implement setting optional defaults field for the `Create` operation first, along with support for custom code.

Once this is implemented, ACK Core team will tackle `Update` and `ReadOne` operation separately as needed by the AWS service teams.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,151 @@
...
+
+    > AWS APIs have different behavior for returning the late initialized to the user. Some services return these fields as part of Create output, while some services will return the late initialized fields as part of  ReadOne output after the resource creation.
+
+3. Wait for asynchronous resource creation to populate late initialized fields.
+
+    > For some services/resources the late initialized fields do not appear immediately.  This happens when there is asynchronous processing involved in resource creation. The proposed solution should allow waiting and retry until the late initialized fields appear in the mentioned API output.
+
+4. Allow the AWS service teams to provide custom code using hooks for handing of late initialized fields.
+
+## Solutions
```
jaypipes, line 27: Recommend changing this section to `## Proposed Solution` and structuring the section's content like so:

### ACK code-generator

> Describe proposed changes to how the code generator discovers server-side defaulting fields>

> Describe proposed changes to the generator config that allows service teams to inform the code generator when it cannot figure out which fields have a server-side defaulting behaviour>

### ACK runtime

> Describe proposed changes to the main ACK runtime reconciler loop for handling server-side default values for resource fields that are set immediately, including before and after pseudo-code>

> Describe proposed changes to the main ACK runtime reconciler loop for handling server-side default values for resource fields that are set after some delay (asynchronously), including before and after pseudo-code

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,151 @@
...
+See the following sections for detailed explanation of this solution: 
+
+#### Reconciler Updates
+
+* Create separate variables for `readOneLatest`, `createLatest` and `updateLatest` instead of a single `latest` variable which gets overwritten after readOne/Create/Update operation in the reconciliation loop.
+* Create a `desiredBase` which is copy of `desired` object after the status is updated(`patchResource()`) in etcd by the reconciler. We will use this copy as base to patch the desired resource spec in etcd.
+* Add `specUpdated, err := r.rd.AddLateInitializedFields(desired, readOneLatest, createLatest, updateLatest)`
+    > If the service teams have provided the configuration for populating late initialized field into `generator.yaml`, this method will update the desired spec accordingly and return `true, nil`. Otherwise it will return `false, nil`.
+* If `specUpdated` is `true`, reconciler will patch the resource in etcd using `desiredBase` as base object, otherwise no patching of K8s custom resource will take place and reconciliation will finish. 
+
```
jaypipes, line 44: and then move this information to an `## Alternate solutions considered` section at the bottom of the doc.

jaypipes (changes requested):
@vijtrip2 I'd like to see a fairly substantial reworking of this proposal document. See inline for details, but overall, I'd like to see the following layout used:

```
# Problem overview
# Solution requirements
# Proposed solution
## ACK code-generator
## ACK runtime
# Alternate solutions considered
```

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,338 @@
...
+    // Path to the field relative to Spec, which will be late initialized
+    // Example for primitives: "a" to replace "desired.ko.Spec.a" with "<source_method>Latest.ko.spec.a"
+    // Example for struct: "a.b" to replace "desired.ko.Spec.a.b" with "<op_name>Latest.ko.spec.a.b"
+    // Example for map: "a.b.[c]" to replace "desired.ko.Spec.a.b[c]" with "<op_name>Latest.ko.spec.a.b[c]"
+    // TODO: support for List
+    
+    Path *string `json:"path"` 
+    
+    // TODO: possible support for source path and target path, which do not have to be exactly similar as they are at present.
+    }
```
jaypipes, line 65: I was thinking that you would have a `LateInitializeConfig` struct that would store instructions for server-side defaulting (and eventually, server-side overrides). And that you would have a `late_initialize` field in the [`FieldConfig` struct](https://github.com/aws-controllers-k8s/code-generator/blob/c77aa9c75d944952dee198029ba9822691cd82b0/pkg/generate/config/field.go#L139-L187).

Something like this maybe...

```go
// LateInitializeConfig contains instructions for how to handle the retrieval and
// setting of server-side defaulted or overridden fields.
type LateInitializeConfig struct {
    // DefaultImmediately indicates that if the user has not specified a value for
    // the field, then directly after a successful call to Create, the reconciler will
    // call ReadOne to get the value of the field as defaulted by the AWS service,
    // set the value of this field on the resource and patch the resource back to
    // the Kubernetes API.
    DefaultImmediately bool `json:"default_immediately"`
}

...
type FieldConfig struct {
...
    LateInitialize *LateInitializeConfig `json:"late_initialize,omitempty"`
}
```

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,338 @@
...
+    These 3 hooks will be:
+    
+    a) **server_side_defaults_initialization_override**: This hook will provide the **complete implementation** for setting default values of optional spec fields which are not provided by the k8s user and defaulted by AWS service.
+    This hook is mutually exclusive with "pre_server_side_defaults_initialization" & "post_server_side_defaults_initialization" hooks.
+
+    b) **pre_server_side_defaults_initialization**:- This hook will provide the custom code to be inserted **before** setting server side defaults for missing optional spec fields.
+
+    c) **post_server_side_defaults_initialization**: This hook will provide the custom code to be inserted **after** setting server side defaults for missing optional spec fields.
+
+3. **Updating Spec fields in sdkCreate and sdkUpdate methods**
```
jaypipes, line 79: I guess I don't understand why we want to make any changes in the `sdkCreate` and `sdkUpdate` methods. Instead, what I thought was being added was a change to the `resourceReconcilers.Sync` method that would call `ReadOne`  the resource manager immediately after calling Create when there were fields in the generator config marked as `DefaultImmediately` **AND** when the user had not specified those fields in the desired CR already.

In other words, we would just change the `resourceReconciler.Sync` method implementation from this:

```go
	rlog.Enter("rm.ReadOne")
	latest, err = rm.ReadOne(ctx, desired)
	rlog.Exit("rm.ReadOne", err)
	if err != nil {
		if err != ackerr.NotFound {
			if latest != nil {
				// this indicates, that even though ReadOne failed
				// there is some changes available in the latest.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latest)
			}
			return err
		}
		if isAdopted {
			return ackerr.AdoptedResourceNotFound
		}
		// Before we create the backend AWS service resources, let's first mark
		// the CR as being managed by ACK. Internally, this means adding a
		// finalizer to the CR; a finalizer that is removed once ACK no longer
		// manages the resource OR if the backend AWS service resource is
		// properly deleted.
		if err = r.setResourceManaged(ctx, desired); err != nil {
			return err
		}

		rlog.Enter("rm.Create")
		latest, err = rm.Create(ctx, desired)
		rlog.Exit("rm.Create", err)
		if err != nil {
			if latest != nil {
				// this indicates, that even though Create failed
				// there is some changes available in the latest.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latest)
			}
			return err
		}
		rlog.Info("created new resource")
...
```

to this:

```go
	rlog.Enter("rm.ReadOne")
	latest, err = rm.ReadOne(ctx, desired)
	rlog.Exit("rm.ReadOne", err)
	if err != nil {
		if err != ackerr.NotFound {
			if latest != nil {
				// this indicates, that even though ReadOne failed
				// there is some changes available in the latest.RuntimeObject()
				// (example: ko.Status.Conditions) which have been
				// updated in the resource
				// Thus, patchResource() call should be made here
				_ = r.patchResource(ctx, desired, latest)
			}
			return err
		}
		if isAdopted {
			return ackerr.AdoptedResourceNotFound
		}
		// Before we create the backend AWS service resources, let's first mark
		// the CR as being managed by ACK. Internally, this means adding a
		// finalizer to the CR; a finalizer that is removed once ACK no longer
		// manages the resource OR if the backend AWS service resource is
		// properly deleted.
		if err = r.setResourceManaged(ctx, desired); err != nil {
			return err
		}

		rlog.Enter("rm.Create")
		latest, err = rm.Create(ctx, desired)
		rlog.Exit("rm.Create", err)
		if err != nil {
			if latest != nil {
				// latest will have any fields in the Spec that were set to server-side defaulted
				// values that were in the Create Output shape and any Status fields, so we
				// patch the entire CR here, Spec/Metadata as well as Status.
				_ = r.patchResource(ctx, desired, latest)
			}
			if r.rd.NeedsLateInitialization(latest) {
				// We will get here only if there are any fields in the `latest` CR that
				// are `nil` AND are marked as having an immediately-available
				// server-side default value in the generator config.
				rlog.Enter("rm.ReadOne (LateInitiliazation")
				latestWithDefaults, err = rm.ReadOne(ctx, latest)
				rlog.Exit("rm.ReadOne (LateInitialization)", err)
				if err != nil {
					return err
				}
				// Patch the latestWithDefaults resource back to the Kubernetes API because
				// it will contain the Spec fields set to server-side defaulted values.
				_ = r.patchResource(ctx, latest, latestWithDefaults)
			}
			return err
		}
		rlog.Info("created new resource")
...
```

So, the `AWSResourceDescriptor.NeedsLateInitialization` method is the thing that would be a new interface method and would need to simply return a `bool` if the supplied resource has any fields that are set to `nil` but may have a server-side defaulted value (retrieved via the `AWSResourceManager.ReadOne` call.

jaypipes (changes requested):
@vijtrip2 I think you may have misunderstood previous comments I made. I've tried to highlight in the inline comments how I think the changes to the reconciler should be done in order to minimize changes needed to the AWSResourceManager interface...

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,338 @@
...
+1. **Provide a new "ServerSideDefaultsConfig" struct for AWS service teams to provide which spec fields(spec.A) or spec field members(spec.A.X) will be late initialized.**
+
+    Only the fields/members mentioned as part of 'ServerSideDefaultsConfig' will be considered for late initialization. (The reason is explained inside step 3 "NOTE" )
+
+    `ServerSideDefaultsConfig` struct will be the member of `ResourceConfig` and it will look like following:
+
+    ```
+    type ServerSideDefaultsConfig struct { 
+    
+    // valid values: 'readOne, readMany, create, update'
```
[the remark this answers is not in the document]
vijtrip2: For Create and Update, we do NOT set spec fields in the latest object because [this](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L93) parameter is true only for sdkFind operation. I have mentioned the reason in the proposal as well.

The reason is because Spec is created from CreateInput shape and there can be discrepancy of member shape if we try to construct/set spec fields from CreateOutput shape. 
One example is apigatewayv2 Integration resource's TLSConfig member. TLSConfig name is same in CreateInput and CreateOutput but their shapes are different.

Because of the above reason, I added Create and Update source in the above config. So that after reading this config, AWSResourceManager Create and Update will only set defaults for fields mentioned by service teams, not all spec fields.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: >"I guess I don't understand why we want to make any changes in the sdkCreate and sdkUpdate methods

My reasoning was that If the late initialized fields are present in the output of create/update response, why make extra readOne calls from reconciler.

And if there are some late initialized fields which will not be part of the Create/Update response, they will be handled separately in AWSResourceManager.SetDefaults() method, which will be called from reconciler.Sync() as mentioned in proposal. 
AWSResourceManager.SetDefaults() will handle those late initialized fields calling ReadOne/Many or any custom code provided by AWS Service teams using hooks.

nit: In the comment, r.patchResource(ctx, latest, latestWithDefaults) , we will have to patch on top of desired correct? so it should be like r.patchResource(ctx, desired, latestWithDefaults)

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: I can move LateInitializationConfig to FieldConfig. 
My reasoning for placing it under resourceConfig was to capture all late-initialization at one place rather than spread around among multiple Field configuration. 
But I can also see why FieldConfig would make sense since it's a late initialization is property of the field.

docs/design/proposals/late-initialization/late-initialization.md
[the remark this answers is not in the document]
jaypipes: > For Create and Update, we do NOT set spec fields in the latest object because [this](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L93) parameter is true only for sdkFind operation. I have mentioned the reason in the proposal as well.
> 
> The reason is because Spec is created from CreateInput shape and there can be discrepancy of member shape if we try to construct/set spec fields from CreateOutput shape.
> One example is apigatewayv2 Integration resource's TLSConfig member. TLSConfig name is same in CreateInput and CreateOutput but their shapes are different.

I think we need to revert that code and deal with the whole member field shape mismatch between Input and Output problem separately from whether we actively set the value of a Spec field to the value in a Create or Update Output shape.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Yeah, I can see your proposed approach working out.

My proposed solution was based on the fact that we do not update Spec fields during AWSResourceManager.Create/Update calls. But Allowing it will simplify handling of server-side-defaults. :+1:

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+    func (rm *resourceManager) sdkCreate(
+	ctx context.Context,
+	desired *resource,
+    ) (created *resource, err error) {
+        ...
+        // NOTE: The performSpecUpdate boolean param will be removed
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 }}
+        ...
+    }
+    ```
```
jaypipes, line 60: :+1:

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+        // LateInitializeConfig contains instructions for how to handle the
+        // retrieval and setting of server-side defaulted fields.
+        type LateInitializationConfig struct {
+            // DefaultImmediately indicates that if the user has not
+            // specified a value for the field, then directly after a
+            // successful call to Create/Update, the reconciler will call
+            // ReadOne to get the value of the field as defaulted by the
+            // AWS service, set the value of this field on the resource and
+            // patch the resource back to the Kubernetes API.
+            DefaultImmidiately bool `json:"default_immidiately"`
```
jaypipes, line 77: s/Immidiately/Immediately/g :)

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+            // DelaySeconds provides the number of seconds to wait to set
+            // late initialized fields after successful Create/Update call
+            // DelaySeconds is only used if DefaultImmidiately is false
+            // Default value for DelaySeconds is 5 seconds if DefaultImmidiately
+            // is false and DelaySeconds is absent.
+            DelaySeconds *int `json:"delay_seconds,omitempty"`
```
jaypipes, line 83: Do we need a `DefaultImmediately` field if we have a `DelaySeconds` field? You could make the `DelaySeconds` Go type an `int` instead of an `*int` and the default zero value would essentially mean "default immediately", right?

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+    These 3 hooks will be:
+    
+    a) **set_late_initialized_defaults_override**: This hook will provide the **complete implementation** for setting default values of optional spec fields which are not provided by the k8s user and defaulted by AWS service.
+    This hook is mutually exclusive with "pre_set_late_initialized_defaults" & "post_set_late_initialized_defaults" hooks.
+
+    b) **pre_set_late_initialized_defaults**:- This hook will provide the custom code to be inserted **before** setting server side defaults for missing optional spec fields.
+
+    c) **post_set_late_initialized_defaults**: This hook will provide the custom code to be inserted **after** setting server side defaults for missing optional spec fields.
+
+4. Add `AWSResourceManager.SetLateInitializedDefaults()` method to handle server side defaults for fields mentioned in LateInitializationConfig.
```
jaypipes, line 102: Let's just call this method `LateInitialize`.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+        type FieldConfig struct {
+            LateInitialize *LateInitializeConfig `json:"late_initialize,omitempty"`
+        }
+    ```
+
+3. **Provide 3 new hooks to AWS service teams to customize the code generation in step 4**
+    
+    These 3 hooks will be:
+    
+    a) **set_late_initialized_defaults_override**: This hook will provide the **complete implementation** for setting default values of optional spec fields which are not provided by the k8s user and defaulted by AWS service.
```
jaypipes, line 95: we don't use hooks for code that replaces entire methods (custom overrides). Hooks represent places within the regular code flow or code file that are executed if present. Instead, we use fields on a configuration struct to indicate that custom code entirely replaces a method.

For example, the [`pkg/generate/config.OperationConfig:CustomImplementation`](https://github.com/aws-controllers-k8s/code-generator/blob/c13e5fe097c1b487ac54d884761178e3cdeac3a7/pkg/model/crd.go#L479-L495) field is [used within the templates](https://github.com/aws-controllers-k8s/code-generator/blob/c13e5fe097c1b487ac54d884761178e3cdeac3a7/templates/pkg/resource/sdk.go.tpl#L57-L62) to entirely replace a particular implementation of a method.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+            // condition to the resource status indicating that resource is still
+            // being late initalized and return with requeue delay as net delay.
+            // 8. Make the ReadOne call.
+            // 9. Filter out the fields from step2, which are not-nil in ReadOne output.
+            // 10. For all the fields in step 9, set the lateInitialized value.
+            // 11. Upsert resource condition 'LateInitialized=True'.
+
+            {{- if $hookCode := Hook .CRD "post_set_late_initialized_defaults" }}
+                {{ $hookCode }}
+            {{- end }}
```
jaypipes, line 139: given where the proposed hook points are, let's call the two hooks:

`late_initialize_pre_read_one`, `late_initialize_post_read_one`

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,552 @@
...
+        lateInitializedLatest, lateInitializeRequeueErr := rm.SetLateInitializedDefaults(ctx, latest)
+
+        // Always persist the object state before exiting.
+        // patchResource will only Patch, if there is a diff.
+        if lateInitializedLatest != nil {
+            r.patchResource(ctx, desired, lateInitializedLatest) // patches metadata, spec and status if needed
+            if lateInitializeRequeueErr != nil {
+                return lateIntializedRequeueErr
+            }
+        // iterate through latest.conditions for requeue
```
jaypipes, line 225: should this be one line below? i.e. after the close of the `if` block?

jaypipes (commented):
This is much better. Getting close. A few suggested renames inline...

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Good point. I didn't realize earlier that empty structs can be mentioned in yaml.

How about following?

```go
type LateInitializationConfig struct {
    DelaySeconds int `json:"delay_seconds,omitempty"`
}
```
and `generator.yaml` will have entry like
```yaml
#For defaulting immediately
late_initialization_config: {}
``` 
```yaml
# for Defaulting with delay
late_initialization_config:
    delay_seconds: 10
```

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: I think this is at the correct place.

if lateInitializedLatest is not nil, 
a) patch the resource completely 
b) If rm.LateInitialize() requested a requeue, perform requeue 
c) Parse through conditions and determine requeue based for ConditionTypeResourceSynced

jaypipes (approved):
I'm good with this. If @RedbackThomson and @A-Hilaly are also good with it, let's merge away.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,532 @@
...
+        ...
+        // NOTE: The performSpecUpdate boolean param will be removed
+        {{ GoCodeSetCreateOutput .CRD "resp" "ko" 1 }}
+        ...
+    }
+    ```
+
+2. **Provide a new "LateInitializationConfig" struct inside generator config for AWS service teams to provide which spec fields will be late initialized.** These fields will be the ones which are late initialized by the server but not present in the Create and Update response.
+
+    These fields will be populated by making the readOne call. AWS service teams can also provide a delay after which these fields will be late initialized.
```
RedbackThomson, line 64: Your requirements outline that there should be sufficient wait/retry, but this seems to only specify a wait. How many attempts to read the late initialised fields will the reconciler make? Is this configurable?

As I understand for some resources there may be a very long delay for some fields - while waiting for the underlying instance to boot, for instance. In those cases, rather than specifying a 15 minute delay, I'd prefer to have it delay and retry every, say, 2 minutes - finally failing after a 15 minute total.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,532 @@
...
+    func (rm *resourceManager) LateInitialize(
+	ctx context.Context,
+    latest acktypes.AWSResource,
+    ) (acktypes.AWSResource, error) {
+        {{- if $hookCode := Hook .CRD "late_initialize_pre_read_one" }}
+            {{ $hookCode }}
+        {{- end }}
+
+        // 1. Filter all the fields which have LateInitializationConfig.
+        // 2. Filter out the fields from step1 with nil value in latest object.
```
RedbackThomson, line 103: Filter out the fields with `non-nil` value? We *want* the fields that have `nil` value?

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,532 @@
...
+        // 3. If number of fields after step2 are nonzero, calculate the delay
+        // for late initialization.
+        // 4. If the latest object has 'LateInitialized=False' condition,
+        // no need to delay the late initialization. Otherwise calculate
+        // net delay for late initialization.(why? see step 7).
```
RedbackThomson, line 108: nit: step 6

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,532 @@
...
+    ) (acktypes.AWSResource, error) {
+        {{- if $hookCode := Hook .CRD "late_initialize_pre_read_one" }}
+            {{ $hookCode }}
+        {{- end }}
+
+        // 1. Filter all the fields which have LateInitializationConfig.
+        // 2. Filter out the fields from step1 with nil value in latest object.
+        // 3. If number of fields after step2 are nonzero, calculate the delay
+        // for late initialization.
+        // 4. If the latest object has 'LateInitialized=False' condition,
```
RedbackThomson, line 106: I believe it could be best to keep an annotation on the object of the `LateInitializationStartTime`, for example. This way, rather than using the existance of the condition to determine when the delay has completed, we can compare it against our starting time. If current time > start time + delay, run late initialize. This will also help with supporting retrying.

We can still keep this condition, but it only serves the purpose of telling the user that `LateInitialization` has or hasn't finished yet.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Yeah, which shows their absence and eligibility to be late-initialized.

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: Fair point, I will think about retry, delay and update the comments/PR.

docs/design/proposals/late-initialization/late-initialization.md
A-Hilaly: IMO and at least in the first implementation, we should keep it as simple as possible as we don't have any similar behavior in any of the controllers.

A-Hilaly (approved):
Looks solid to me! I'm good with merging this PR once @RedbackThomson comments are addressed

docs/design/proposals/late-initialization/late-initialization.md
vijtrip2: I have added details of implementing retry w/exponential backoff into the PR proposal.

docs/design/proposals/late-initialization/late-initialization.md
RedbackThomson: Sorry, by this I meant it currently reads `filter out ... with nil value` whereas I think you mean `filter out ... with non-nil value`. "Filter out" meaning to "discard those with"

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,536 @@
...
+    These fields will be populated by making the readOne call. AWS service teams can also provide a delay after which these fields will be late initialized.
+    LateInitializationConfig will be added as part of FieldConfig. 
+
+    ```go
+        // LateInitializeConfig contains instructions for how to handle the
+        // retrieval and setting of server-side defaulted fields.
+        type LateInitializationConfig struct {
+            // DelaySeconds provides the base delay to attempt late initialization again after an unsuccessful attempt
+            // to late initialized fields from ReadOne output
+            DelaySeconds int `json:"delay_seconds,omitempty"`
```
RedbackThomson, line 73: nit: We should call this `MinDelaySeconds` (`min_delay_seconds`) since the reconciler logic will get the maximum across all fields, which means that most fields will be delayed by more than this amount.

docs/design/proposals/late-initialization/late-initialization.md
```diff
@@ -0,0 +1,536 @@
...
+        // 3. DeepCopy `latest` object into `latestWithDefaults`
+        // 4. Perform ReadOne operation and set late initialized fields in `latestWithDefaults` with nil checks
+        // 5. Collect the lateInitialized fields which are still not set in `latestWithDefault` into 
+        // `uninitializedFieldNames` slice.
+        // 6. If `uninitializedFieldNames` is not empty
+                  // 6.a. Using `services.k8s.aws/late-initialization-attempt` annotation, calculate the
+                  //    maximum delay needed for requeue amongst the `uninitializedFieldNames`
+                  // 6.b. Update the `LateInitialized` condition and `services.k8s.aws/late-initialization-attempt` annotation
+                  //    in latestWithDefaults object
+                  // 6.c. return latestWithDefaults, requeueNeededAfter(max-delay-from-6a)
```
RedbackThomson, line 118: Sorry, how does this support retries? This seems to imply it will continue indefinitely - it never references `MaxRetries`.

vijtrip2: My miss. Adding it now.

vijtrip2: I thought about it and I do not think we need MaxRetries for late initialization config.

In Kubernetes, the resource should keep trying to sync if the desired and latest state of object do not match, unless there is a terminal condition.

In the context of late initialization, I believe there is no terminal condition when trying to set defaults using ReadOne output. Since the service teams have defined a certain field will be defaulted at some point of time when it is not originally provided by the user, then the reconciler should keep up the attempts to find defaulted value of the field.

Your suggestion about exponential backoff with max backoff will help prevent the retry storm for these async cases.

RedbackThomson: That sounds solid. Exponential backoff while waiting for defaults, yep. It's also a one-way door should we ever need to implement that later down the road.

Issue #855, opened 2021-07-06
Handle server side field overrides


vijtrip2:
**Is your feature request related to a problem?**
During reconciliation, ACK service controller consumes the custom resource spec (i.e. desired state) from “etcd” and then creates/updates the AWS resource to match that desired state. ACK service controller uses the custom resource spec fields to construct Create/Update AWS API request.

However, some AWS API’s behavior does not completely rely on the fields present in Create/Update AWS API request alone. In the backend, these AWS APIs sometime update the value of field present in request with newly calculated value which represents the latest state of the resource in AWS. AWS APIs return these updated values with Create/Update API output and Read/List/Describe API output after resource creation.

Following problem arise due to this behavior of AWS APIs,

* When an AWS API does not strictly use the property value as is in the backend, but updates it, the desired spec of the K8s resource in the “etcd” does not correctly represent the actual state of resource in AWS. This is a problem because for K8s user, the spec of custom resource is the source of truth for state of K8s resource.

**Describe the solution you'd like**
ACK controller should store these overrides in custom resource's metadata.annotations and use it to infer the desired state of a resource.

jlbutler:
this proposal seems worth finally taking a look at, proposal in https://github.com/aws-controllers-k8s/community/pull/856

knottnt:
/close

largely implemented by late initialize feature.

PR #856, opened 2021-07-06
Proposal for handling server-side-overrides in ACK.

vijtrip2:
Issue #855 

Description of changes:

* Proposal for handling server-side overrides in ACK controllers. 
* The solution will help with handling the differences between desired and latest state of custom resource due to server side overrides during reconciliation loop.
* The solution will also provide a way to communicate to k8s user that server-side overrides have taken place for a custom resource.
* The solution will help AWS service teams simplify the custom `delta` logic, which they currently use to handle server-side overrides.


docs/design/proposals/server-side-override/server-side-override.md
```diff
@@ -0,0 +1,224 @@
...
+       }”
+       Spec:
+        fieldA: valueA
+        overrideFieldB: valueB
+       Status:
+        fieldC: valueC
+        conditions:
+        - type: ACK.ServerSideOverride
+          status: True
+          reason: ServerSideSpecOverride
```
RedbackThomson, line 107: Could probably just be shortened to `SpecOverride` - `ServerSide` is implied

docs/design/proposals/server-side-override/server-side-override.md
```diff
@@ -0,0 +1,224 @@
...
+          message: Spec override values are present in services.aws.k8s/spec-overrides annotation
```
RedbackThomson, line 108: Could this message include a list of overridden keys, saving users the hassle of looking up the annotations.

docs/design/proposals/server-side-override/server-side-override.md
```diff
@@ -0,0 +1,224 @@
...
+    for these fields. fieldA → “valueA“, overrideFieldB → “valueB-xyz“, fieldC → ”valueC“
+    
+    Based on this result, ACK service controller will persist the following object in etcd:
+    
+    Annotations:
+    
+    ```yaml
+    services.k8s.aws/spec-overrides: 
+       “{ 
+           \“overrideFieldB\”:
```
RedbackThomson, line 93: nit: formatting for this document - maybe add `Annotations:` like you have for `Spec:` and `Status:`

docs/design/proposals/server-side-override/server-side-override.md
```diff
@@ -0,0 +1,224 @@
...
+    “{
+        \“overrideFieldB\”:
+            {
+                \“userInput\”:\“updatedValueB\”,
+                \“serverOverride\”:\“updatedValueB-xyz\”
+            }
+    }”
+    Spec:
+     fieldA: valueA
+     overrideFieldB: updatedValueB-xyz
```
RedbackThomson, line 160: This should be `updatedValueB`, right? We don't patch over the user's Spec input?

RedbackThomson (commented):
Seems solid. Added some questions inline

docs/design/proposals/server-side-override/server-side-override.md
```diff
@@ -0,0 +1,224 @@
...
+## Solution Requirements
+
+1. Consistency with K8s conventions
+2. Consistency across ACK service controllers
+3. GitOps compatibility
+4. User experience for custom resource lifecycle 
+
+## Solution
+
+In this solution, If an existing value is updated as part of Create/Update API output, ACK service controller will persist
```
echen-98, line 30: I know this may be obvious to some, but it may be useful to make explicit why the annotation/mapping is only written during Creates/Updates, and not during Reads

PR #857, opened 2021-07-07
fix provision-kind-cluster.sh to use K8S_VERSION

Ashon:
The K8S_VERSION environment variable is being used to select the Kind node version,
which is currently not working as expected, and the code is modified
to allow the environment variable to properly specify the Kind node version.

Add kind node 1.21 version to choices


jaypipes (approved):
Thank you very much @Ashon! :)

PR #989, opened 2021-09-27
Keep installation docs up-to-date with latest s3-controller version

vijtrip2:
Description of changes:
* use latest s3-controller release for getting-started guide
* This will encourage keeping the guide up-to-date
* there will be less surprises for beginner users when they move from `getting-started` to using latest controllers for their workload
* this will provide the beginner users most up-to-date behavior of ACK controller runtime.

### DIFF
--- a/docs/content/docs/user-docs/install.md
+++ b/docs/content/docs/user-docs/install.md
@@ -3,7 +3,7 @@ title: "Installation"
 description: "Installing an ACK controller"
 lead: ""
 draft: false
-menu: 
+menu:
   docs:
     parent: "installing"
 weight: 10
@@ -12,7 +12,7 @@ toc: true
 
 The following guide will walk you through the installation of an [ACK service controller][ack-services].
 
-Individual ACK service controllers may be in different maintenance phases and follow separate release cadences. Please check the [project stages][proj-stages] and [maintenance phases][maint-phases] of the ACK service controllers you wish to install, including how controllers are [released and versioned][rel-ver]. Controllers in a preview maintenance phase have at least one container image and Helm chart released to a public repository. 
+Individual ACK service controllers may be in different maintenance phases and follow separate release cadences. Please check the [project stages][proj-stages] and [maintenance phases][maint-phases] of the ACK service controllers you wish to install, including how controllers are [released and versioned][rel-ver]. Controllers in a preview maintenance phase have at least one container image and Helm chart released to a public repository.
 
 {{% hint title="Be mindful of maintenance phases" %}}
 Check the [project stage](../../community/releases/#project-stages) and [maintenance phase](../../community/releases/#maintenance-phases) of the ACK service controller you wish to install. Be aware that controllers in a preview maintenance phase may have significant and breaking changes introduced in a future release.
@@ -47,7 +47,7 @@ Before installing a Helm chart, you must first make the Helm chart available on 
 ```bash
 export HELM_EXPERIMENTAL_OCI=1
 export SERVICE=s3
-export RELEASE_VERSION=v0.0.1
+export RELEASE_VERSION=$(git -c 'versionsort.suffix=-' ls-remote --tags --sort='v:refname' https://github.com/aws-controllers-k8s/s3-controller.git | tail -n 1 | cut -d'/' -f3)
 export CHART_EXPORT_PATH=/tmp/chart
 export CHART_REF=$SERVICE-chart
 export CHART_REPO=public.ecr.aws/aws-controllers-k8s/$CHART_REF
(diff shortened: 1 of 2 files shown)


docs/content/docs/user-docs/install.md
```diff
@@ -47,7 +47,7 @@ Before installing a Helm chart, you must first make the Helm chart available on
 ```bash
 export HELM_EXPERIMENTAL_OCI=1
 export SERVICE=s3
-export RELEASE_VERSION=v0.0.1
+export RELEASE_VERSION=$(git -c 'versionsort.suffix=-' ls-remote --tags --sort='v:refname' https://github.com/aws-controllers-k8s/s3-controller.git | tail -n 1 | cut -d'/' -f3)
```
jaypipes, line 50: I actually thought of proposing this in the original PR from @judyheflin :) However I decided against it because it would add `git` as a dependency for installation. I suppose a dependency on `git` isn't all that bad, but generally we're trying to keep the deps to a minimum and consider that installing folks may not be devs.

docs/content/docs/user-docs/install.md
```diff
@@ -59,19 +59,28 @@ helm pull oci://$CHART_REPO --version $RELEASE_VERSION -d $CHART_EXPORT_PATH
 tar xvf $CHART_EXPORT_PATH/$CHART_PACKAGE -C $CHART_EXPORT_PATH
 ```
 
+{{% hint type="info" title="s3-controller version" %}}
+Above commands will download the latest version of `s3-controller`. To select a
+different version, change `RELEASE_VERSION` variable and execute the commands again.
```
jaypipes, line 64: :+1: 

I actually think that putting the `git` command from above **here** in this hint would be a better idea. Basically, something like "if you have Git on your system, you can find the latest release tag using this Bash snippet..."

docs/content/docs/user-docs/install.md
```diff
@@ -59,19 +59,28 @@ helm pull oci://$CHART_REPO --version $RELEASE_VERSION -d $CHART_EXPORT_PATH
...
+    --set aws.account_id="$AWS_ACCOUNT_ID" \
+    --set aws.region="$AWS_REGION" \
     $CHART_EXPORT_PATH/$SERVICE-chart
 ```
 
 The `helm install` command should return relevant installation information:
 
 ```bash
-helm install --namespace $ACK_K8S_NAMESPACE ack-$SERVICE-controller $CHART_EXPORT_PATH/$SERVICE-chart
+helm install --namespace $ACK_K8S_NAMESPACE ack-$SERVICE-controller --set aws.account_id="$AWS_ACCOUNT_ID" --set aws.region="$AWS_REGION" $CHART_EXPORT_PATH/$SERVICE-chart
```
jaypipes, line 83: How about we split this long line across multiple lines like above on lines 74-78

jaypipes (changes requested):
Great stuff @vijtrip2. Just a couple small suggestions inline for you.

docs/content/docs/user-docs/install.md
RedbackThomson: I hope that we can do away with manually setting the account ID, seeing as Vijay has the PRs out for defaulting it. I'd prefer to wait for that and leave it out.

docs/content/docs/user-docs/install.md
```diff
@@ -99,9 +114,15 @@ helm list --namespace $ACK_K8S_NAMESPACE -o yaml
   updated: 2020-12-17 13:09:17.309002201 -0500 EST
 ```
 
+{{% hint type="important" title="NOTE" %}}
+s3-controller should be installed successfully now but it is not fully functional.
```
RedbackThomson, line 118:
```suggestion
The `s3-controller` should be installed successfully now, but it is not yet fully functional.
```

docs/content/docs/user-docs/install.md
RedbackThomson: To clarify why I'm against this in the documentation:
This method currently assumes that you have authenticated with the same account as the one you are using for ACK. It is possible that you have been giving access to an EKS cluster in another account, or don't have programmatic (CLI) access to the same account in which the cluster resides. In that case, `get-caller-identity` will either return the wrong value or no value. I personally don't want to debug that nightmare of a scenario.

docs/content/docs/user-docs/install.md
jaypipes: I suppose so, sure.

docs/content/docs/user-docs/install.md
vijtrip2: Interesting perspective on this @RedbackThomson , But solely relying on sts `GetCallerIdentity` to find AWS AccountId(which is right thing IMO) will require moving `irsa.md` and `authorization.md` doc before `install.md` in order. 
This is because if IAM is not configured for controller-pod, ack runtime will not be able to find AWS AccountId.

Is that something we are okay with?

RedbackThomson: If the runtime can't find an account ID, will the pod crash loop? If so, then it's fine to have authentication installation afterward - it will crash loop until IRSA (or otherwise) is configured properly. In fact this might be a good indication that the user has configured their authentication properly.

vijtrip2: Good point. But I will leave this as is right now because we have are seeing some [customer friction](https://github.com/aws-controllers-k8s/community/issues/990)

Once my changes are merged for runtime and code-gen and new controller is released for s3, I will come back and remove accountid option from here.

jaypipes (approved):
:+1: thanks @vijtrip2!

PR #1025, opened 2021-10-12
Support multiple managed and inline policies

RedbackThomson:
Fixes https://github.com/aws-controllers-k8s/community/issues/1024

Description of changes:
Updates the user installation documentation for creating the IAM role, for services with multiple managed policies or services that require inline policy statements.


RedbackThomson:
> JFYI: `test-recommended-policy` will need the change as well.. currently the test assumes that only single line will be present inside `recommended-policy-arn` file.

On it!

PR #1057, opened 2021-11-07
proposal for supporting resource references

vijtrip2:
Issue #545 

Description of changes:

* Proposal for handling resource relationship in ACK custom resources.
* View the rendered proposal [here](https://github.com/vijtrip2/community/blob/res-rel-proposal/docs/design/proposals/resource-reference/resource-reference.md)

### DIFF
--- a/docs/design/proposals/resource-reference/resource-reference.md
+++ b/docs/design/proposals/resource-reference/resource-reference.md
@@ -248,13 +248,13 @@ func NewReferenceField(
 
 type AWSResourceManager interface {
         ...
-+       // DereferenceResource finds if there are any AWSResourceReference fields
++       // ResolveReferences finds if there are any AWSResourceReference fields
 +       // present in the AWSResource passed in the parameters and attempts to resolve
 +       // those reference fields into target field.
 +       // It returns an AWSResource with resolved references, boolean representing
 +       // whether any reference fields were present and an error if the passed
 +       // AWSResource's reference fields cannot be resolved
-+       DereferenceResource(context.Context, client.Reader, AWSResource) (AWSResource, bool, error)
++       ResolveReferences(context.Context, client.Reader, AWSResource) (AWSResource, bool, error)
 }
 ```
 
@@ -273,7 +273,7 @@ res acktypes.AWSResource,
     if res.IsBeingDeleted() {
 +       // Resolve references before deleting the resource.
 +       // Ignore any errors while resolving the references
-+    	  res, _, _ = rm.DereferenceResource(ctx, r.apiReader, res)
++    	  res, _, _ = rm.ResolveReferences(ctx, r.apiReader, res)
         return r.deleteResource(ctx, rm, res)
     }
     return r.Sync(ctx, rm, res)
@@ -297,9 +297,9 @@ func (r *resourceReconciler) Sync(
 
 	isAdopted := IsAdopted(desired)
 	rlog.WithValues("is_adopted", isAdopted)
-+	rlog.Enter("rm.DereferenceResource")
-+	dereferencedDesired, containsResourceRef, err := rm.DereferenceResource(ctx, r.apiReader, desired)
-+	rlog.Exit("rm.DereferenceResource", err)
++	rlog.Enter("rm.ResolveReferences")
++	resolvedRefDesired, containsResourceRef, err := rm.ResolveReferences(ctx, r.apiReader, desired)
++	rlog.Exit("rm.ResolveReferences", err)
 +	if err != nil {
 +		// Set the condition in copy of desired resources so that it can be
 +		//patched back in etcd
@@ -315,13 +315,13 @@ func (r *resourceReconciler) Sync(
 +	// If the desired resource had references, set the ReferenceResolved
 +	//condition to true
 +	if containsResourceRef {
-+		condition.SetReferenceResolved(dereferencedDesired, corev1.ConditionTrue,
++		condition.SetReferenceResolved(resolvedRefDesired, corev1.ConditionTrue,
 +			&condition.ReferenceResolvedMessage,
 +			&condition.ReferenceResolvedReason)
 +	} else {
-+		condition.RemoveReferenceResolved(dereferencedDesired)
++		condition.RemoveReferenceResolved(resolvedRefDesired)
 +	}
-+	desired = dereferencedDesired
++	desired = resolvedRefDesired
 
 	rlog.Enter("rm.ReadOne")
 	latest, err = rm.ReadOne(ctx, desired)
@@ -410,13 +410,13 @@ var lateInitializeFieldNames = []string{}
 
 ...
 
-// DereferenceResource finds if there are any AWSResourceReference fields present
+// ResolveReferences finds if there are any AWSResourceReference fields present
 // in the AWSResource passed in parameter and attempts to resolve those reference
 // fields into target field.
 // It returns a resolved AWSResource, boolean representing whether any reference
 // fields were present and an error if the passed AWSResource's reference fields
 // cannot be resolved
-func (rm *resourceManager) DereferenceResource(
+func (rm *resourceManager) ResolveReferences(
 	ctx context.Context,
 	apiReader client.Reader,
 	res acktypes.AWSResource,
@@ -487,13 +487,13 @@ var lateInitializeFieldNames = []string{}
 
 ...
 
-// DereferenceResource finds if there are any AWSResourceReference fields present
+// ResolveReferences finds if there are any AWSResourceReference fields present
 // in the AWSResource passed in parameter and attempts to resolve those reference
 // fields into target field.
 // It returns a resolved AWSResource, boolean representing whether any reference
 // fields were present and an error if the passed AWSResource's reference fields
 // cannot be resolved
-func (rm *resourceManager) DereferenceResource(
+func (rm *resourceManager) ResolveReferences(
 	ctx context.Context,
 	apiReader client.Reader,
 	res acktypes.AWSResource,
--- a/docs/design/proposals/resource-reference/resource-reference.md
+++ b/docs/design/proposals/resource-reference/resource-reference.md
@@ -5,7 +5,7 @@ in another custom resource. The reconciler will wait for the referenced custom
 resource to be created, and in a stable state, before proceeding to create the
 resource which references it.
 
-For example, currently if ACK customer needs to create an APIGatewayv2 integration,
+For example, currently if ACK customer needs to create an APIGatewayv2 `Integration`,
 they first need to
 1. create an `API` custom resource
 2. wait for `API` resource creation to finish
@@ -80,27 +80,19 @@ of a field but also check if the Status is available etc...
 ```go
 type FieldConfig struct {
     ...
-    Reference *ReferenceConfig `json:"references,omitempty"`
+    Reference *ReferenceConfig `json:"reference,omitempty"`
     ...
 }
 
 // ReferenceConfig contains the instructions for how to add the referenced resource
 // configuration for a field.
 type ReferenceConfig struct {
-	// Kind mentions the K8s Kind of referenced resource
-	Kind string `json:"kind"`
-	// FieldPath refers to the the path of field which should be copied
-	// from referenced resource into TargetFieldName
-	FieldPath string `json:"field_path"`
-	// TargetFieldName is spec field name which gets value from
-	// referenced resource's FieldPath
-	TargetFieldName string `json:"target_field"`
-	// IsList mentions whether the new field should be AWSResourceReference or
-	// List of AWSResourceReference
-	IsList          bool   `json:"is_list"`
-	// Required mentions whether either of AWSResourceReference or
-	// TargetFieldName must be present in custom resource
-	Required        bool   `json:"required"`
+	// TargetResource mentions the K8s resource which is read to resolve the
+	// reference
+	TargetResource string `json:"target_resource"`
+	// SourcePath refers to the the path of field which should be copied
+	// to resolve the reference
+	SourcePath string `json:"source_path"`
 }
 ```
 
@@ -109,18 +101,11 @@ type ReferenceConfig struct {
 resources:
   Integration:
     fields:
-      Api:
-        references:
-          kind: API
-          field_path: Status.APIID
-          target_field: APIID
-          is_list: false
-          required: true
       ApiId:
-        is_required: false
+        reference:
+          target_resource: API
+          source_path: Status.APIID
 ```
-NOTE: `ApiId` field is marked as not required because the new reference field
-`Api` can also provide APIId.
 
 #### New AWSResourceReference Type
 
@@ -154,7 +139,7 @@ kind: Integration
 metadata:
   name: my-integration
 spec:
-  api:
+  apiIDRef:
     name: my-api
   integrationType: HTTP_PROXY
   integrationURI: https://example.org
@@ -193,6 +178,17 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
 		}
 		crdNames := names.New(crdName)
 ...
+			crd.AddSpecField(memberNames, memberShapeRef)
+
++			fConfig := m.cfg.ResourceFields(crdName)[fieldName]
++			if fConfig != nil && fConfig.Reference != nil {
++				referenceFieldNames := names.New(fieldName + "Ref")
++				rf := NewReferenceField(crd, referenceFieldNames, memberShapeRef)
++				crd.SpecFields[referenceFieldNames.Original] = rf
++				crd.Fields[referenceFieldNames.Camel] = rf
++			}
+		}
+
 		// Now any additional Spec fields that are required from other API
 		// operations.
 		for targetFieldName, fieldConfig := range m.cfg.ResourceFields(crdName) {
@@ -200,14 +196,6 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
 				// It's a Status field...
 				continue
 			}
-
-+			if fieldConfig.Reference != nil {
-+				// Add the spec field here
-+				referenceFieldNames := names.New(targetFieldName)
-+				rf := NewReferenceField(crd, referenceFieldNames, fieldConfig)
-+				crd.SpecFields[referenceFieldNames.Original] = rf
-+				crd.Fields[referenceFieldNames.Camel] = rf
-+			}
 
 ...
 	m.crds = crds
@@ -220,13 +208,15 @@ func (m *Model) GetCRDs() ([]*CRD, error) {
 func NewReferenceField(
 	crd *CRD,
 	fieldNames names.Names,
-	cfg *ackgenconfig.FieldConfig,
+	shapeRef *awssdkmodel.ShapeRef,
 ) *Field {
 	gt := "*ackv1alpha1.AWSResourceReference"
-	gtp := "*ackv1alpha1.AWSResourceReference"
-	if cfg.Reference.IsList {
+	gtp := "*github.com/aws-controllers-k8s/runtime/apis/core/v1alpha1.AWSResourceReference"
+	gte := ""
+	if shapeRef.Shape.Type == "list" {
 		gt = "[]" + gt
 		gtp = "[]" + gtp
+		gte = "*ackv1alpha1.AWSResourceReference"
 	}
 	return &Field{
 		CRD:               crd,
@@ -234,9 +224,9 @@ func NewReferenceField(
 		Path:              fieldNames.Original,
 		ShapeRef:          nil,
 		GoType:            gt,
-		GoTypeElem:        "AWSResourceReference",
+		GoTypeElem:        gte,
 		GoTypeWithPkgName: gtp,
-		FieldConfig:       cfg,
+		FieldConfig:       nil,
 	}
 }
 ```
@@ -361,24 +351,16 @@ referencing the resource from same api-group.
 resources:
   Integration:
     fields:
-      Api:
-        references:
-          kind: API
-          field_path: Status.APIID
-          target_field: APIID
-          is_list: false
-          required: true
       ApiId:
-        is_required: false
+        reference:
+          target_resource: API
+          source_path: Status.APIID
   VpcLink:
     fields:
-      SecurityGroups:
-        references:
-          kind: API
-          field_path: Status.APIID
-          target_field: SecurityGroupIDs
-          is_list: true
-          required: false
+      SecurityGroupIds:
+        reference:
+          target_resource: API
+          source_path: Status.APIID
 ```
 
 **integration/manager.go**
@@ -423,16 +405,16 @@ func (rm *resourceManager) ResolveReferences(
 ) (acktypes.AWSResource, bool, error) {
 	ko := rm.concreteResource(res).ko.DeepCopy()
 	referencePresent := false
-	if ko.Spec.API != nil && ko.Spec.APIID != nil {
-		return &resource{ko}, true, fmt.Errorf("'APIID' field should not be present when using reference field 'API'")
+	if ko.Spec.APIIDRef != nil && ko.Spec.APIID != nil {
+		return &resource{ko}, true, fmt.Errorf("'APIID' field should not be present when using reference field 'APIIDRef'")
 	}
-	if ko.Spec.API == nil && ko.Spec.APIID == nil {
-		return &resource{ko}, referencePresent, fmt.Errorf("At least one of 'APIID' or 'API' field should be present")
+	if ko.Spec.APIID == nil && ko.Spec.APIIDRef == nil {
+		return &resource{ko}, referencePresent, fmt.Errorf("At least one of 'APIID' or 'APIIDRef' field should be present")
 	}
-	// Checking Referenced Field API
-	if ko.Spec.API != nil {
+	// Checking Referenced Field APIIDRef
+	if ko.Spec.APIIDRef != nil {
 		referencePresent = true
-		arr := ko.Spec.API
+		arr := ko.Spec.APIIDRef
 		namespacedName := types.NamespacedName{Namespace: res.MetaObject().GetNamespace(), Name: *arr.Name}
 		obj := acksvcv1alpha1.API{}
 		err := apiReader.Get(ctx, namespacedName, &obj)
@@ -456,6 +438,7 @@ func (rm *resourceManager) ResolveReferences(
 	}
 	return &resource{ko}, referencePresent, nil
 }
+
 ```
 
 **vpc_link/manager.go**
@@ -500,14 +483,14 @@ func (rm *resourceManager) ResolveReferences(
 ) (acktypes.AWSResource, bool, error) {
 	ko := rm.concreteResource(res).ko.DeepCopy()
 	referencePresent := false
-	if ko.Spec.SecurityGroups != nil && ko.Spec.SecurityGroupIDs != nil {
-		return &resource{ko}, true, fmt.Errorf("'SecurityGroupIDs' field should not be present when using reference field 'SecurityGroups'")
+	if ko.Spec.SecurityGroupIDsRef != nil && ko.Spec.SecurityGroupIDs != nil {
+		return &resource{ko}, true, fmt.Errorf("'SecurityGroupIDs' field should not be present when using reference field 'SecurityGroupIDsRef'")
 	}
-	// Checking Referenced Field SecurityGroups
-	if ko.Spec.SecurityGroups != nil && len(ko.Spec.SecurityGroups) > 0 {
+	// Checking Referenced Field SecurityGroupIDsRef
+	if ko.Spec.SecurityGroupIDsRef != nil && len(ko.Spec.SecurityGroupIDsRef) > 0 {
 		referencePresent = true
 		resolvedReferences := []*string{}
-		for _, arr := range ko.Spec.SecurityGroups {
+		for _, arr := range ko.Spec.SecurityGroupIDsRef {
 			namespacedName := types.NamespacedName{Namespace: res.MetaObject().GetNamespace(), Name: *arr.Name}
 			obj := acksvcv1alpha1.API{}
 			err := apiReader.Get(ctx, namespacedName, &obj)
--- a/docs/design/proposals/resource-reference/resource-reference.md
+++ b/docs/design/proposals/resource-reference/resource-reference.md
@@ -1,0 +1,546 @@
+## Introduction
+
+This proposal discusses adding the support for referencing an ACK custom resource
+in another custom resource. The reconciler will wait for the referenced custom
+resource to be created, and in a stable state, before proceeding to create the
+resource which references it.
+
+For example, currently if ACK customer needs to create an APIGatewayv2 integration,
+they first need to
+1. create an `API` custom resource
+2. wait for `API` resource creation to finish
+3. manually find the `APIID` from `Status` of `API` resource
+4. create an `Integration` resource with the `APIID`
+
+Having to manually wait for the resource creation prevents declarative configuration
+techniques that rely on describing all desired resources at the same time using
+one or more Kubernetes YAML manifests. But with approach of referenced resources,
+ACK controller will wait for `API` resource to get created and then use the API
+reference to find `APIID` and create Integration resource.
+
+The deletion strategy for the referenced resource is out of scope for this proposal
+right now. For Ex: If API resource is referenced in integration resource, the deletion
+of API resource will not be blocked. We plan to handle deletion strategy in future.
+
+## Scope
+
+### First Revision
+* Allow adding resource name(s) from same api-group as reference in the custom
+resource definition
+* Only the resources from same namespace will be allowed to use as reference
+* Only the fields of type '*string' or '[]*string' can be referenced from
+another resource. This satisfies the use case of referencing 'ID(s)/Name(s)/ARN(s)'
+from other resources
+
+### Future Improvements
+* Add support for custom hooks while dereferencing resources
+* Add support for `selector` for referencing the objects
+* Add support for referencing resource from different api-groups
+* Add support for not deleting a resource if it is referenced in another resource
+* Add support for finite retries and terminal conditions when resolving references
+
+## Preferred Approach
+
+### High Level Overview
+
+* Introduce new construct in generator.yaml which will indicate to code-generator
+if a new spec field with type `AWSResourceReference` should be added to the custom
+resource definition.
+
+* This new spec field will work as additional input for an existing spec field.
+Addition of the extra spec field keeps the backward compatibility with existing
+custom resource definitions.
+
+* Validation will be present that only one of existing spec field OR new reference
+field is present in custom resource, not both.
+
+* If the existing spec field was marked as required, validation will be added that
+at least one of existing spec field or new reference field is present.
+
+* If `AWSResourceReference` type field is present in the desired state, first ACK
+controller will query for the referenced resource. If the referenced resource
+has `ACK.ResourceSynced` condition status `True` & specified field is found
+in the referenced resource, the reconciliation loop will progress. Otherwise
+the request will get requeued for future processing.
+
+* Watchers for the referenced resource type will not be implemented in the initial
+implementation. The initial implementation will mostly be used for finding
+identifier of referenced resource and any update in referenced resource should not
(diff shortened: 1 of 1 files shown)


docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+	Kind string `json:"kind"`
+	// FieldPath refers to the the path of field which should be copied
+	// from referenced resource into TargetFieldName
+	FieldPath string `json:"field_path"`
+	// TargetFieldName is spec field name which gets value from
+	// referenced resource's FieldPath
+	TargetFieldName string `json:"target_field"`
+	// IsList mentions whether the new field should be AWSResourceReference or
+	// List of AWSResourceReference
+	IsList          bool   `json:"is_list"`
```
A-Hilaly, line 100: Can't it just be always a list? i think it's way easier to deal with one type.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+```
+NOTE: `ApiId` field is marked as not required because the new reference field
+`Api` can also provide ApiId.
+
+#### New AWSResourceReference Type
+
+```go
+type AWSResourceReference struct {
+	Name *string `json:"name,omitempty"`
+}
```
A-Hilaly, line 130: Maybe we should make this more an interface, and have something like:
```go
type AWSResourceReference interface {
	Name() string
        ...
}
```

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+
+b) Using the API resource reference
+```yaml
+apiVersion: apigatewayv2.services.k8s.aws/v1alpha1
+kind: Integration
+metadata:
+  name: my-integration
+spec:
+  api:
+    name: my-api
```
A-Hilaly, line 158: Does this proposal assume that the resources should be in the same namespace? will cross namespace reference be supported?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+
+* Introduce new construct in generator.yaml which will indicate to code-generator
+if a new spec field with type `AWSResourceReference` should be added to the custom
+resource definition.
+
+* This new spec field will work as additional input for an existing spec field.
+Addition of the extra spec field keeps the backward compatibility with existing
+custom resource definitions.
+
+* Validation will be present that only one of existing spec field OR new reference
```
A-Hilaly, line 55: Not sure if i understand this statement. Why can't we have spec fields and reference at same time?

A-Hilaly (commented):
Great start on this @vijtrip2 ! i left few comments and questions to understand more the design aspects - didn't review the implementation part yet.

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: What benefits will that provide?

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Yes, that is correct. There is no plan to support cross namespace references. 
TL;DR: In K8s namespaces are security boundaries and we should not be referring resources across namespace. There is more discussion on the same topic in PR #856

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: For ex: Why would one want to provide an `ApiId` and API reference at the same time?
If we allow this, it will make us choose which one to prioritize etc. I thought it's better to not go down that road.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+// configuration for a field.
+type ReferenceConfig struct {
+	// Kind mentions the K8s Kind of referenced resource
+	Kind string `json:"kind"`
+	// FieldPath refers to the the path of field which should be copied
+	// from referenced resource into TargetFieldName
+	FieldPath string `json:"field_path"`
+	// TargetFieldName is spec field name which gets value from
+	// referenced resource's FieldPath
+	TargetFieldName string `json:"target_field"`
```
RedbackThomson, line 97: I find these field names pretty difficult to understand since they're fairly similar. Perhaps we could go with something more assertive rather than descriptive. For example, `take_path` and `override_path`? `take_path` would be the field on the target resource that we are taking and `override_path` is the spec field on the current resource to override?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+	// IsList mentions whether the new field should be AWSResourceReference or
+	// List of AWSResourceReference
+	IsList          bool   `json:"is_list"`
+	// Required mentions whether either of AWSResourceReference or
+	// TargetFieldName must be present in custom resource
+	Required        bool   `json:"required"`
```
RedbackThomson, line 103: Can be determined dynamically? Any field currently required in the spec would continue to be required if referenced as well?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+    fields:
+      Api:
+        references:
+          kind: API
+          field_path: Status.APIID
+          target_field: APIID
+          is_list: false
+          required: true
+      ApiId:
+        is_required: false
```
RedbackThomson, line 120: I would love for this to be determine dynamically as well - we know it's being overridden already, since it's referenced in `target_field` above.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+* The code to determine when the referenced resource is synced will be generated,
+and additional hooks will be added for customization. Ex: Do not just check presence
+of a field but also check if the Status is available etc...
+
+### Low Level Implementation
+
+#### generator.yaml Changes
+
+```go
+type FieldConfig struct {
```
RedbackThomson, line 81: I don't agree with this being a property of a field. I want to remove the need for service teams to choose names for the reference fields - in case every team chooses a different naming convention. I would greatly prefer that we append `Ref` to the end of every field we hope to override so that's it's absolutely clear to users that it's an overriding reference to another field.

eg. `apiID` becomes `apiIDRef`

RedbackThomson: This would mean that we have a list of references defined per resource.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+// It returns a copy of the resource that represents the latest observed state.
+func (r *resourceReconciler) reconcile(
+ctx context.Context,
+rm acktypes.AWSResourceManager,
+res acktypes.AWSResource,
+) (acktypes.AWSResource, error) {
+    if res.IsBeingDeleted() {
++       // Resolve references before deleting the resource.
++       // Ignore any errors while resolving the references
++    	  res, _, _ = rm.DereferenceResource(ctx, r.apiReader, res)
```
RedbackThomson, line 276: I think we should ignore deletion in this proposal as it's a whole other can of worms to get into.

RedbackThomson (changes requested):
I'm super happy with the overall formula for this proposal. It's a shame we have to create a separate reference field, rather than overriding the existing ones, but I understand the limitation.

I have some concerns regarding the format of the generator - I would like to reduce complexity for the service teams as much as possible and I think we can simplify it greatly. See the inline comments for more details.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
+## Introduction
+
+This proposal discusses adding the support for referencing an ACK custom resource
+in another custom resource. The reconciler will wait for the referenced resource
+to get created first and then create the custom resource.
```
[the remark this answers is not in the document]
vijtrip2: Good point. Having the check that referenced resource is stable(Condition:Synced = True) should be taken into consideration. 
I will add this.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+
+The deletion strategy for the referenced resource is out of scope for this proposal
+right now. For Ex: If API resource is referenced in integration resource, the deletion
+of API resouorce will not be blocked. We plan to handle deletion strategy in next
+revision.
+
+## Scope
+
+### First Revision
+* Allow referencing resource from same api group in custom resource definitions
```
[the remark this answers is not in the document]
vijtrip2: > Referencing only by name/namespace pair?

Yes

> Referencing only primitive types - ID/name strings, ARNs, etc.

Yes, I will add this detail.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+
+* Validation will be present that only one of existing spec field OR new reference
+field is present in custom resource, not both.
+
+* If the existing spec field was marked as required, validation will be added that
+atleast one of existing spec field or new reference field is present.
+
+* If `AWSResourceReference` type field is present in desired state, first ACK
+runtime will query for the referenced resource and only when the specified field
+is found in referenced resource, the reconciliation loop will progress. Otherwise
```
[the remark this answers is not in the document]
vijtrip2: Yes, I missed the detail. 
I will update that we check for ResourceSynced condition while dereferencing a resource

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Yep, but we do need the referenced field in order to construct `DeleteInput`

RedbackThomson: Should we be storing the dereferenced value in some way after we create the resource? If the referenced resource changes after we create the resource that references it, we lose context about what the original Create call contained?

vijtrip2: > Should we be storing the dereferenced value in some way after we create the resource?

I don't think so. I did not persist the dereferenced value back into spec either.

> If the referenced resource changes after we create the resource that references it, we lose context about what the original Create call contained?

Currently, we are only using referenced resources to find identifiers which do not change when referenced resource is updated. This was the basis on which I skipped implementing watchers for referenced resource.

So, ATM i don't think we need to preserve anything from dereferenced field. It is also NOT a one way door decision.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+
+* This new spec field will work as additional input for an existing spec field.
+Addition of the extra spec field keeps the backward compatibility with existing
+custom resource definitions.
+
+* Validation will be present that only one of existing spec field OR new reference
+field is present in custom resource, not both.
+
+* If the existing spec field was marked as required, validation will be added that
+at least one of existing spec field or new reference field is present.
```
jaypipes, line 63: :+1:

docs/design/proposals/resource-reference/resource-reference.md
jaypipes: What about `source_path` and `target_path`? Or `from_path` and `to_path`?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,526 @@
...
+    ...
+    Reference *ReferenceConfig `json:"references,omitempty"`
+    ...
+}
+
+// ReferenceConfig contains the instructions for how to add the referenced resource
+// configuration for a field.
+type ReferenceConfig struct {
+	// Kind mentions the K8s Kind of referenced resource
+	Kind string `json:"kind"`
```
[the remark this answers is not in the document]
jaypipes: Agree that using "resource" would be better. Either `target_resource` or `from_resource` would be good, IMHO.

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+```go
+type FieldConfig struct {
+    ...
+    Reference *ReferenceConfig `json:"references,omitempty"`
+    ...
+}
+
+// ReferenceConfig contains the instructions for how to add the referenced resource
+// configuration for a field.
+type ReferenceConfig struct {
```
jaypipes, line 94: Can we either call this `ReferencesConfig` or change the `json:"references,omitempty"` above to `json:"reference,omitempty"`? I would like consistency in how this is named.

docs/design/proposals/resource-reference/resource-reference.md
jaypipes: I believe we can continue to have `ReferenceConfig` be part of `FieldConfig`, but can avoid your naming issues if we encourage the original field (e.g., `apiID`) to get the ReferenceConfig and not a new field. In other words, if this were the generator.yaml file for API Gateway v2:

```yaml
resources:
  Integration:
    fields:
      APIID:
        reference:
          from_resource: API
          from_path: Status.APIID
          to_field: APIID
```

then that would indicate that the `Integration.Spec.APIID` field could be populated with the `API.Status.APIID` field value of a resource referred to in a `Integration.Spec.APIIDRef` field, which we would automatically add to the `API` custom resource's schema when we found a `reference`. That way, we don't give anyone the ability to name their reference field something other than `<original field name>Ref`.

docs/design/proposals/resource-reference/resource-reference.md
jaypipes: Yes, we should be able to always programmatically determine if the target/to field is a list type or non-list type.

docs/design/proposals/resource-reference/resource-reference.md
jaypipes: Not necessarily, @RedbackThomson, because the original field in the Spec would need to have its kubebuilder required annotation removed and replaced with optional...

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+
+```go
+//field.go
+func NewReferenceField(
+	crd *CRD,
+	fieldNames names.Names,
+	cfg *ackgenconfig.FieldConfig,
+) *Field {
+	gt := "*ackv1alpha1.AWSResourceReference"
+	gtp := "*ackv1alpha1.AWSResourceReference"
```
jaypipes, line 231: We *may* want to make this configurable, since Crossplane has its own referencer type:

https://github.com/crossplane/crossplane-runtime/blob/5141d0bb35f4a42d23366d49b2a1d9fed9806c6c/apis/common/v1/resource.go#L89

/cc @muvaf

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+res acktypes.AWSResource,
+) (acktypes.AWSResource, error) {
+    if res.IsBeingDeleted() {
++       // Resolve references before deleting the resource.
++       // Ignore any errors while resolving the references
++    	  res, _, _ = rm.DereferenceResource(ctx, r.apiReader, res)
+        return r.deleteResource(ctx, rm, res)
+    }
+    return r.Sync(ctx, rm, res)
+}
```
jaypipes, line 285: We can remove this from the proposal since it deals with something explicitly out-of-scope for the proposal (deletion propogation)

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+	rlog := ackrtlog.FromContext(ctx)
+	exit := rlog.Trace("r.Sync")
+	defer exit(err)
+
+	var latest acktypes.AWSResource // the newly created or mutated resource
+
+	isAdopted := IsAdopted(desired)
+	rlog.WithValues("is_adopted", isAdopted)
++	rlog.Enter("rm.DereferenceResource")
++	dereferencedDesired, containsResourceRef, err := rm.DereferenceResource(ctx, r.apiReader, desired)
```
jaypipes, line 306: Pretty sure you don't mean `DereferenceResource` here, but rather `ResolveReferences` or something like that?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+	ackcompare "github.com/aws-controllers-k8s/runtime/pkg/compare"
+...
+	svcsdk "github.com/aws/aws-sdk-go/service/apigatewayv2"
+	svcsdkapi "github.com/aws/aws-sdk-go/service/apigatewayv2/apigatewayv2iface"
++	"k8s.io/apimachinery/pkg/types"
++	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=integrations,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=apigatewayv2.services.k8s.aws,resources=integrations/status,verbs=get;update;patch
```
jaypipes, line 407: this would be required once we support cross-controller/cross-apigroup references, but this is not necessary given that a controller automatically has read permissions for all resources it manages.

jaypipes (changes requested):
great start @vijtrip2! Some comments and suggestions inline for you!

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+
+### High Level Overview
+
+* Introduce new construct in generator.yaml which will indicate to code-generator
+if a new spec field with type `AWSResourceReference` should be added to the custom
+resource definition.
+
+* This new spec field will work as additional input for an existing spec field.
+Addition of the extra spec field keeps the backward compatibility with existing
+custom resource definitions.
```
brycahta, line 57: > Addition of the extra spec field keeps the backward compatibility with existing custom resource definitions

Could you clarify to help me understand how it wouldn't maintain backwards compat when you don't add an extra field? if no reference is provided, then can't we default to current behavior?

It seems weird to me that we need a new field when I just want my resource's existing field substituted with this referenced value.

docs/design/proposals/resource-reference/resource-reference.md
brycahta: > I don't agree with this being a property of a field.

+1, in addition to reasons above I think it also muddies the interface and use of `custom_field`. If you were to keep it in FieldConfig, then it'd be clearer to extend custom_field to have a `ref_of` type (list_of, map_of, ref_of) imo

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+	// TargetFieldName must be present in custom resource
+	Required        bool   `json:"required"`
+}
+```
+
+`APIGatewayv2 generator.yaml` Example:
+```yaml
+resources:
+  Integration:
+    fields:
```
brycahta, line 116: This may be outside the scope of this proposal, but any thoughts in making `fields` more explicit such as `spec_fields` ?


The mixing of these different fields is getting confusing. This override requires knowledge of CRDs Spec/Status field paths already, so I think it'd be clearer to do something like:

```
resources:
  Integration:
    spec_fields:
      APIID:
        references:
          kind: API
          value_from_path: Status.APIID
          is_list: false
          required: true
```

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+
+type AWSResourceManager interface {
+        ...
++       // DereferenceResource finds if there are any AWSResourceReference fields
++       // present in the AWSResource passed in the parameters and attempts to resolve
++       // those reference fields into target field.
++       // It returns an AWSResource with resolved references, boolean representing
++       // whether any reference fields were present and an error if the passed
++       // AWSResource's reference fields cannot be resolved
++       DereferenceResource(context.Context, client.Reader, AWSResource) (AWSResource, bool, error)
```
brycahta, line 262: Name is a bit confusing; I thought this was removing the reference. Something clearer would be `ResolveReference` , `ResolveReferenceFields`

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,551 @@
...
+	return &resource{ko}, referencePresent, nil
+}
+```
+
+## Alternate Approach
+* Change the type of member from `*string` to `*AWSResourceReference` and
+  `[]*string` to `[]*AWSResourceReference`
+* `AWSResourceReference` will either have the value for member field(string) or
+  contains the reference for another resource.
+* This approach was not chosen because it breaks the backward compatibility for
```
brycahta, line 548: could we maintain backward compat by enforcing the resource type and reference types are the same?

brycahta (commented):
Nice proposal; love the feature

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: This is needed to keep the existing deletion functionality. When reference is used, resolving the reference is required prior to deletion.
Ex: We need APIID to construct a DeleteIntegrationInput.

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Can you please elaborate? 
(The resource member type(*string/[]*string) cannot contain all information needed for referencing a resource.)

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: > Could you clarify to help me understand how it wouldn't maintain backwards compat when you don't add an extra field?

Because then we will be changing the type of existing field to accommodate passing another resource's reference.

> if no reference is provided, then can't we default to current behavior?

This will be the behavior for preferred approach

> It seems weird to me that we need a new field when I just want my resource's existing field substituted with this referenced value.

We would need a way to pass the reference in resource spec. The new field is precisely for that.

docs/design/proposals/resource-reference/resource-reference.md
brycahta: Your other answers cleared this up for me -- misunderstanding on my end.

docs/design/proposals/resource-reference/resource-reference.md
brycahta: I like `source_field` and `destination_field` both of which can support fully qualified names (paths)

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,546 @@
...
+        is_required: false
+```
+NOTE: `ApiId` field is marked as not required because the new reference field
+`Api` can also provide APIId.
+
+#### New AWSResourceReference Type
+
+```go
+type AWSResourceReference struct {
+	Name *string `json:"name,omitempty"`
```
jaypipes, line 129: We could also name this field `From` so that YAML would look like:

```yaml
spec:
  apiIDRef:
    from: my-api-name
```

jaypipes (changes requested):
I still believe the is_list attribute can be removed...

docs/design/proposals/resource-reference/resource-reference.md
RedbackThomson: Looking at our discussion notes from the other week, I thought we decided to go with:
```yaml
spec:
  apiIDRef:
    from:
      name: my-api-name
```

Leaving the door open for a `from.labelSelector` or otherwise?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -80,27 +80,19 @@ of a field but also check if the Status is available etc...
...
-	IsList          bool   `json:"is_list"`
-	// Required mentions whether either of AWSResourceReference or
-	// TargetFieldName must be present in custom resource
-	Required        bool   `json:"required"`
+	// TargetResource mentions the K8s resource which is read to resolve the
+	// reference
+	TargetResource string `json:"target_resource"`
+	// SourcePath refers to the the path of field which should be copied
+	// to resolve the reference
+	SourcePath string `json:"source_path"`
```
jaypipes, line 95: This should be SourceResource and SourcePath, no?

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: @jaypipes , will using `from` instead of `name` create confusion when in future we add labelSelector. The example that @RedbackThomson provided seems like better use of 'from' keyword while generating the manifest. wdyt?

--------------

To implement `from` keyword, how does the following look? (Note the new struct we will need to add here.)

```go
type IntegrationSpec struct {
	APIID    *string                           `json:"apiID,omitempty"`
	APIIDRef *ackv1alpha1.ReferenceProvider `json:"apiIDRef,omitempty"`
...
}

type ReferenceProvider struct {
        ResourceReference *AWSResourceReference `json:"from,omitempty"`
}

type AWSResourceReference struct {
	Name *string `json:"name,omitempty"`
}
```

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Yes, you are right. Correcting now.

docs/design/proposals/resource-reference/resource-reference.md
jaypipes: @vijtrip2 yeah, agreed. @RedbackThomson's example shows a better use of From.

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Do you think we can drop the 'Source' from name?
The generator yaml will look like

```yaml
resource:
  Integration:
    fields:
      ApiId:
        references:
          resource: API
          path: Status.APIID
```

jaypipes: > Do you think we can drop the 'Source' from name? @jaypipes The generator yaml will look like
> 
> ```yaml
> resources:
>   Integration:
>     fields:
>       ApiId:
>         references:
>           resource: API
>           path: Status.APIID
> ```

@vijtrip2 yes, absolutely

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: As per Jay's guidance, I am keeping the config inside fieldConfig but I have added another item for future to support references in custom fields

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: @jaypipes , can you please elaborate on this?

I am assuming that you are pointing towards taking the reference type as input(option/env variable) from user who performs `make build-controller`? The default will be `ackv1alpha1.AWSResourceReference` . 
This will allow crossplane to generate CRDs compatible with their reference type.

RedbackThomson: I believe he's referring to the fact that crossplane has a separate set of config defaults that are used exclusively when running the `ackgenerate crossplane` verb - https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/crossplane/config.go#L19

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,555 @@
...
+	Path string `json:"path"`
+}
+```
+
+`APIGatewayv2 generator.yaml` Example:
+```yaml
+resources:
+  Integration:
+    fields:
+      ApiId:
```
RedbackThomson, line 105: Should this be `APIID`? The name of the spec field?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,555 @@
...
+
+type AWSResourceManager interface {
+        ...
++       // ResolveReferences finds if there are any AWSResourceReference fields
++       // present in the AWSResource passed in the parameters and attempts to resolve
++       // those reference fields into target field.
++       // It returns an AWSResource with resolved references, boolean representing
++       // whether any reference fields were present and an error if the passed
++       // AWSResource's reference fields cannot be resolved
++       ResolveReferences(context.Context, client.Reader, AWSResource) (AWSResource, bool, error)
```
RedbackThomson, line 273: nit: I think the `bool` return value is unnecessary here because returning a nil reference (changing the first return value to a pointer) would suffice.

RedbackThomson (approved):
Looks essentially ready to me. I had 2 small nits.

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: The code inside `crd.go` works with "Original" field names not "Camel". Hence generator.yaml has field names as their original name in aws-sdk-go. To change this, We'll need to update logic inside `crd.go`

```go
func (r *CRD) AddSpecField(
	memberNames names.Names,
	shapeRef *awssdkmodel.ShapeRef,
) {
	fPath := memberNames.Camel
	fConfigs := r.cfg.ResourceFields(r.Names.Original)
	fConfig := fConfigs[memberNames.Original]
	f := NewField(r, fPath, memberNames, shapeRef, fConfig)
	if fConfig != nil && fConfig.Print != nil {
		r.addSpecPrintableColumn(f)
	}
	r.SpecFields[memberNames.Original] = f
	r.Fields[fPath] = f
}```

File: docs/design/proposals/resource-reference/resource-reference.md (hunk 27 above)
Comment (vijtrip2) [reply docs/design/proposals/resource-reference/resource-reference.md:273]: Good suggestion, but we will be introducing pointer to an interface. We have not used pointer to interface before in runtime and I am not sure If it's worth it to remove a return type. 

I'll see what Jay or Amine think about it.

File: docs/design/proposals/resource-reference/resource-reference.md [hunk 28]
```diff
@@ -0,0 +1,555 @@
...
+* Add support for not deleting a resource if it is referenced in another resource
+* Add support for finite retries and terminal conditions when resolving references
+
+## Preferred Approach
+
+### High Level Overview
+
+* Introduce new construct in generator.yaml which will indicate to code-generator
+if a new spec field with type `AWSResourceReferenceWrapper` or
+`AWSResourcesReferenceWrapper` should be added to the custom resource definition.
```
jaypipes, line 49: I don't think we'd need a `AWSResourcesReferenceWrapper` struct. This would just be `[]*AWSResourceReferenceWrapper`, no?

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,555 @@
...
+// AWSResourcesReferenceWrapper provides a wrapper around slice of *AWSResourceReference
+// type to provide more user friendly syntax for references using 'from' field
+// Ex:
+// SecurityGroupIDs:
+//   from:
+//     - name: my-sg-1
+//     - name: my-sg-2
+type AWSResourcesReferenceWrapper struct {
+	From []*AWSResourceReference `json:"from,omitempty"`
+}
```
jaypipes, line 133: I would have expected slice of identifier fields to use `[]*AWSResourceReferenceWrapper`:

```yaml
SecurityGroupIDs:
  - from:
      name: my-sg-1
  - from:
      name: my-sg-2
```

docs/design/proposals/resource-reference/resource-reference.md
brycahta: +1 to removing bool, interface will be simpler and more consistent with existing runtime methods like `updateResource` and `reconcile`

jaypipes: What about making a new `AWSResourceResolver` interface, inheriting from `AWSResource` and `ConditionManager` and just returning `error`?

```go
type AWSResourceResolver interface {
       AWSResource
       ConditionManager
       // ResolveReferences finds if there are any AWSResourceReference fields
       // present in the AWSResource and returns an error if the
       // reference fields cannot be resolved
       ResolveReferences(context.Context, client.Reader) error
}
```

There's no reason to have the copies of desired/latest, IMHO. Just have the `AWSResourceResolver.ResolveReferences` mutate the resource as needed to resolve the referenced resource identifiers and add the appropriate `Condition` to the resource itself (which it can do because it inherits from `ConditionManager`.

Within the reconciler's `Sync` method, you would simplify the code do just this:

```go
	if err := desired.ResolveReferences(ctx, r.apiReader); err != nil {
		return desired, err
	}
```

docs/design/proposals/resource-reference/resource-reference.md
```diff
@@ -0,0 +1,555 @@
...
+
++	// If the desired resource had references, set the ReferenceResolved
++	//condition to true
++	if containsResourceRef {
++		condition.SetReferenceResolved(resolvedRefDesired, corev1.ConditionTrue,
++			&condition.ReferenceResolvedMessage,
++			&condition.ReferenceResolvedReason)
++	} else {
++		condition.RemoveReferenceResolved(resolvedRefDesired)
++	}
```
jaypipes, line 339: do all of this condition-setting within the `AWSResourceResolver.ResolveReferences` method implementation...

jaypipes (changes requested):
I think we've got a layering problem. See inline for a suggestion in how to rework the interface abstraction.

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: I was hoping to avoid the many 'from' keyword, But above works as well.
It can be done either way i.e. `[]*AWSResourceReference` or `[]*AWSResourceReferenceWrapper`

docs/design/proposals/resource-reference/resource-reference.md
vijtrip2: Done. 
I changed the return type of `ResolveReferences` method to `(AWSResource, error)` and updated the conditions within that method.

@jaypipes , your suggestion was very good but when I started to implement it, I found that I will need to change type of `desired` to new interface to be able to call `desired.ResolveReferences` method.
**More importantly**, if we change the desired(by resolving references), the `PatchStatus` method will not see the new conditions as diff between `desired` & `latest` and these condition will not be persisted in K8s cluster.

Since we need a new `AWSResource` after resolving references, I sticked to existing `AWSResourceManager.ResolveReferences()` method and was able to cleanly update conditions within the method. 

PTAL.

docs/design/proposals/resource-reference/resource-reference.md
RedbackThomson: I agree that the wrapper structure is ugly, but from the the YAML output I prefer the single `from` keyword that @vijtrip2 is assuming. `from` is there to signify that it's a reference, rather than a value. Multiple `from` doesn't add any extra value to that IMO.

jaypipes: @RedbackThomson the problem with using the wrapper like that is the plural property of the source field is lost. In other words, the fact that the field stores *multiple* identifiers or a single identifier is lost from the schema of the resource itself.

RedbackThomson: Ah totally valid, then. You're using the `from` to identify a single reference. Thanks for the explanation.

vijtrip2:
@jaypipes , do you think this is good enough to be merged now?

jaypipes (approved):
Yeah, I'm good with this proposal as-is. I have some nits about code cleanup but will add those as review comments in the code PR. :)

Leweyy:
Will this support getting fields from a service in the future? i.e getting a loadbalancer dns name to be used in an intergration as a url?

vijtrip2:
> Will this support getting fields from a service in the future? i.e getting a loadbalancer dns name to be used in an intergration as a url?

If you mean will the controller make AWS ELB `Describe` API call, and find LB dns name then No.
However we will extend the support for different api-groups, which will enable you to create a Load balancer using ACK ELB controller and then refer that resource to find LB dns name.

Leweyy:
Not quite, we currently use the k8s service to create load balancers and it would be nice if we could reference that service in an intergration to get the dns name. Other wise it’s a copy and paste job

vijtrip2:
@Leweyy, Can you please create a GH issue for the same so that in future ACK team can circle back on implementing such functionality.

PR #1083, opened 2021-12-01
Fix example crossplane generation cmd

clareliguori:
The crossplane cmd does not appear to have an "apis" sub-command (anymore?)


jaypipes (approved):
I think this was just a latent copy/pasta mistake. Thank you @clareliguori! :)

PR #1100, opened 2021-12-14
code-generator refactor proposals

brycahta:
Issue #, if available: N/A

Description of changes:
* A summary doc, `proposal-summary.md` introducing 3 detailed proposals. Rendered markdown of [proposal-summary](https://github.com/brycahta/ack-community/blob/refactor-codegen/docs/design/proposals/codegen-refactor/proposal-summary.md).
* Although these proposals do not necessary depend on one another it is *recommended* to read in the following order (rendered markdowns): 
   * [`ackgenconfig` Categories](https://github.com/brycahta/ack-community/blob/refactor-codegen/docs/design/proposals/codegen-refactor/inference.md)
   * [Generator Enhancements](https://github.com/brycahta/ack-community/blob/refactor-codegen/docs/design/proposals/codegen-refactor/generator.md)
   * [`model` Command](https://github.com/brycahta/ack-community/blob/refactor-codegen/docs/design/proposals/codegen-refactor/infer_model_cmd.md)
 

vijtrip2 (commented):
Great work and inventive ideas! I like the overall proposal and in agreement with it.

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Test plan
+* add "sufficient" testing beforehand for quick feedback
+* update existing unit tests to reflect any changed logic due to migration
+* run a "healthy" subset of service controller e2e tests for validation
+
+## Appendix
+
+### Long paramaters
+The majority of functions in `code` package take both `ackgenconfig` and `ackresource` as parameters, but this is unnecessary because [`ackresource` has `ackgenconfig` as an attribute](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/crd.go#L65):
```
bwagner5, line 163: Can you change these links to perma links so that they are relevant after any changes to the code in those files.

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+  * [GetOutputShape](https://github.com/aws-controllers-k8s/code-generator/blob/59c6892e4b61b5e2076e5e5504daba8278b82980/pkg/model/crd.go#L442)
+  * [GetAllRenames](https://github.com/aws-controllers-k8s/code-generator/blob/59c6892e4b61b5e2076e5e5504daba8278b82980/pkg/model/crd.go#L644)
+  * [GetOutputWrapperFieldPath](https://github.com/aws-controllers-k8s/code-generator/blob/59c6892e4b61b5e2076e5e5504daba8278b82980/pkg/model/crd.go#L420)
+
+* Helpers like [FindIdentifiersInShape](https://github.com/aws-controllers-k8s/code-generator/blob/26e5da2e7656bb836ee438c05df14f2adc50197d/pkg/generate/code/common.go#L31-L33) are implemented in the `code` package, but are not related to code generation
+
+
+### More work for contributors
+Developers needs to keep a mental model of which configs have been applied or need to be consulted:
```
bwagner5, line 192: nit s/needs/need/

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+  * [//TODO: should these fields be renamed before looking them up in spec?](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L143)
+  * [//TODO: check generator config for exceptions?](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/generate/code/set_sdk.go#L231)
+
+And because adding new features requires consulting config throughout code, some "lower priority" areaas get put on the back burner:
```
bwagner5, line 196: nit: s/areaas/areas/

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+## Problem
+*code-generator's* accelerated growth is resulting in tech debt that degrades the experience for platform maintainers and contributors. The codebase is becoming difficult to read and extend given the quick addition of features to onboard more services. This doc summarizes and introduces the proposals to eliminate significant tech debt so that development on ACK Platform is as pleasant and clear as it is open.
+
+
+## Solution
+
+### Overview
+To improve the implementation, I propose consolidating `ackconfig` access and any *ackconfig-aws-sdk* `reconciliation` into `ackmodel`. Taking this idea further, `ackmodel` should be the "source of truth"/required input for `api` and `controller` commands instead of *generator.yaml* and *aws-sdk*.
```
bwagner5, line 16: ++ I like the idea of a single-pane of glass with `ackmodel`! Seems much easier to reason about.

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,57 @@
...
+* `m` represents `ackmodel`
+* `SetResource` calls `m.GetOutputShape(op)` now instead of `crd`
+* Under the hood, `ackmodel` will access its `ackgenconfig` to get config values
+* Then use helpers to reconcile between `sdk-shape` and `ackgenconfig`
+* With `m` being the source of truth and reconciliation logic, it is safe to assume the output shape you get back takes all configs into account
+* Also, `resource` is no longer bogged down with helpers unrelated to a `resource`
+
+
+### New command `./ack-generate model`
+With a new `model` command, the code generation can flow from a common data source which also creates opportunity for future commands:
```
bwagner5, line 49: ++  model command for dev workflow!

bwagner5 (commented):
Nice docs! I like a lot of these ideas to streamline the dev process and provide a more centralized view of the world with ackmodel.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+![updated-pipeline](./images/proposed_pipeline.png)
+* This is not an exhaustive diagram of the calls, but shows the clear responsibility of `generateModel` and how downstream commands like `api` and `controller` become significantly lighter and easier to follow.
+* `ackmodel` will be cached in the same folder as `aws-sdk` and downstream commands will check for `ackmodel` instead of `aws-sdk`
+* users can override with a path to their own `ackmodel` (for even quicker testing!), but by default will check `--cach-dir`
+* No more dependency on $SERVICE-controller repo. The `ackmodel` will be generated with a specific version which can be extracted after commands `GetACKModel()`
+
+Updated Dev Loop:
+---
+![updated-dev-loop](./images/proposed_dev_loop.png)
+* Diffing the `ackmodel` between *code-generator* runs will be quicker than manually checking files or running e2e tests
```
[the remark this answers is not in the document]
brycahta: Good point. I think it depends on what changes/configs are being edited, but for the most part I would agree that comparing `ackmodel`s won't necessarily be easier (at least today). One example where I do think it's easier would be `output_wrapper_field_path` addition/removal. Given we edit `aws-sdk` directly in `ackmodel` (per 2nd proposal), it would be a lot easier to do a diff between `ackmodel.sdkapi` before/after `./ack-generate model` runs to see whether or not a shape is being unwrapped properly vs. today's approach where we not only diff files in `$SERVICE-controller` but also manually checking for syntax errors/incorrect types being used.

Overall, comparing `ackmodel` will probably start having more benefits in the future especially as we simplify the interface in `generator.yaml` and push down complexity. Adding a dev tool to help compare `ackmodel` between runs could also be a possibility and will result in easier debugging and quicker feedback loops for devs.

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+* `reconciliation`: logic involving access to or relations between `resource`, `shape`, `ackgenconfig`, and `aws-sdk`
+* `ackmodel`: the [code](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L36) representation of ACK's view of the world; the source of truth for `aws-sdk`, `ackgenconfig`, and `reconciliation`
+
+## Problem
+It is becoming increasingly difficult to contribute to and maintain *code-generator* due to unclear encapsulations and dispersed logic throughout the codebase. For example, `ackgenconfig` is passed throughout code and accessed in a variety of ways, and logic involving `reconciliation` is inconsistent, both in implementation and interface. Some specific examples include:
+  * [overextended encapsulations](#overextended-encapsulations)
+  * [long function parameter lists](#long-parameters)
+  * [inconsistent use of `ackgenconfig`](#ackgenconfig-use)
+  * [increased cognitive load for contributors](#more-work-for-contributors)
+
```
RedbackThomson, line 16: Might I add, while we are going for this, inconsistent use of `resource` structs i.e. references to `latest` and `desired`, and setting properties on either seemingly arbitrarily at points.

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+### Requirements
+* No impact to existing functionality
+* Code is easy to read, intuitive, and extendable
+* Naming in code is consistent (ex: Find vs. Get)
+* Navigating code is easier because pkg responsibility is clear
+
+
+### Prerequisites
+*code-generator* contains unit tests and also runs a subset of $SERVICE-controller e2e tests against PRs. Are these comprehensive enough to detect breakage in the code that we'll be touching/refactoring? If there are gaps between tests and code being refactored, then **tests will need to be created or updated** prior to implementing the solution.
```
RedbackThomson, line 34: > Are these comprehensive enough to detect breakage in the code that we'll be touching/refactoring?

We need to identify the possibly affected code paths and determine whether these are being covered in any of our tests

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+// Current
+func getSortedLateInitFieldsAndConfig(cfg *ackgenconfig.Config, r *model.CRD) ([]string, map[string]*ackgenconfig.LateInitializeConfig)
+
+
+// Proposed
+func (c *Config) GetLateInitFieldsAndConfig(cfg *ackgenconfig.Config, r *model.CRD) (map[string]*ackgenconfig.LateInitializeConfig, error)
+
+// Note, this simply gets the config; sorting and any processing AFTER getting config is done by the client
```
RedbackThomson, line 76: I disagree with the idea that we cannot have additional helper logic in the `config` package. I think it's standard practice to sanitize user input in a getter. I think the solution here is that the name should reflect the underlying data `GetLateInitFieldsAndConfig`, leaving the sorting and sanitising as an implementation detail (left in the go doc).

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+
+```
+
+#### Centralize `reconciliation`
+Similar to the above, move `reconcialiation`-related functions from `code` package to `ackmodel` to realign responsibilities.
+
+1. Refactor `type CRD struct` in [crd.go](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/model/crd.go#L64) to remove `sdkapi`  as attribute and fix breakage
```
RedbackThomson, line 83: I can tell you I tried to do this when working on the [`custom_shapes`](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/sdk/custom_shapes.go) work. This is a lot harder than you might originally think. I believe it will take quite a bit of untangling and abstraction to pull off, but it is absolutely worth it to do.

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+```
+// Current
+func FindIdentifiersInShape(r *model.CRD, shape *awssdkmodel.Shape, op *awssdkmodel.Operation) []string
+
+// Proposed
+func (m *Model) GetIdentifiersInShape(r *model.CRD, shape *awssdkmodel.Shape, op *awssdkmodel.Operation) ([]string, error)
+
+```
+
+* Note, for scenarios that do both `reconcile` and output Go code (ex: [late_initialize::FindLateInitializedFieldNames](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/generate/code/late_initialize.go#L31), these will need to be refactored into separate pieces prior to migration
```
RedbackThomson, line 108: ++

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+#### Align access/interface for `__ForStruct` funcs
+The goal is to make it easier for clients to use `SetResource` and `SetSDK` by removing the need to know the target type.
+
+1. Refactor access (public vs. private) for `SetResourceFor___`, and `SetSDKFor___`
+  * `setResourceForContainer` and `setSDKForContainer` are private while `SetResourceForStruct` and `SetSDKForStruct` are public despite the former pair being more general than the latter. Access should be swapped so users calling this in [controller.go](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/generate/ack/controller.go#L102) don't need to worry about Struct or what type.. just pass container. Leaves it open-ended for override types too.
```
RedbackThomson, line 115: This was done in an effort to use these code generator methods within service controller template files. I still think that use case is valuable, and all of how we access and pass parameters to any of the `Set*` code generator methods needs some review with this in mind.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+* `ackmodel`: the [code](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L36) representation of ACK's view of the world; the source of truth for `aws-sdk`, `ackgenconfig`, and `reconciliation`
+
+## Problem
+There are issues that are starting to degrade the code generation `pipeline` experience and will compound as *code-generator* continues to grow:
+* Duplicate work between `api` and `controller` commands; both generate `ackmodel`
+* Overloaded functions such as [SetResource](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/generate/code/set_resource.go#L74); responsible for a lot more than generating Go code
+* Confusing CX. There's a not-so-clear dependency on a folder in the $SERVICE-controller via [getLatestAPIVersion](https://github.com/aws-controllers-k8s/code-generator/blob/26e5da2e7656bb836ee438c05df14f2adc50197d/cmd/ack-generate/command/common.go#L271). This goes against the flow of the `pipeline` and creates confusion as there is a `--version` flag for the `apis` command (the above function takes precedence).
```
RedbackThomson, line 14: I think some of this will be cleared up as multi-version support is fully implemented. At the moment the versioning is a work in progress.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+* Long dev loop; after making changes to either *generator.yaml* or logic in *code-generator* users need to either manually verify files or run e2e tests-- both of which are time consuming today
+* How *code-generator* reconciles `ackgenconfig` with `aws-sdk` to create `ackmodel` is inconsistent (and some duplicated work). For example, overrides such as `custom_shapes` [edit aws-sdk directly](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/sdk/custom_shapes.go#L62) while others such as `ignored_operations` [set operation values to nil](https://github.com/aws-controllers-k8s/code-generator/blob/26e5da2e7656bb836ee438c05df14f2adc50197d/pkg/model/model.go#L295) in memory leaving the actual operations in `aws-sdk` untouched.
+
+Current Pipeline
+---
+
+![current-pipeline](./images/current_pipeline.png)
+* Displayed are the important and mostly duplicated calls between the `apis` and `controller` commands
+* The cache is only used to store the `aws-sdk` repo
+* The red line denotes a dependency on $SERVICE-controller (output directory) repo that should be removed
```
RedbackThomson, line 24: For the purposes of multi-version support, we need to understand the previous structure of the CRDs to create the translation methods between the versions. I agree that reading output directories is a no-no. The problem is that we cannot know which runtime and code-generator version a previous model is compatible with, or even if we did the current code-generator cannot know what the output would look like for that version. I think @A-Hilaly will have a better understanding for the complexities in this.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+* Only `GetCRDs` is duplicated between the commands when creating the model
+
+
+Current Dev Loop
+---
+
+![current-dev-loop](./images/current_dev_loop.png)
+* Validating the result of a change in *code-generator* and/or *generator.yaml* can be a lengthy process
+* In the current feedback loop, users and contributors need to manually verify files or run lengthy e2e tests even for minor changes to `ackgenconfig`
```
RedbackThomson, line 33: This could be solved with unit testing. We have unit testing solutions, although not standardised.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+* The cache is only used to store the `aws-sdk` repo
+* The red line denotes a dependency on $SERVICE-controller (output directory) repo that should be removed
+* Only `GetCRDs` is duplicated between the commands when creating the model
+
+
+Current Dev Loop
+---
+
+![current-dev-loop](./images/current_dev_loop.png)
+* Validating the result of a change in *code-generator* and/or *generator.yaml* can be a lengthy process
```
RedbackThomson, line 32: I don't understand the "save previous `apis/` `pkg/`" in this diagram? Is this referring to verifying the git diff after generating code?

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+* In the current feedback loop, users and contributors need to manually verify files or run lengthy e2e tests even for minor changes to `ackgenconfig`
+
+
+## Solution
+Add a new command, `model`, which takes in `aws-sdk` and *generator.yaml* as inputs to create and cache `ackmodel`. Existing commands, `api` and `controller`, will remove parameters for `aws-sdk` and *generator.yaml* and replace them with `ackmodel` input and/or let it default to `--cache-dir`.
```
RedbackThomson, line 37: In your problem statement and "current dev loop", you mention the long validation process when changing the `generator.yaml` file. However, modifying the `generator.yaml` will bust the cache for this proposed `ackmodel` - meaning you would have to regenerate this cache every time anyway? This solution would only really be useful for quickly iterating on templates?

Furthermore, how much time does generating the `ackmodel` take? Assuming you are using the same AWS SDK version, it is probably less than a second to generate that model? Is this added complexity justified instead of caching `generator.yaml` files instead?

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+
+Updated Pipeline
+---
+![updated-pipeline](./images/proposed_pipeline.png)
+* This is not an exhaustive diagram of the calls, but shows the clear responsibility of `generateModel` and how downstream commands like `api` and `controller` become significantly lighter and easier to follow.
+* `ackmodel` will be cached in the same folder as `aws-sdk` and downstream commands will check for `ackmodel` instead of `aws-sdk`
+* users can override with a path to their own `ackmodel` (for even quicker testing!), but by default will check `--cach-dir`
```
RedbackThomson, line 44: When would users have their own `ackmodel`? Would users not already have their own `generator.yaml`?

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+---
+To address the inconsistent `ackmodel` creation, I propose we edit `aws-sdk` directly as much as we can, then introduce more fields to represent other relations that can't be applied by editing `aws-sdk`. Therefore, if we want to rename a field/resource, it would be instead completely removed from model. Conceptually:
+
+```
+func (m *Model) RenameShapes() error {
+  for shapeName, shape := range m.SDKAPI.API.Shapes {
+    if m.cfg.shapeRenamed(shapeName) {
+      updateName := m.cfg.GetShapeName(shapeName)
+      m.SDKAPI.Shapes[updatedName] = shape
+      delete(shapeName, shape)
```
RedbackThomson, line 64: I like this concept! Keep a modified version of the AWS SDK as our new base.

Unfortunately it's not quite as simple as this in reality. `ShapeName` is referenced both in the key of the `Shapes` map, but also in the `ShapeName` attribute of a `ShapeRef`. This means that updating the shape name would require finding all `ShapeRef` objects which match the shape and updating their attributes (this is non-trivial and the time complexity is recursive on number of attributes per shape).

Another concern I have is that we are now strongly coupled to the AWS SDK. We don't have plans to update to v2, but keeping some level of abstraction above the AWS SDK protects us from the subtle changes they could make to their internals.

docs/design/proposals/codegen-refactor/model_cmd_2.md
```diff
@@ -0,0 +1,143 @@
...
+
+```
+updatedShape := m.GetOutputShape(outputShape)
+
+for memberIndex, memberName := range updatedShape.MemberNames() {
+  // updatedShape already has its fields renamed, removed, or added to based
+  ...
+}
+
+```
```
RedbackThomson, line 80: ++ I love this

RedbackThomson (changes requested):
I really appreciate this deep dive into the ACK generator thought process. I absolutely agree with the necessity of these proposals, and generally agree with the big picture. 

I have left inline comments with concerns primarily around caching and retrieving the `ackmodel` object - the trade off we would have to make with complexity and coupling for this to be viable. As for the `ackgenconfig` work, I am really fond of that idea and am happy to brainstorm ways to untangle the code generator package structure mess.

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: > However, modifying the generator.yaml will bust the cache for this proposed ackmodel - meaning you would have to regenerate this cache every time anyway?

Good callout, I'll clarify in the next revision, but the intent is that the time savings is coming from being able to quickly compare `ackmodel`s between changes so users wouldn't need to follow longer validation processes involving building and running $SERVICE-controller. I had something like this in mind:
0. Set `cache-dir` to directory in code-generator, ex: `local-ack-cache`
1. Generate `ackmodel` using latest on *main*
2. Commit `local-ack-cache/ackmodel.json`
3. Change `generator.yaml`
4. Generate `ackmodel` with changed `generator.yaml`
5. `git diff local-ack-cache/ackmodel.json` to see changes


> This solution would only really be useful for quickly iterating on templates?

I agree that's mostly true for today. However, as the interface in `generator.yaml` becomes simpler and more powerful, 1-2 lines of config in `generator.yaml` may result in multiple complex relationships and overrides(stored in `ackmodel`) that may not be obvious. For platform devs, being able to quickly compare `generator.yaml` changes to relationships/overrides could mean quicker iteration for development and debugging.

> Furthermore, how much time does generating the ackmodel take? Assuming you are using the same AWS SDK version, it is probably less than a second to generate that model? Is this added complexity justified instead of caching generator.yaml files instead?

I don't have any performance metrics to compare, but also the small time savings by caching `ackmodel` isn't the focus. The biggest benefit to caching `ackmodel` is **consistency**:
* *Consistently* capture and provide the ACK state of the world to downstream commands like `generate apis` and `generate controller` and have confidence they generated controller artifacts based on the same source of truth. 


Caching *generator.yaml* only still leaves opportunity for compatibility issues:
  *  unintentionally changing `version` and `aws-sdk-go-version` flags between code-generator runs could impact artifacts despite *generator.yaml* being the same
  * generating apis, making an update to CRD creation logic, then generating controller would most likely result in incompatible types/syntax

These might be extreme examples, but hopefully shows the intent on reducing points of failure and possible inconsistencies (mainly for platform devs).

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: This would mainly be for power users/experimentation, but in general users wouldn't come with their own `ackmodel` but generate a valid `ackmodel` as a base, then work on top of that. There wouldn't be many scenarios, but one example is adding a new config to *generator.yaml*. One could directly edit a CRD's Spec/Status field or a Shape in `typeDefs` assuming the new config "worked", then generate apis/controller to ensure a valid/working approach.

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: Agreed this would be a very complex endeavor, but you're saying it's possible?? :)  Is your concern feasibility or the performance hit and having to maintain the more complex code? This could also be an ongoing effort where configs slowly start to edit aws sdk over time.

> Another concern I have is that we are now strongly coupled to the AWS SDK. We don't have plans to update to v2, but keeping some level of abstraction above the AWS SDK protects us from the subtle changes they could make to their internals.

I see your point. I think moving methods that directly edit aws-sdk from `Model` to [SDK Helper](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/sdk/helper.go#L50) and having an SDK Helper implemented for each supported api version could be a better abstraction.  This way we can support `v1.RenameShapes()` and `v2.RenameShapes()` if the sdks differed significantly. What do you think?

docs/design/proposals/codegen-refactor/centralize_1.md
brycahta: Could you clarify a bit? Is there a reason why we can't use setForContainer vs. the concrete type?

docs/design/proposals/codegen-refactor/centralize_1.md
RedbackThomson: > Feels like pre-optimizing the client's use case by sorting for them

I think we want as much of this stuff to run deterministically as possible, and sorting at the raw input will save us sorting in every other place where we access these values.

> Does it always make sense to sort a list of config? 

Nope. But in some places, yep. Case-by-case basis.

> Additional overhead of adding sort to every new getter (for list configs)-- both performance cost and developer effort

`O(n)` on the size of the config (probably <10) values. I would consider worrying about this early optimisation and not worth stressing.

docs/design/proposals/codegen-refactor/centralize_1.md
RedbackThomson: Initially every method was private, but we needed to expose some of these for custom hook code. I actually can't find where either of these are being used anymore (see [this search](https://github.com/search?q=org%3Aaws-controllers-k8s+SetSDKForStruct&type=code)), so I'm not too sure why we didn't expose the container. 

From memory, it was because the container version requires we pass an additional field that wasn't available within the hook code. But that's a vague guess at best.

docs/design/proposals/codegen-refactor/model_cmd_2.md
RedbackThomson: Both, although I was implying unit tests within the $SERVICE-controller repository.

docs/design/proposals/codegen-refactor/model_cmd_2.md
RedbackThomson: > Agreed this would be a very complex endeavor, but you're saying it's possible

Tentatively yes. Doesn't mean we should be, though.

jaypipes (changes requested):
@brycahta I really like a lot of the motivation behind these proposals, but I'm afraid we're going to need to more closely consider some of the ideas here. I've left some notes inline but it would be best to have a videochat to go over the primary concerns I have. I think there are a number of work items -- particularly around being consistent in naming and how we use config or model structs internally -- that will be relatively easy to accomplish and would quickly and positively impact the readability of the code generator. Other work items, particular the consolidation of the model and config packages, would likely be a lot of heartache and should be taken on after other more incremental improvements are made (such as breaking the larger Go code generator functions into smaller more reusable functions).

docs/design/proposals/codegen-refactor/centralize_1.md
```diff
@@ -0,0 +1,197 @@
...
+
+1. Remove `ackgenconfig` as an attribute in `type CRD struct` in [crd.go](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/model/crd.go#L65) and fix breakage
+  * move the associated methods to `config` package and update callers, ex: [GetOutputWrapperFieldPath](https://github.com/aws-controllers-k8s/code-generator/blob/59c6892e4b61b5e2076e5e5504daba8278b82980/pkg/model/crd.go#L420):
+
+```
+// Current
+wrapperFieldPath := r.GetOutputWrapperFieldPath(op)
+
+// Proposed
+wrapperFieldPath := cfg.GetOutputWrapperFieldPath(op)
```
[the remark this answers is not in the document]
brycahta: Maybe I'm misunderstanding, but today's [GetOutputWrapperFieldPath](https://github.com/aws-controllers-k8s/code-generator/blob/59c6892e4b61b5e2076e5e5504daba8278b82980/pkg/model/crd.go#L420), `CRD` is used only to access `ackgenconfig` to obtain `opConfig`, if any. Therefore, it's the operation not the resource/CRD that needs to be provided.

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: Do you know if any $SERVICE-controller repos run unit tests today? Only seeing e2e tests on my search

RedbackThomson: Yes check out elasticache - https://github.com/aws-controllers-k8s/elasticache-controller/blob/main/pkg/resource/replication_group/manager_test_suite_test.go

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: Correct me if I'm wrong, but this is assuming an immutable `aws-sdk` being compared to an altered [Model.SDKAPI](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L37) which is not the case in this proposal. 

I'm proposing we modify the `aws-sdk-go` directly to adhere to `generator.yaml`'s "reality" so that `CreateBucketInputShape` in `Model.SDKAPI` would also be called `Name` instead of `Bucket` (using your example). If for whatever reason we cannot edit `aws-sdk-go` directly to match `generator.yaml`, then we can leverage new or existing fields in `ackmodel` such as `shapeRenames` (new!) or `typeRenames` to capture this override and resolve `API Inference` in places like `SetResource`:

If we cannot apply a config from `generator.yaml` to `aws-sdk-go` for reasons, we can use aux fields in `ackModel` to resolve the inference:

```
shape := m.GetOutputShape(outputShape)
for memberIndex, memberName := range shape.MemberNames() {
  // shape is the unaltered shape from aws-sdk-go
 // consult the model if API inference cannot be resolved
  if !memberFound {
    _, found := m.ShapeRenames[memberName]
   ...
}
}
```

The idea stems from if we need to do the work of applying `generator.yaml` to `aws-sdk-go` during model creation, then we should retain that knowledge somehow so we don't need to figure it out all over again in SetResource/SetSdk

docs/design/proposals/codegen-refactor/model_cmd_2.md
brycahta: `aws-sdk-go` is edited and part of `ackmodel` in this proposal/example not an immutable version. More info in comment above

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+### Requirements
+* Refactor does **not change** existing functionality
+* Code is easy to read, intuitive, and extendable
+* Code generation processes remain clear and transparent
+
+## Approach
+
+### Centralize config access and reconciliation
```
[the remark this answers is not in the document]
brycahta: Correct, saw your comment above and can re-word in next revision

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+---
+
+![proposed-config-access](./images/proposed_config_access.png)
+* `m` represents `ackmodel`
+* `SetResource` calls `m.GetOutputShape(op)` now instead of `crd`
+* Under the hood, `ackmodel` will access its `ackgenconfig` to get config values
+* Then use helpers to reconcile between `sdk-shape` and `ackgenconfig`
+* With `m` being the source of truth and reconciliation logic, it is safe to assume the output shape you get back takes all configs into account
```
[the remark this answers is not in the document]
brycahta: Yes, that's the intent here. thank you for making that clearer; I'll update the description.

docs/design/proposals/codegen-refactor/proposal-summary.md
brycahta: Thanks for pointing this out, I'll make it clearer in next revision.

The output would be an [ackmodel](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/model/model.go#L36), serialized into JSON*, and stored in `~./cache/aws-controllers-k8s/ack-model.json`.

Something like:
```
./ack-generate model <params>
...
ACK model generated and stored successfully in ~./cache/aws-controllers-k8s/ack-model.json
```

Downstream commands will use default cache_dir to look for `ackmodel`. If none can be found, then error out requesting `ackmodel` to be built first.


* The current, proposed `ackmodel` as JSON would be huge! Experimenting locally and discussing offline with @RedbackThomson  to see how to break it up

docs/design/proposals/codegen-refactor/proposal-summary.md
brycahta: By combining config with ACK model, we are able to create this "source of truth" so that downstream commands (apis, controller) don't need to "think" for themselves, but instead focus on single responsibilities like how to generate Go code. For example, `SetResource` is accessing and applying `ackgenconfig` as well as generating Go code which has overloaded it with responsibility. Also, this work of applying `ackgenconfig` and resolving it with `aws-sdk` should have already been done since it would be required to generate the CRDs before kicking this process off; therefore, removing duplicate work.

We also organize the codebase significantly by consolidating this logic to a central module and extending code-generator becomes more intuitive. If I can't find a field, resolve an api inference, or need info about `ackgenconfig`, `aws-sdk`, etc. then ask the source of truth, `ackmodel`. Compared to today's less clear experience where you can access `ackgenconfig` in a number of ways (directly or thru CRD) and resolve yourself or try to find a related helper function. It'd also easier to extend config/sdk-resolve logic as it will all live `model` pkg vs. spread out like today, `code/common.go`, `model/crd.go` for example.

docs/design/proposals/codegen-refactor/centralize_1.md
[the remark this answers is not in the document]
jaypipes: This is one of those things that I've been meaning to refactor when I've said I want to move from an operation-centric approach to a resource-centric approach :)

docs/design/proposals/codegen-refactor/centralize_1.md
brycahta: resolving; this was related to `runtime`

docs/design/proposals/codegen-refactor/centralize_1.md
brycahta: Aligned offline. We expect that user-configurable slices (provided in `generator.yaml`) will **never** need to *preserve order*; therefore, it should be safe/consistent/intuitive to sanitize and sort in `Getter`.

If such a config/override is needed in the future, then the order should be preserved using an index mechanism or some other abstraction more powerful than a slice.

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,70 @@
...
+## Approach
+
+### Split `ackgenconfig` into 2 categories
+Move configs to its own pkg, `pkg/config`, then split `ackgenconfig` into 2 categories:
+  * **inference**: configs used to *infer* a relation between `resource` and `aws-sdk` such as `renames` located in `pkg/generate/config/inference.go`
+  * **code-generating**: configs used to instruct the code-generator on how to *generate* Go code for a resource such as `output_wrapper_field` located in `pkg/generate/config/generate.go`
+
+
+
+![current-config-access](./images/current_config_access.png)
```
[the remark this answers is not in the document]
brycahta: "access" here refers to the ability to fetch data. The diagram shows how `ackgenconfig` is *accessed* under the hood when calling `r.GetOutputWrapperFieldPath`

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Problem
+*code-generator's* accelerated growth is resulting in tech debt that degrades the experience for platform maintainers and contributors. The codebase is becoming difficult to read and extend given the quick addition of features to onboard more services. This doc summarizes and introduces proposals to eliminate significant tech debt and bring clarity to the `pipeline` so that development on ACK Platform is as pleasant and clear as it is open.
+
+
+## Solution
+
+### Overview
+The tech debt will be reduced after stratifying `ackgenconfig` into catgories and refactoring Go code generating logic. Code will be moved, modified, and new packages will be created in order to accomplish this effort. A new command, `infer-model`, will be added to the code-generator `pipeline` to make the overall flow clearer and more efficient. `infer-model` will resolve relations between `ackgenconfig` and `aws-sdk` and caches the "inference" data so that it can be consumed by downstream commands, `apis` and `controller`.
```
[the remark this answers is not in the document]
brycahta: In the [current proposal](https://github.com/brycahta/ack-community/blob/refactor-codegen/docs/design/proposals/codegen-refactor/infer_model_cmd.md#updated-ackmodel), `InferredModel` is a struct and attribute within `Model`:

```
type InferredModel struct {
	apiVersion         string
	crds               []*CRD
	typeDefs           []*TypeDef
	typeImports        map[string]string
	typeRenames        map[string]string
}

type Model struct {
	SDKAPI             *SDKAPI
	servicePackageName string
         InferredModel    *InferredModel `json:"inferred_model"`
	// Instructions to the code generator how to handle the API and its
	// resources
	cfg *ackgenconfig.Config
}

```

hence the naming discrepancy. I can replace `Model` with the contents of `InferredModel` instead and drop the rest instead. Let me know if that works and I can update naming throughout and make a note to fix the breakage caused by removing those fields from `Model`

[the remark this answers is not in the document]
jaypipes: @brycahta Yeah, I would just merge InferredModel with Model.

docs/design/proposals/codegen-refactor/proposal-summary.md
```diff
@@ -0,0 +1,70 @@
...
+  * **inference**: configs used to *infer* a relation between `resource` and `aws-sdk` such as `renames` located in `pkg/generate/config/inference.go`
+  * **code-generating**: configs used to instruct the code-generator on how to *generate* Go code for a resource such as `output_wrapper_field` located in `pkg/generate/config/generate.go`
+
+
+
+![current-config-access](./images/current_config_access.png)
+* `r` represents a `resource` or `crd`
+* `SetResource` calls `r.GetOutputShape(op)` to retrieve the shape for a given operation
+* After, `SetResource` calls `r.GetOutputWrapperFieldPath(op)`; it's basically a wrapper for accessing `ackgenconfig`
+* Finally, `r.getWrapperOutputShape()` is called recursively and resolves which shape to return
```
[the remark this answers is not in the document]
brycahta: good point; definitely not flowing clearly. I think this was leftover from the original proposal where I wanted to consolidated ALL configs and how to access them. I'll remove diagrams and clarify intent

docs/design/proposals/codegen-refactor/inference.md
```diff
@@ -0,0 +1,225 @@
...
+Agree on terms for consistent code. Terms such as Get vs. Find, CRDS/Resource, etc.
+
+### Implementation
+
+#### Consolidate `ackgenconfig` access and expose Getters
+The purpose is to realign responsibilities of `CRD` and `Config` making code more consistent and intuitive.
+
+1. Move `ackgenconfig` to its own package, `pkg/config`
+
+2. Remove `ackgenconfig` as an attribute in `type CRD struct` in [crd.go](https://github.com/aws-controllers-k8s/code-generator/blob/25c43e827527b43c652b6e1995265c8d6027567f/pkg/model/crd.go#L65) and fix breakage
```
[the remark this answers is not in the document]
brycahta: IMO this is overextending `CRD`. The single resource shouldn't have need for the entire `ackgenconfig` and if it needs to access config/inferences then it should use `model` and `config` helpers to do so.

docs/design/proposals/codegen-refactor/inference.md
```diff
@@ -0,0 +1,225 @@
...
+
+```
+type Config struct {
+    ...
+	ModelName string `json:"model_name,omitempty"`
+    InferenceConfig *InferenceConfig
+    GenerateConfig *GenereateConfig
+}
+```
+*Note, Inference/Generate config is NOT exposed in `generator.yaml` as this is an implementation detail*
```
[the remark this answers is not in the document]
brycahta: I was thinking *generator.yaml* remains unchanged and we absorb the complexity by populating `InferenceConfig` and `GenerateConfig` fields ourselves after sanitizing/validating configurations. The new category fields will be used as the data source backing config helpers. Ex: `GetOutputWrapperShape() --> <access GenerateConfig>`

docs/design/proposals/codegen-refactor/inference.md
```diff
@@ -0,0 +1,225 @@
...
+  * [inconsistent use of `ackgenconfig`](#ackgenconfig-use)
+  * [increased cognitive load for contributors](#more-work-for-contributors)
+
+
+## Solution
+The proposal is to delineate `ackgenconfig` into 2 categories, **inference** and **code-generating**. The structures, `Getters/Setters`, and helpers will be defined in the `config` package (migrate from `pkg/generate/config` to `pkg/config`) and encapsulated in files by category:
+  * **inference**: configs used to *infer* a relation between `resource` and `aws-sdk` such as `renames`
+  * **code-generating**: configs used to instruct the code-generator on how to *generate* Go code for a resource such as `output_wrapper_field`
+
+Encapsulation will be improved because functions consuming `ackgenconfig` will grab configurations *relevant to its responsibility only*. For example, functions responsible for generating Go code, ex: `SetResource`, need only **code-generating** configs; therefore, one can eliminate any calls/logic relating to **inference** or other `ackgenconfig` parsing.
```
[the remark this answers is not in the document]
brycahta: I agree that's clearer. My only reservation is the inference stage isn't introduced/clearly defined until it's broken out into `model` command and I didn't want too many dependencies between docs. But If you think that's a non-issue I can include the detail 👍

jaypipes (approved):
I'm cool with all of these proposal pieces. I may have a few little points of debate to raise when the PRs come in, but I'm supportive of this overall effort, @brycahta, thanks for your efforts in putting this together!

PR #1148, opened 2022-01-28
Semantics for destructive operations design proposal

RedbackThomson:
Issue #, if available: https://github.com/aws-controllers-k8s/community/issues/82

Description of changes:
Draft proposal for mechanisms to protect users against accidental deletion of ACK and/or AWS resources

### DIFF
--- a/docs/design/proposals/destructive-operations/destructive-operations.md
+++ b/docs/design/proposals/destructive-operations/destructive-operations.md
@@ -4,42 +4,19 @@
 
 This design document stems from the initial discussion on the [original Github issue #82](https://github.com/aws-controllers-k8s/community/issues/82).
 
-ACK currently treats all resources with the same amount of protection for deletion. That is, if you delete an ACK custom resource, the underlying AWS resource is deleted as well. While this might be the expected behaviour for the vast majority of resources, it also increases the chance that an AWS resource is accidentally deleted - through prematurely or accidentally deleting the ACK CR. Some sort of deletion protection should be introduced to limit the odds of this happening, requiring the user to unset a property manually before the controller will continue with deletion.
+When deleting an ACK custom resource, the underlying AWS resource is first deleted. Users should expect this to be the default behaviour, as creating and updating CR similarly have the same respective actions in AWS. However users have requested that some resources should continue to live on after the ACK CR has been deleted. Resources with stateful data, such as S3 buckets, RDS instance or EBS volumes, may want to be retained so that they may be migrated between ACK installations or off ACK entirely without requiring they be backed up, deleted and then restored.
 
-Another consideration of destructive operations is that some resources should continue to live on after the ACK CR has been deleted. Similar to the CDK[`RemovalPolicy` type](https://docs.aws.amazon.com/cdk/api/v2//docs/aws-cdk-lib.RemovalPolicy.html#members) and the native [K8s PV reclaim policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming), ACK should support the ability to leave the AWS resource alive while still deleting the custom resource object. Users typically use these removal policy options when working with stateful data, such as S3, EBS and RDS resources, in cases where they would like to migrate the management of the resources from one system to another without having to recreate the data.
-
-This design document proposes different mechanisms to address each of the following concerns:
-
-* Protection against deletion
-* Control over AWS removal policy
-
-## Delete Protection
-
-### Requirements
-
-1. Multiple levels of granularity over resource protection (namespace-wide and resource-wide)
-2. Protection should be defined per-service (if namespace-wide)
-3. Controller must strictly check for existence of resource protection before any deletion action
-4. The controller must not edit any of the resource protection marks
-
-### Investigation
-
-Kubernetes finalizers are the standard way to protect a resource from deletion. A finalizer placed on a resource prohibits it from being deleted out of etcd until it is removed - either through code or by manually editing the resource. ACK already places finalizers on every resource, however these are used to guard the custom resource object being deleted before the the underlying AWS resource. Finalizers, however, do not apply in a cascading effect to resources below them ie. a finalizer placed on a namespace will not protect the deletion of a custom resource within it. This violates requirement 1, as there is only resource-wide granularity over resource protection. 
-
-Finalizers do not fit with requirement 3. If the user places a finalizer on an ACK resource, then issues a deletion of that resource, the ACK controller will immediately delete the AWS resource and then remove its own finalizer, regardless of any other finalizer on the object. The current ACK controller does not look for other finalizers on the object before running its delete logic. Only the K8s control plane uses finalizers to protect it from deletion in the key-value store, it does not limit the interaction of custom controllers.
-
-There is on ongoing issue regarding similar functionality in the [Kubernetes Github repository](https://github.com/kubernetes/kubernetes/issues/10179).
-There is a feature request for the same function in the [Azure service operator repository](https://github.com/Azure/azure-service-operator/issues/1633). 
+Similar to the CDK[`RemovalPolicy` type](https://docs.aws.amazon.com/cdk/api/v2//docs/aws-cdk-lib.RemovalPolicy.html#members) and the native [K8s PV reclaim policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming), ACK should support the ability to leave the AWS resource alive while still deleting the custom resource object. 
 
 ## AWS Removal Policy
 
 ### Requirements
 
-5. Multiple levels of granularity over resource protection (controller-wide, namespace-wide and resource-wide)
-6. Protection should be defined per-service (if namespace-wide)
-7. Users are able to define either of:
-    - The AWS resource should be deleted along with the CR
-    - The AWS resource is left in-tact after the CR is deleted
+1. Multiple levels of granularity over resource protection (controller-wide, namespace-wide and resource-wide)
+2. Protection should be defined per-service (if namespace-wide)
+3. Users are able to define either of:
+    1. The AWS resource should be deleted along with the CR
+    2. The AWS resource is left in-tact after the CR is deleted
 
 ### Investigation
 
@@ -48,7 +25,7 @@ A number of K8s resources already have similar functionality:
 * Helm uses an annotation: `helm.sh/resource-policy: keep`
 * Google Config Connector uses an annotation: `cnrm.cloud.google.com/deletion-policy: abandon`
 * Skaffold uses a label: `skaffold.dev/cleanup: "false"`
-* K8s PV use a spec field: `persistentVolumeReclaimPolicy`: Retain`
+* K8s PV use a spec field: `persistentVolumeReclaimPolicy``: Retain`
 
 
 Existing AWS products have their own variants:
@@ -60,19 +37,18 @@ CloudFormation and CDK both provide the ability to snapshot given resources befo
 
 ## Proposed Option
 
-The following proposed options will introduce:
+The following proposed option will introduce:
 
-* Two new annotations (each with a service-wide and an ACK-wide variation) for namespace and resource metadata fields
+* Two new annotations, one service-wide and one ACK-wide, to be applied on namespace and ACK resources
 * One new optional command line argument for the controller binary
 
-(Requirements 1 & 5) Following the current CARM design, which also requires the same levels of granularity, it is consistent to mark the policies in the annotation metadata for the namespace and resources, and to create a new command line argument on the controller for the controller-wide policies. The controller would place precedence on the resource policy, then the namespace policy and finally the controller policy - the annotation on the object with the highest precedence will set the policy for the given resource.
+(Requirements 1) Following the current CARM design, which also requires the same levels of granularity, it is consistent to mark the policies in the annotation metadata for the namespace and resources, and to create a new command line argument on the controller for the controller-wide policies. The controller would place precedence on the resource policy, then the namespace policy and finally the controller policy - the annotation on the object with the highest precedence will set the policy for the given resource.
 
-(Requirements 2 & 6) The full annotation path needs to consist of an annotation prefix (typically a URL) before the annotation name. One variation should concern every ACK controller regardless of service, and one should discriminate for controllers only concerned with a single service. Cross-service annotations, finalizers and CRDS are already configured with the `services.k8s.aws/` prefix. For this option, prepending the service identifier to the beginning would identify that the annotation applies only to the given service, eg. `s3.services.k8s.aws/` . Hence, there would be two prefixes:
+(Requirements 2) The full annotation path needs to consist of an annotation prefix (typically a URL) before the annotation name. One variation should concern every ACK controller regardless of service, and one should discriminate for controllers only concerned with a single service. Cross-service annotations, finalizers and CRDS are already configured with the `services.k8s.aws/` prefix. For this option, prepending the service identifier to the beginning would identify that the annotation applies only to the given service, eg. `s3.services.k8s.aws/` . Hence, there would be two prefixes:
 
 * `services.k8s.aws/`: All ACK controllers will read the policy of this namespace/resource
 * `<service>.services.k8s.aws/`: Only the <service> ACK controller will read the policy of this namespace (taking precedence over the former option)
 
-### AWS Removal Policy
 
 The name of this annotation could be either:
 
@@ -81,7 +57,7 @@ The name of this annotation could be either:
 
 Users familiar with K8s would find the existing K8s variations more recognizable but may incorrectly assume that the annotation has the same values and behaves identically to native K8s resources. Therefore, it may be better to introduce the AWS terminology into the annotation - as the proposed functionality will behave identically to CDK and CloudFormation. Considering this annotation takes effect after the user calls “Delete”, I propose using `DeletionPolicy` as the annotation name.
 
-(Requirement 7) In order to account for a future possibility of adding additional policy options, the value of this annotation should be an enumeration. Considering the name already references the existing CloudFormation `DeletionPolicy` field, the values should similarly follow those values:
+(Requirement 3) In order to account for a future possibility of adding additional policy options, the value of this annotation should be an enumeration. Considering the name already references the existing CloudFormation `DeletionPolicy` field, the values should similarly follow those values:
 
 * `Delete` - deletes the AWS resource alongside the CR
 * `Retain` - retains the AWS resource, but deletes the CR
@@ -146,57 +122,9 @@ spec:
   name: retained-bucket
 ```
 
-### Delete Protection
-
-Proposed annotation names:
-
-* `no-delete`
-* `block-delete` / `prevent-delete` / `lock-delete`
-* `delete-lock`
-
-Acceptable values for this annotation will only need to be boolean `true/false` . The annotation set to `true` will indicate to the controller that it, or any ACK resources within its scope, should not be deleted. The default value for this annotation (when the annotation is omitted) will be `false` - the resource can be deleted.
-
-When a user attempts to delete a resource, a validating webhook will read through each of the annotations on the resource and the namespace. If the premature deletion annotation is defined on either of those, if the one with the higher precedence has the value set to `true`, the validating webhook will reject the delete request with a corresponding error message. 
-
-#### Examples
-
-Definition at the namespace (protecting all ACK resources across any service):
-
-```yaml
-apiVersion: v1
-kind: Namespace
-metadata:
-  annotations:
-    services.k8s.aws/prevent-delete: true
-  name: my-safe-namespace
-```
-
-Definition at the namespace (protecting only S3 ACK resources):
-
-```yaml
-apiVersion: v1
-kind: Namespace
-metadata:
-  annotations:
-    s3.services.k8s.aws/prevent-delete: true
-  name: my-safe-for-s3-namespace
-```
-
-Definition at the resource:
-
-```yaml
-apiVersion: s3.services.k8s.aws/v1alpha1
-kind: Bucket
-metadata:
-  name: test-bucket
-  annotations:
-    s3.services.k8s.aws/prevent-delete: true
-spec:
-  name: safe-bucket
-```
-
 ## Backwards Compatibility and Risks
 
 The proposed solution introduces two new annotations and an optional command line argument. The default behaviour for any resource without these annotations, or for the controller binary without this optional command, is backwards compatible with all current ACK controllers. There are no backwards compatibility issues with this solution.
 
 TODO: Identify possible risks with this proposal
+
--- a/docs/design/proposals/destructive-operations/destructive-operations.md
+++ b/docs/design/proposals/destructive-operations/destructive-operations.md
@@ -1,0 +1,202 @@
+# Semantics of destructive operations
+
+## Summary
+
+This design document stems from the initial discussion on the [original Github issue #82](https://github.com/aws-controllers-k8s/community/issues/82).
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
+# Semantics of destructive operations
+
+## Summary
+
+This design document stems from the initial discussion on the [original Github issue #82](https://github.com/aws-controllers-k8s/community/issues/82).
+
+ACK currently treats all resources with the same amount of protection for deletion. That is, if you delete an ACK custom resource, the underlying AWS resource is deleted as well. While this might be the expected behaviour for the vast majority of resources, it also increases the chance that an AWS resource is accidentally deleted - through prematurely or accidentally deleting the ACK CR. Some sort of deletion protection should be introduced to limit the odds of this happening, requiring the user to unset a property manually before the controller will continue with deletion.
```
jaypipes, line 7: > Some sort of deletion protection should be introduced to limit the odds of this happening, requiring the user to unset a property manually before the controller will continue with deletion.

Honestly, I really don't care about this "use case". It's so very imperative and GUI/console experience-centric. There is a reason we override all the `--dry-run` and `--skip-final-snapshot` attributes in our ACK controllers, and that is because we cater to an automated, declarative configuration management strategy. I don't really want to replicate the giant mess of "type `delete` here" (and myriad variations of this) AWS console dialog boxes in the ACK user experience.

On the other hand, I am 100% supportive of the opposite direction of thought: facilitating an **easy** and **consistent** way of deleting *child/dependent resources* when a parent resource has been deleted -- in other words, supporting a consistent cascading deletion experience for all AWS API resources.

Note that I do *not* consider a Kubernetes Namespace to be the parent of any ACK resource. In other words, deleting a Kubernetes Namespace should not delete all ACK CRs that happen to be in that Namespace. The Namespace should not be able to be deleted until all CRs or other Kubernetes-native resources in that Namespace are deleted themselves.

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+
+Another consideration of destructive operations is that some resources should continue to live on after the ACK CR has been deleted. Similar to the CDK[`RemovalPolicy` type](https://docs.aws.amazon.com/cdk/api/v2//docs/aws-cdk-lib.RemovalPolicy.html#members) and the native [K8s PV reclaim policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming), ACK should support the ability to leave the AWS resource alive while still deleting the custom resource object. Users typically use these removal policy options when working with stateful data, such as S3, EBS and RDS resources, in cases where they would like to migrate the management of the resources from one system to another without having to recreate the data.
```
jaypipes, line 9: > ACK should support the ability to leave the AWS resource alive while still deleting the custom resource object.

Sure, agree on this one, and this should basically be the reverse of AdoptedResource, IMHO.

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+
+Finalizers do not fit with requirement 3. If the user places a finalizer on an ACK resource, then issues a deletion of that resource, the ACK controller will immediately delete the AWS resource and then remove its own finalizer, regardless of any other finalizer on the object. The current ACK controller does not look for other finalizers on the object before running its delete logic. Only the K8s control plane uses finalizers to protect it from deletion in the key-value store, it does not limit the interaction of custom controllers.
+
+There is on ongoing issue regarding similar functionality in the [Kubernetes Github repository](https://github.com/kubernetes/kubernetes/issues/10179).
```
jaypipes, line 31: There is a reason that k/k issue 10179 is 7 years old and still not implemented. And that, IMHO, is because many developers realized that trying to keep users from fat-fingering things and accidentally deleting things is a completely futile effort. Users will always do stupid things. 99% of users will learn from their mistake. Interfaces and user experiences should not cater to babysitting and hand-holding imperative-style user interactions. Instead, interfaces and user experiences should focus on having no surprises and being hard to misuse.

In the case of ACK and the user experience of resource deletion, I posit that the behaviour of our interface has no surprises and is actually hard to misuse. It doesn't have surprises because our interface to delete a resource is to call `kubectl delete` on that resource. This is as non-surprising as it gets. It is hard to misuse because calling `kubectl delete` is an explicit action the user has to take and it's not a confusing verb or semantic. "delete" means to remove the resource.

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+    - The AWS resource should be deleted along with the CR
+    - The AWS resource is left in-tact after the CR is deleted
+
+### Investigation
+
+A number of K8s resources already have similar functionality:
+
+* Helm uses an annotation: `helm.sh/resource-policy: keep`
+* Google Config Connector uses an annotation: `cnrm.cloud.google.com/deletion-policy: abandon`
+* Skaffold uses a label: `skaffold.dev/cleanup: "false"`
```
jaypipes, line 50: There were also proposals originally to use an annotation/label for our resource adoption functionality. Instead of doing the annotation/label on a CR, we chose a user experience where the user needs to explicitly adopt an AWS resource by creating a `services.k8s.aws/AdoptedResource` CR. I continue to believe that we made the right decision in this regard: an explicit adoption action (via the `kubectl create -f <services.k8s.aws/AdoptedResource>` process) is the *least surprising, hardest to misuse interface*.

For the reverse of adoption action, I would prefer to have a similarly explicit action, essentially an "unadopt" or "unmanage" action. We could do this in a couple ways:

* Create a new kubectl plugin/verb called `kubectl unmanage` that the user could call to delete the CR but keep the AWS resource around
* `kubectl delete services.k8s.aws/AdoptedResource/<resource arn>` and have the AdoptedResourceReconciler run some "delete the CR but keep the AWS resource around" code

jaypipes (changes requested):
I am opposed to the entire "deletion protection" feature, for reasons stated inline in my comments.

I would be fine with a "unadopt resource" or "unmanage resource"  explicit action and user experience, however.

Please rename this proposal to `Deletion protection` because I'd like to see a separate proposal on cascading delete functionality (which I consider to be more aligned with the declarative configuration management principles that ACK should be encouraging).

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+(Requirement 7) In order to account for a future possibility of adding additional policy options, the value of this annotation should be an enumeration. Considering the name already references the existing CloudFormation `DeletionPolicy` field, the values should similarly follow those values:
+
+* `Delete` - deletes the AWS resource alongside the CR
+* `Retain` - retains the AWS resource, but deletes the CR
+
+The default value for this annotation will be `Delete` as it is in-line with the existing ACK functionality of deleting the underlying AWS resource.
+
+#### Examples
+
+Definition at the controller - all resources reconciled by this controller will use the “retain” policy (unless overridden by an annotation):
```
vijtrip2, line 93: I do not think "retain" should be the default use-case. We can keep current implementation as default i.e. Deleting a resource cleans things up on AWS side as well. In my opinion this will be default thinking of most users.

And option to override this cli-option for production use, stateful resources and databases will give users ability to prevent accidental deletion.

---------

Another less popular option can be if we get opinionated about AWS service controllers and store metadata about which service controllers will have "retain" as default option. Then helm/kustomization artifacts will only have deletion policy "retain" for those service controllers. However this bring inconsistency among service controllers by default.

RedbackThomson: > I do not think "retain" should be the default use-case

"delete" is the default. All default ACK functionality will remain the same for all considerations of this proposal. Only in *this particular example* do we override the default and set it to be "retain" at the command line argument level. 

> And option to override this cli-option for production use, stateful resources and databases will give users ability to prevent accidental deletion.

That is what this example is attempting to show.

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+
+Proposed annotation names:
+
+* `no-delete`
+* `block-delete` / `prevent-delete` / `lock-delete`
+* `delete-lock`
+
+Acceptable values for this annotation will only need to be boolean `true/false` . The annotation set to `true` will indicate to the controller that it, or any ACK resources within its scope, should not be deleted. The default value for this annotation (when the annotation is omitted) will be `false` - the resource can be deleted.
+
+When a user attempts to delete a resource, a validating webhook will read through each of the annotations on the resource and the namespace. If the premature deletion annotation is defined on either of those, if the one with the higher precedence has the value set to `true`, the validating webhook will reject the delete request with a corresponding error message.
```
vijtrip2, line 159: Clarifying Question: 

Is the purpose of using webhook instead of handling this logic in controller that the delete call will not even get persisted in etcd? 
IOW, w/o webhook, the delete call will go through and if someone removes finalizer manually , then custom-resource will get deleted. However with webhook, the delete will never get registered. Is my assumption correct?

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+    s3.services.k8s.aws/deletion-policy: retain
+spec:
+  name: retained-bucket
+```
+
+### Delete Protection
+
+Proposed annotation names:
+
+* `no-delete`
```
vijtrip2, line 153: Clarifying Question:

How does no-delete behavior work with GitOps method, where source repository removes the manifest?
Would the GitOps operation try to keep deleting the resource and keep failing?

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+
+Definition at the resource:
+
+```yaml
+apiVersion: s3.services.k8s.aws/v1alpha1
+kind: Bucket
+metadata:
+  name: test-bucket
+  annotations:
+    s3.services.k8s.aws/deletion-policy: retain
```
vijtrip2, line 144:
```suggestion
    services.k8s.aws/deletion-policy: retain
```

RedbackThomson:
@jaypipes 

> I am opposed to the entire "deletion protection" feature, for reasons stated inline in my comments.

I actually do agree with this. I wanted to write this proposal to get critical feedback like this. As I was writing the proposal I felt that we would be overengineering functionality. However, I still felt it was important to explore the technical challenges involved just so that we could all see the pros and cons of the function.

I will be removing the "Deletion protection" from this proposal - I might add an appendix where I mention this discussion to keep it for posterity.  I will pivot this proposal into discussing only the "AWS Removal Policy" functionality.

On that front:
> For the reverse of adoption action, I would prefer to have a similarly explicit action, essentially an "unadopt" or "unmanage" action

The reason we chose to have a separate CRD for adoption was mainly because we could not fit enough information into the existing CRD (or metadata) of the ACK resources, or we would have had to significantly redesign the optional/required fields. Further, the adoption reconciler *creates* a new ACK resource which the controller then reconciles as if it were previously created.

I don't believe the removal policy needs anything as sophisticated as this. This proposal only addresses whether the controller will or will not call `Delete*` in the SDK before it deletes the finalizer. We do not need to modify any elements of the spec or metadata (apart from this single annotation) on the ACK resource. The ACK resource is also immediately deleted after the delete operation, so having a remaining "Unadoption" custom resource would essentially only act as a log.

Your `kubectl unadopt` alias could easily be made, and would be as simple as:
```bash
kubectl annotate bucket my-bucket services.k8s.aws/deletion-policy=retain
kubectl delete bucket my-bucket
```

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    services.k8s.aws/deletion-policy: retain
+  name: my-retained-namespace
+```
+
+Definition at the namespace (protecting only S3 ACK resources):
```
vijtrip2, line 125: can this made be granular upto resource at namespace level?

Ex: `bucket.s3.services.k8s.aws/deletion-policy: retain`

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,202 @@
...
+#### Examples
+
+Definition at the namespace (protecting all ACK resources across any service):
+
+```yaml
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    services.k8s.aws/prevent-delete: true
```
vijtrip2, line 170: Clarifying Question:

Do you know what the namespace deletion experience is like when some custom resources fail to delete in the namespace?

vijtrip2 (commented):
I like this proposal. Great job!
I like how you broke the problem into two main parts, and referenced similar issues from other communities too.

I just left some clarifying questions and small comments.

docs/design/proposals/destructive-operations/destructive-operations.md
RedbackThomson: That's it, yes. 

Also I thought about a situation where we handle it within the controller directly. The controller would get called when etcd updates the deletion timestamp. We would then early return because it has the annotation on it. But how would we know when to requeue? We need to requeue the deletion either when the namespace or resource annotation has been removed (we would need a watcher) or we would requeue on an interval continuously and check the conditions every single time. Checking continuously would eat up CPU cycles and just seems messy, and using informers would require persisting state and a bunch of other ugly stuff. Therefore, let's just go with webhooks and block it before we even get to the controller - let the controller do only what it's made to do.

Anyway, this part of the proposal is being removed. But it was a nice thought.

docs/design/proposals/destructive-operations/destructive-operations.md
RedbackThomson: Yeah ... this one I'm not sure about.
As I stated above, I would like two variations: One just for an individual service controller, and one to protect any ACK resources within. So do we also allow both for an individual resource? They would both have the same behaviour if placed onto a resource. Or do we choose to only allow one variant - and which one would that be? I don't think I have a good answer for that either way - let me know if you do!

docs/design/proposals/destructive-operations/destructive-operations.md
RedbackThomson: That's an interesting idea. Like wildcard matching the resource type and API groups at any level.

docs/design/proposals/destructive-operations/destructive-operations.md
vijtrip2: I was suggesting that within the resource, the "s3" part of annotation was not serving any relevant purpose. On the resource we can just have "services.k8s.aws" and on namespace we can have multiple annotation prefix for different impact radius.

docs/design/proposals/destructive-operations/destructive-operations.md
vijtrip2: I was not necessarily suggesting a wildcard/regex matching, but just ability to specify that only retain a single resource from the service at the namespace level.

Although having a powerful matcher wouldn't hurt but we can just start small. :)

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,130 @@
+# Semantics of destructive operations
+
+## Summary
+
+This design document stems from the initial discussion on the [original Github issue #82](https://github.com/aws-controllers-k8s/community/issues/82).
+
+When deleting an ACK custom resource, the underlying AWS resource is first deleted. Users should expect this to be the default behaviour, as creating and updating CR similarly have the same respective actions in AWS. However users have requested that some resources should continue to live on after the ACK CR has been deleted. Resources with stateful data, such as S3 buckets, RDS instance or EBS volumes, may want to be retained so that they may be migrated between ACK installations or off ACK entirely without requiring they be backed up, deleted and then restored.
+
+Similar to the CDK[`RemovalPolicy` type](https://docs.aws.amazon.com/cdk/api/v2//docs/aws-cdk-lib.RemovalPolicy.html#members) and the native [K8s PV reclaim policy](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming), ACK should support the ability to leave the AWS resource alive while still deleting the custom resource object.
```
jaypipes, line 9: I'll be honest here... any time I see a reference to CloudFormation's (this is CDK, yes, but it's all CloudFormation...) [DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html), I shudder. I really despise the entire user experience of *configuring some tool to behave different ways when the user takes a specific action* (the action in this case is the `DELETE` action.  This is a non-obvious interface and I am opposed to it. Instead, we should encourage a user experience that requires the user to explicitly call a *different action* in order to have the tool behave differently.

jaypipes (commented):
I'm still not in favour of any of this feature as described in this proposal. The only thing I would be in favour of is something equivalent to an explicit `kubectl unadopt` command.

The absolute last UI we should be emulating is CloudFormation, IMHO.

docs/design/proposals/destructive-operations/destructive-operations.md
jaypipes: After chatting with @RedbackThomson a bit more about this, I'm going to take a slightly less rigid approach to this. Kubernetes API unfortunately does not have a "control channel" for which to send *instructions* to a *controller* about how to handle a particular *action*. Instead, we have the `metadata.annotations` Hammer which is unfortunate IMHO because this is on the *resource* and not where it should be: as decorative data on the *request*.

Alas, the `metadata.annotations` Hammer is all we have so I will reconsider this proposal in light of that and see if I can come up with a variation that doesn't offend my sense of poor coupling.

bwhaley:
Having read the proposal and the comments, I can see that there are some strong opinions and fierce opposition to this feature, so perhaps the point is moot. Nevertheless, I'd like to weigh in for some additional perspective since my team is starting to evaluate ACK and this was one area of concern.

Many teams, ours included, use k8s only for stateless applications. Losing a deployment/service/configmap due to an erroneous merge is easy to rollback and no permanent damage is done. Losing an RDS instance or S3 bucket, OTOH, could be a disaster from which there is no recovery. For that reason, some kind of protection would be welcomed.

Additionally, when using managed Kubernetes services like EKS in which the customer has no control plane access, there's a certain nervous tick that something could go wrong in etcd completely outside of customer control. What if a weird bug in EKS caused objects in etcd to disappear and as a result the underlying AWS resources are deleted? Some of the protections mentioned in the proposal might protect from this eventuality (e.g. a policy on the controller). Perhaps this is highly unlikely/nigh impossible, but is it fair to write off this worry entirely? We've already experienced unexplainable behavior with etcd on EKS (not this specific thing) so I think the concern is well founded.

"Unadopting" resources is only useful when removal is planned. It's the unplanned errors that I'd worry about.

I do hope this proposal will remain open for others to offer still more perspectives.

bwhaley:
> We would essentially be recreating RBAC in a very naive way, introducing yet another mechanism for people to configure permissions on top of K8s RBAC and AWS IAM.

AFAICT this proposal or any resource protection feature doesn't resemble RBAC in the least. There isn't anything about access control or permissions. Rather, this is about how to decide whether a resource should be deleted. To your point

> However, the proposal for adding a retain setting on resources I think still stands

Interesting. So you're suggesting something like (though not exactly) the following:
```
s3.services.k8s.aws/deletion-policy: retain
```
for any individual resource, but not for an entire controller or namespace? Am I understanding your point correctly?

If so, that's great, but that smells like a deletion policy to me.

RedbackThomson:
> AFAICT this proposal or any resource protection feature doesn't resemble RBAC in the least. There isn't anything about access control or permissions. Rather, this is about how to decide whether a resource should be deleted.

Oh, actually, I removed that from the proposal altogether. That's what the majority of the disagreement was about - designing an annotation that prevents deletion of a given resource. Nevermind, then.

> for any individual resource, but not for an entire controller or namespace

Correct. This is the current intention for this version of the proposal. 

> smells like a deletion policy to me

Well it was borrowed from the CloudFormation `DeletionPolicy` resource property, so that's not a coincidence.

bwhaley:
Got it! I think I was missing all the RBAC context. Now it makes more sense.

> > for any individual resource, but not for an entire controller or namespace
> Correct. This is the current intention for this version of the proposal.

Nice, seems really clean then.

> Well it was borrowed from the CloudFormation DeletionPolicy resource property, so that's not a coincidence.

FWIW, Terraform uses the term "prevent destroy" which also seems apropos to this proposal.

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,130 @@
...
+
+(Requirements 2) The full annotation path needs to consist of an annotation prefix (typically a URL) before the annotation name. One variation should concern every ACK controller regardless of service, and one should discriminate for controllers only concerned with a single service. Cross-service annotations, finalizers and CRDS are already configured with the `services.k8s.aws/` prefix. For this option, prepending the service identifier to the beginning would identify that the annotation applies only to the given service, eg. `s3.services.k8s.aws/` . Hence, there would be two prefixes:
+
+* `services.k8s.aws/`: All ACK controllers will read the policy of this namespace/resource
+* `<service>.services.k8s.aws/`: Only the <service> ACK controller will read the policy of this namespace (taking precedence over the former option)
```
A-Hilaly, line 50: Since there will always be only one controller responsible of deleting specific resources. Why do we need the `<service>.services.k8s.aws/` prefixed annotation?

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,130 @@
...
+    - --aws-region
+    - "$(AWS_REGION)"
+    - --enable-development-logging
+    - "$(ACK_ENABLE_DEVELOPMENT_LOGGING)"
+    - --log-level
+    - "$(ACK_LOG_LEVEL)"
+    - --resource-tags
+    - "$(ACK_RESOURCE_TAGS)"
++  - --deletion-policy
++  - "retain"
```
A-Hilaly, line 87: For the drift remediation feature, [we went for resource level configuration](https://github.com/aws-controllers-k8s/runtime/pull/106#discussion_r1055738553). On top of the `--deletion-policy` flag, do we want to add another flag to have the similar experience for deletion protection? thinking:
```bash
--resource-deletion-policy Function=Retain \
--resource-deletion-policy EventSourceMappings=Retain
```

docs/design/proposals/destructive-operations/destructive-operations.md
```diff
@@ -0,0 +1,130 @@
...
+kind: Bucket
+metadata:
+  name: test-bucket
+  annotations:
+    s3.services.k8s.aws/deletion-policy: retain
+spec:
+  name: retained-bucket
+```
+
+## Backwards Compatibility and Risks
```
A-Hilaly, line 125: This might out of scope for this proposal: Currently resources are deleted right after a DELETE API Call is successful. Some resources still hang for multiple minutes, and even days.. e.g `RDS::DBInstance`, `CloudTrail::EDS` ... few open questions:
- Do we still want to continue having same user experience in the future?
- If a user sets the deletion policy to `Delete`, do we want to keep the CR alive reflecting the real state of a resource?

A-Hilaly (commented):
Very clear 👍 can't wait to see this in action

docs/design/proposals/destructive-operations/destructive-operations.md
A-Hilaly: NVM looks like this is an namespace level annotation ++

docs/design/proposals/destructive-operations/destructive-operations.md
RedbackThomson: I think I actually have some issues with the resource level configuration of drift remediation, in that it's a non-standard format and seems a little bit convoluted. But I'll leave those comments on your PR.

For now I'll stick to having a global deletion policy flag, and leave the door open to more granular CLI flags later

PR #1226, opened 2022-03-15
Generate runtime API reference documentation

RedbackThomson:
Description of changes:
Updates the API reference generator script to support the APIs defined in the `runtime` repository. Also updates the HTML of the API reference page to place the APIs in their own section at the top under `Common`.

### DIFF
--- a/docs/scripts/gen_reference.py
+++ b/docs/scripts/gen_reference.py
@@ -34,7 +34,7 @@ class Field:
 class Resource:
     name: str
     service: str
-    service_package: str
+    crd_package_path: str
     group: str
     api_version: str
     description: str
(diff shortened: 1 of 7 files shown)


docs/scripts/gen_reference.py
```diff
@@ -144,6 +146,7 @@ def convert_crd_to_resource(crd: Dict, service) -> Resource:
     return Resource(
         name=crd['spec']['names']['kind'],
         service=service,
+        service_package=f"{service}-controller/apis" if service_package == "" else service_package,
```
jaypipes, line 149: this seems backward to me, no?

RedbackThomson: The Pythonic ternary operator just feels backwards to me in general. But this is correct. `{true_statement} if {condition} else {false_statement}`

RedbackThomson: Hmm let me just rewrite this as a Python default argument value

jaypipes (approved):
great work on this, @RedbackThomson

PR #1227, opened 2022-03-15
Add documentation for `FieldExport`

RedbackThomson:
Description of changes:
Adds a user guide describing the context, use case and example for `FieldExport`. 

### DIFF
--- a/docs/content/docs/tutorials/rds-example.md
+++ b/docs/content/docs/tutorials/rds-example.md
@@ -76,18 +76,15 @@ The examples below use the `db.t3.micro` instance type. Please review the [RDS i
 To create a [AWS RDS for PostgreSQL](https://aws.amazon.com/rds/postgresql/) instance, you must first set up a master password. You can do this by [creating a Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret), e.g.:
 
 ```bash
-RDS_DB_USERNAME="<your username>"
+RDS_INSTANCE_NAME="<your instance name>"
 
-kubectl create secret generic rds-postgresql-user-creds \
-  --from-literal=username="${RDS_DB_USERNAME}" \
+kubectl create secret generic "${RDS_INSTANCE_NAME}-password" \
   --from-literal=password="<your password>"
 ```
 
 Next, create a `DBInstance` custom resource. The example below shows how to provision a RDS for PostgreSQL 14 instance with the credentials created in the previous step:
 
 ```bash
-RDS_INSTANCE_NAME="<your instance name>"
-
 cat <<EOF > rds-postgresql.yaml
 apiVersion: rds.services.k8s.aws/v1alpha1
 kind: DBInstance
@@ -99,10 +96,10 @@ spec:
   dbInstanceIdentifier: "${RDS_INSTANCE_NAME}"
   engine: postgres
   engineVersion: "14"
-  masterUsername: "${RDS_DB_USERNAME}"
+  masterUsername: "postgres"
   masterUserPassword:
     namespace: default
-    name: rds-postgresql-user-creds
+    name: "${RDS_INSTANCE_NAME}-password"
     key: password
 EOF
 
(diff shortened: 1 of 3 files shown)


docs/content/docs/user-docs/field-export.md
```diff
@@ -0,0 +1,128 @@
...
+Kubernetes data plane. The ACK `FieldExport` custom resource was designed to
+bridge the gap between managing the control plane of your ACK resources and
+using the properties of those in your application.
+
+The `FieldExport` resource configures any of the ACK
+controllers to bind and export any spec or status field from an ACK resource
+into a Kubernetes config map or secret. These fields are automatically updated
+as the field value changes over time. You are then able to mount the config map
+or secret onto your Kubernetes pods as environment variables to ingest those
+values.
```
jkatz, line 25:
```suggestion
The `FieldExport` resource configures an ACK
controllers to bind and export any spec or status field from an ACK resource
into a Kubernetes config map or secret. These fields are automatically updated
when any field value changes You are then able to mount the `ConfigMap`
or `Secret` onto your Kubernetes Pods as environment variables that can ingest
those values.
```

jkatz: Also, I believe you can mount the CM/Secret as a file in a Pod?

docs/content/docs/user-docs/field-export.md
```diff
@@ -0,0 +1,128 @@
...
+    env:
+    - name: USER_DATA_BUCKET_LOCATION
+      valueFrom:
+        configMapKeyRef:
+          name: application-user-data-cm # The config map that we created earlier
+          key: "default.export-user-data-bucket"
+```
+
+Looking at the container logs, you can see the environment variable has been
+properly included:
```
jkatz, line 110:
```suggestion
Looking at the container logs, you can see the `USER_DATA_BUCKET_LOCATION`
environment is set with the value from the `ConfigMap`:
```

docs/content/docs/user-docs/field-export.md
```diff
@@ -0,0 +1,128 @@
...
+USER_DATA_BUCKET_LOCATION=http://doc-example-bucket.s3.amazonaws.com/
+```
+
+{{% hint type="warning" title="`FieldExport` RBAC permissions" %}}
+The ACK controller will fetch the source path from the ACK resource assuming
+its service account has the RBAC permissions to read that type of resource. If a
+user has the privileges to create a `FieldExport` resource, it is possible that
+they can create one which fetches fields from a resource they do not have RBAC
+permissions to read directly. This could potentially expose that resource's
+properties to the user.
```
jkatz, line 121:
```suggestion
properties to the unprivileged user.
```

jkatz (commented):
Looks good overall. I left a few suggestions.

docs/content/docs/user-docs/field-export.md
```diff
@@ -0,0 +1,128 @@
...
+  to:
+    name: application-user-data-cm # Matches the ConfigMap we created above
+    kind: configmap
+  from:
+    path: ".status.location"
+    resource:
+      group: s3.services.k8s.aws
+      kind: Bucket
+      name: application-user-data
+```
```
jkatz, line 66: 🤔 It may be good idea to demonstrate how to export two fields. This is the first question I have in my mind after reading this, and it answer the "What does it look like when I export more than one field?" question.

jaypipes (changes requested):
@jkatz has some excellent suggestions, but this is a great start, thank you @RedbackThomson!

docs/content/docs/tutorials/rds-example.md
```diff
@@ -165,14 +161,105 @@ When the `DB Instance Status` says `Available`, you can connect to the database
...
+    - name: PGUSER
+      valueFrom:
+        configMapKeyRef:
+          name: ${RDS_INSTANCE_CONN_CM}
+          key: "default.${RDS_INSTANCE_NAME}-user"
+    - name: PGPASSWORD
+      valueFrom:
+        secretRef:
+          name: "${RDS_INSTANCE_NAME}-password"
+          key: password
```
jaypipes, line 260: awesomesauce.

jaypipes (approved):
fabulous stuff. thank you @RedbackThomson!

PR #1248, opened 2022-04-04
Fix docs dark mode flashing

RedbackThomson:
Description of changes:
If a user had set their dark mode preference to light mode, the website would flash on every page load as it needed to load the dark mode switcher script and then update the CSS. This brief flash becomes very annoying when trying to navigate the website. This pull request adds changes from the latest version of Doks to inline a script which loads the dark mode switcher, so that it's loaded before the page has started rendering. 


jaypipes (approved):
Yes, please! :)

PR #1268, opened 2022-04-20
change some docs references from docker to container

jlbutler:
TL;DR s/docker/container/

Abstracted some docs references from 'docker' to more general 'container' terms. Left all references that make sense (e.g. references from KinD, and links).


RedbackThomson:
Thanks Jesse! We'll try to keep this in mind when we update docs in the future too. While we're at it, what do you think about `Dockerfile` vs `Containerfile` nomenclature?

/lgtm

PR #1307, opened 2022-06-02
Updating the path for the Helm Chart

eminalemdar:
Description of changes:

When installing the ACK Resource Helm Chart there is an environment variable set as CHART_REF=$SERVICE-chart. This creates the folder under /tmp/chart as $SERVICE-CHART. The path in the commands for deleting the CRDs doesn't exist because of that. 

I've updated the page with the right path.


A-Hilaly (approved):
Thanks @eminalemdar !

/lgtm

PR #1310, opened 2022-06-02
Updating the ACK Namespace environment variable

eminalemdar:
When installing the ACK Resource Helm Chart there is an environment variable set as ACK_SYSTEM_NAMESPACE=ack-system. The given variable for deletion is different than what was set before.

I've updated the environment variable with the right one.

Issue #, if available:

Description of changes:


vijtrip2:
Thank you, @eminalemdar !

Issue #1406, opened 2022-07-27
Document FieldExport and AdoptedResource in OpenShift docs


acornett21:
**What is the URL of the document?**
[Red Hat OpenShift](https://aws-controllers-k8s.github.io/community/docs/user-docs/openshift/)

**Which section(s) is the issue in?**
N/A

**What needs fixing?**
Section 4 should become section 5 and section 4 should be optional if users want to use  FieldExport and AdoptedResource in their cluster. This will ensure that the controller starts up reconciling the requested resources.

**Additional context**
Links to the CRD yamls can be found [here](https://github.com/aws-controllers-k8s/runtime/tree/main/config/crd/bases)

PR #1427, opened 2022-08-15
updating OpenShift install docs for optional CRDs

acornett21:
Issue #, if available:
- Fixes: #1406 

Description of changes:
Adding in documentation for users that use OpenShift/OLM on how to install optional CRDs (AdoptedResource/FieldExport) into their clusters before they install any of the controllers.

Signed-off-by: Adam D. Cornett <adc@redhat.com>

### DIFF
--- a/docs/content/docs/user-docs/openshift.md
+++ b/docs/content/docs/user-docs/openshift.md
@@ -109,7 +109,22 @@ Delete `config.txt` and `secrets.txt`.
 If you change the name of either the `ConfigMap` or the `Secret` from the values given above, i.e. `ack-$SERVICE-user-config` and `ack-$SERVICE-user-secrets`, then installations from OperatorHub will not function properly. The Deployment for the controller is preconfigured for these key values.
 {{% /hint %}}
 
-### Step 4: Install the controller
+### Step 4 (Optional): Apply Additional Custom Resource Definitions(CRD)
+To prevent CRD installation conflicts for CRDs shared across multiple AWS Controllers for Kubernetes,
+the [AdoptedResource](../../../reference/common/v1alpha1/adoptedresource/) and [FieldExport](../../../reference/common/v1alpha1/fieldexport/) CRDs are not included in the OpenShift Embedded OperatorHub.
+These must be installed manually by a cluster administrator before any controller is installed by running the following commands:
+
+Apply the `AdoptedResource` CRD
+```bash
+oc apply -f https://raw.githubusercontent.com/aws-controllers-k8s/runtime/main/config/crd/bases/services.k8s.aws_adoptedresources.yaml
+```
+
+Apply the `FieldExport` CRD
+```bash
+oc apply -f https://raw.githubusercontent.com/aws-controllers-k8s/runtime/main/config/crd/bases/services.k8s.aws_fieldexports.yaml
+```
+
+### Step 5: Install the controller
 
 Follow the instructions for [installing the controller using OperatorHub](../install/#install-an-ack-service-controller-with-operatorhub-in-red-hat-openshift).
 


acornett21:
@komish Please take a look.

docs/content/docs/user-docs/openshift.md
```diff
@@ -109,7 +109,22 @@ Delete `config.txt` and `secrets.txt`.
 If you change the name of either the `ConfigMap` or the `Secret` from the values given above, i.e. `ack-$SERVICE-user-config` and `ack-$SERVICE-user-secrets`, then installations from OperatorHub will not function properly. The Deployment for the controller is preconfigured for these key values.
 {{% /hint %}}
 
-### Step 4: Install the controller
+### Step 4 (Optional): Apply Additional Custom Resource Definitions(CRD)
+CRDs are global resources, they cannot be shared, or installed by multiple controllers. This is why [AdoptedResource](../../../reference/common/v1alpha1/adoptedresource/) and [FieldExport](../../../reference/common/v1alpha1/fieldexport/)
+are not packaged with any of the ACK controllers within OpenShift. If either of these resources is required, they will need to be applied manually to the cluster
+before the operator is installed.
+
+Apply the `AdoptedResoutce` CRD
```
komish, line 117:
```suggestion
Apply the `AdoptedResource` CRD
```

docs/content/docs/user-docs/openshift.md
```diff
@@ -109,7 +109,22 @@ Delete `config.txt` and `secrets.txt`.
 If you change the name of either the `ConfigMap` or the `Secret` from the values given above, i.e. `ack-$SERVICE-user-config` and `ack-$SERVICE-user-secrets`, then installations from OperatorHub will not function properly. The Deployment for the controller is preconfigured for these key values.
 {{% /hint %}}
 
-### Step 4: Install the controller
+### Step 4 (Optional): Apply Additional Custom Resource Definitions(CRD)
+CRDs are global resources, they cannot be shared, or installed by multiple controllers. This is why [AdoptedResource](../../../reference/common/v1alpha1/adoptedresource/) and [FieldExport](../../../reference/common/v1alpha1/fieldexport/)
```
komish, line 113: I might reword/reorder this to something like:

"To prevent CRD installation conflicts for CRDs shared across multiple AWS Controllers for Kubernetes, the AdoptedResource and FieldExport CRDs are not included in the OpenShift Embedded Operator Hub. These must be installed manually by a cluster administrator before any controller is installed by running the following commands":

docs/content/docs/user-docs/openshift.md
```diff
@@ -109,7 +109,22 @@ Delete `config.txt` and `secrets.txt`.
...
+are not packaged with any of the ACK controllers within OpenShift. If either of these resources is required, they will need to be applied manually to the cluster
+before the operator is installed.
+
+Apply the `AdoptedResoutce` CRD
+```bash
+oc apply -f https://raw.githubusercontent.com/aws-controllers-k8s/runtime/main/config/crd/bases/services.k8s.aws_adoptedresources.yaml
```
komish, line 119: I'm not familiar with these resources off hand, aside from the GVK versioning, is there a reason why we should not point to `main`'s representation of the AdoptedResource or FieldExport YAMLs? As in, is there any reason why we would need to point to a different tag/branch for this manifest?

komish (changes requested):
Just a couple of quick notes!

docs/content/docs/user-docs/openshift.md
A-Hilaly: It is very unlikely that `AdoptedResource` and `FieldExport` will receive some updates, at least in the short to mid-term. I would say that it's completely fine to point to `main` for now. 

The best way to prevent any issues is by pointing to the runtime version used in a controller's `go.mod`.

docs/content/docs/user-docs/openshift.md
acornett21: That's alot better then what I had. Thanks.

docs/content/docs/user-docs/openshift.md
acornett21: I think for this documentation `main` is really the only thing we have, and if new fields are added, the controller will still use the `old` CRD and be able to reconcile the resource, until the GVK changes. This really isn't the ideal way to document this, ideally it would be tied to version, but I think it's the best we have.

komish (approved):
/lgtm

Issue #1289, opened 2022-05-10
Creating new status fields for RDS DBInstance CRD


davirobin:
**What Is Your Question?**
Is it possible to create new `status` fields for the RDS DBInstance CRD through `ack-generate` code generation?

**What Is The Context For This Question?**
I am currently investigating the integration of [Redhat Service Binding Operator](https://github.com/redhat-developer/service-binding-operator) (SBO) with ACK RDS, more specifically the DBInstance CRD, to enhance native binding. Going through the `code-generator` documentation and the community docs for [determining status fields](https://aws-controllers-k8s.github.io/community/docs/contributor-docs/api-inference/#determining-the-status-fields), there is a lot of material on how the custom code generator works using existing Create `Operation’s` `Output` shape. I need to create fields outside of the `Output` scope.

For example, to integrate the ACK RDS DBInstance with SBO I need to turn the DBInstance into a [Provisioned Service](https://redhat-developer.github.io/service-binding-operator/userguide/exposing-binding-data/provisioned-service.html) via  adding `status.binding.name`.

**Any Additional Info?** 
This question is separate, but adjacent to the [ACK K8s Native Application Binding](https://github.com/aws-controllers-k8s/community/blob/main/docs/design/proposals/native-binding/native-binding.md#ack-k8s-native-application-binding) proposal. `FieldExport` allows the user to export fields from the `status` to secrets whereas I need to include a new field in `status` to point to an existing secret.

RedbackThomson:
This issue we have from the Crossplane team asks something similar: https://github.com/aws-controllers-k8s/community/issues/881

The current answer I have for you is that we currently don't have a configuration for the code-generator to create an arbitrarily shaped field. We do support creating [custom lists or maps](https://github.com/aws-controllers-k8s/code-generator/pull/222) using existing AWS SDK shapes and for [setting the shape of a field](https://github.com/aws-controllers-k8s/code-generator/blob/main/pkg/config/field.go#L100-L108). 

The problem with adding `status.binding.name` to the RDS CRDs is that it is only applicable to SBO. The team at Crossplane maintain their own separate set of CRDs for every service, and in that they inject new fields that are only applicable to them and then also write custom code to initialize those fields. I understand it's annoying to maintain your own code, but putting custom SBO code into ACK controllers makes for a confusing experience for those who aren't using SBO. 

[...]

PR #1539, opened 2022-11-09
Service Binding Specification for Kubernetes

baijum:
Signed-off-by: Baiju Muthukadan <baiju.m.mail@gmail.com>

Issue #1289 , if available:

Description of changes:

Proposal to add support for [Service Binding Specification for Kubernetes](https://github.com/servicebinding/spec)

### DIFF
--- a/docs/design/proposals/provisioned-service/provisioned-service.md
+++ b/docs/design/proposals/provisioned-service/provisioned-service.md
@@ -20,7 +20,7 @@ Reference as per the
 specification](https://github.com/servicebinding/spec#direct-secret-reference).
 This solution needs more effort from the application developers.
 
-The FieldExports resources required for application developers are the
+The FieldExport resources required for application developers are the
 same, but every developer needs to repeat this same configuration for
 every application they deploy. The proposed solution avoids this
 repetition through Provisioned Services.
--- a/docs/design/proposals/provisioned-service/provisioned-service.md
+++ b/docs/design/proposals/provisioned-service/provisioned-service.md
@@ -1,0 +1,108 @@
+# Service Binding Specification for Kubernetes
+
+## Problem Statement
+
+Application developers require specific values to connect their
+applications to various services. In order to facilitate this connection
+problem, the [Service Binding Specification for
+Kubernetes](https://servicebinding.io) defines Provisioned
+Services as a standard way to expose connection information from backing
+services. Supporting Provisioned Services within AWS Controllers for
+Kubernetes makes application connectivity to ACK-backed services easy
+and seamless.
+
+## Existing Solutions
+
+AWS Controllers for Kubernetes have support for a [FieldExport
+API](https://github.com/aws-controllers-k8s/community/blob/main/docs/design/proposals/native-binding/native-binding.md).
+This can be used to create a Secret that can be used for [Direct Secret
+Reference as per the
+specification](https://github.com/servicebinding/spec#direct-secret-reference).
+This solution needs more effort from the application developers.
+
+The FieldExports resources required for application developers are the
+same, but every developer needs to repeat this same configuration for
+every application they deploy. The proposed solution avoids this
+repetition through Provisioned Services.
+
+## Proposed Solution
+
+The proposed solution is to make various ACK resources become
+Provisioned Services. For example, make the DBInstance resource provided
+by the RDS controller a Provisioned Service. Each controller would use
+their main custom resource as a Provisioned Service resource. For each
+Provisioned Service, identify the fields that need to be exported to
+create the Secret resource.
+
+The application developer experience is explained in this demo video:
+[https://www.youtube.com/watch?v=AXXWv7N12JM](https://www.youtube.com/watch?v=AXXWv7N12JM)
+
+(This is created as part of a [Proof of
+Concept](https://github.com/aws-controllers-k8s/community/issues/1289))
+
+The structure change for DBInstance:
+
+```
+// ServiceBindingSecretReference defines a mirror of corev1.LocalObjectReference
+type ServiceBindingSecretReference struct {
+	// Name of the referent secret.
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string `json:"name"`
+}
+
+// DBInstanceStatus defines the observed state of DBInstance
+type DBInstanceStatus struct {
+	// Binding exposes the Secret for Service Binding to conform the Provisioned Service
+	// as per the Service Binding Specification for Kubernetes.
+	// Ref. https://github.com/servicebinding/spec#provisioned-service
+	Binding *ServiceBindingSecretReference `json:"binding,omitempty"`
+```
+
+### Benefits
+
+When there are many microservices connecting to the same service, the
+label selector feature helps to easily bind all of them with minimal
+configuration. Service Binding Specification standard provides a uniform
+experience for all the applications connectivity with services.
+
+### Implementation
+
+Since the API changes required for various custom resources are
+standard, and they can be generated for all controllers. However, the
+changes for each of these controllers can be implemented gradually.
+
+### RBAC
+
+Since a Secret resource needs to be created, updated, and deleted, the
+ClusterRole requires "*create",* "*update", and "delete"* permissions.
+Currently, this permission is not set in the existing controllers.
+
+## RDS Controller Fields
+
+The RDS controller supports multiple databases. Based on the database,
+the value for type could be set. For example, postgresql for PostgreSQL
+and mysql for MySQL.
+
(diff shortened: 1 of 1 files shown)


docs/design/proposals/provisioned-service/provisioned-service.md
```diff
@@ -0,0 +1,108 @@
...
+## Existing Solutions
+
+AWS Controllers for Kubernetes have support for a [FieldExport
+API](https://github.com/aws-controllers-k8s/community/blob/main/docs/design/proposals/native-binding/native-binding.md).
+This can be used to create a Secret that can be used for [Direct Secret
+Reference as per the
+specification](https://github.com/servicebinding/spec#direct-secret-reference).
+This solution needs more effort from the application developers.
+
+The FieldExports resources required for application developers are the
```
jaypipes, line 23:
```suggestion
The FieldExport resources required for application developers are the
```

docs/design/proposals/provisioned-service/provisioned-service.md
```diff
@@ -0,0 +1,108 @@
...
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+	Name string `json:"name"`
+}
+
+// DBInstanceStatus defines the observed state of DBInstance
+type DBInstanceStatus struct {
+	// Binding exposes the Secret for Service Binding to conform the Provisioned Service
+	// as per the Service Binding Specification for Kubernetes.
+	// Ref. https://github.com/servicebinding/spec#provisioned-service
+	Binding *ServiceBindingSecretReference `json:"binding,omitempty"`
```
jaypipes, line 58: We would want to place this field within the `Status.ACKResourceMetadata` subfield. That way, it won't conflict with any resources that might have a `Binding` field already in their status...

docs/design/proposals/provisioned-service/provisioned-service.md
```diff
@@ -0,0 +1,108 @@
...
+
+Since the API changes required for various custom resources are
+standard, and they can be generated for all controllers. However, the
+changes for each of these controllers can be implemented gradually.
+
+### RBAC
+
+Since a Secret resource needs to be created, updated, and deleted, the
+ClusterRole requires "*create",* "*update", and "delete"* permissions.
+Currently, this permission is not set in the existing controllers.
```
jaypipes, line 78: The [current `ClusterRole`](https://github.com/aws-controllers-k8s/s3-controller/blob/01c6f0aa4a8a1b33c8c552242affcdbc8f4df651/helm/templates/cluster-role-controller.yaml#L32-L40) allows getting, patching, listing and watching Secret resources. Why would we need to add Create and Delete permissions?

docs/design/proposals/provisioned-service/provisioned-service.md
```diff
@@ -0,0 +1,108 @@
...
+
+|**Field**  | **Value**                |  **Remarks**
+|-----------|--------------------------|-------------------------------
+|type       |  postgresql/mysql        |   Based on Spec.Engine
+|provider   |  aws                     |   
+|host       |  Status.Endpoint.Address |   
+|port       |  Status.Endpoint.Port    |   
+|username   |  Spec.MasterUsername     |   
+|password   |                          |   Based on Spec.MasterUserPassword
+|database   |  Spec.Engine?            |
```
jaypipes, line 96: Are the above fields intended to be populated as keys within a Secret?

jaypipes (commented):
Hi @baijum, cool idea! I like this proposal. Couple inline questions/comments for you to think about.

docs/design/proposals/provisioned-service/provisioned-service.md
baijum: Yes, the keys will change depending on the service. The above keys are examples that can be used with a relational database.

docs/design/proposals/provisioned-service/provisioned-service.md
baijum: The Service Binding Specification expects the Provisioned Service (e.g., `DBInstance`) to point to a Secret resource with all values required for connectivity from the application. In the case of `DBInstance`, the corresponding controller needs to ensure the presence of the Secret resource. To create a Secret, `Create` permission is required. Similarly when the `DBInstance` is getting deleted, cleaning up the corresponding Secret resource would be good. Hence the `Delete` permission.

docs/design/proposals/provisioned-service/provisioned-service.md
baijum: The Service Binding Specification expects the Provisioned Service (e.g., `DBInstance`) to point to the Secret resource from `.status.binding.name` attribute. I think we can confirm if that field is not conflicting with any existing fields.

jaypipes: @baijum the problem is that the `Status` and `Spec` fields for all ACK resources are automatically generated by looking at the AWS service API's model definition. We don't know if there are APIs that have a `Binding` field in any Create operation's Output shape. If there were, that would conflict with this.

docs/design/proposals/provisioned-service/provisioned-service.md
jaypipes: @baijum how does the controller know what the Secret's name will be?

baijum: The `DBInstance` controller generates the Secret. For example, the name could be the same as the `DBInstance` name.

docs/design/proposals/provisioned-service/provisioned-service.md
baijum: Can we check if the`.status.binding.name` attribute exists for any API? Also is it possible to reserve this field for service binding, so that it will not conflict in the future?

jaypipes: We could do this, sure, but the problem is we cannot predict the future. Since ACK CRDs are generated from the upstream API models, what happens if an upstream service API adds a Binding field in the future? This is why we place these kinds of fields into `Status.ACKResourceMetadata` to ensure that we don't conflict.

baijum: I have created a proposal to the spec here: https://github.com/servicebinding/spec/pull/223
Let me know if creating a Secret resource with the same name as that of `DBInstance` resource (or any other service) would be acceptable.

jaypipes: @baijum I left a comment in the other PR. I would be OK with a separate reconciler (ala the `FieldExportReconciler` we already have in ACK) that would create/manage Secret resources like servicebinding.io expects.

RedbackThomson: +1 with Jay's comment here. This functionality seems similar enough to `FieldExport`, in that it accesses CRD fields and outputs into a `Secret`, that I believe it could simply be handled by that reconciler. In fact, we could add this feature by simply adding an optional field within the `FieldExport` CRD that provides an alternative to `from.path`, instead being `from.bindingSpecification: true/false` (name TBD). This would sidestep the need to reserve the `.status.binding` field.

docs/design/proposals/provisioned-service/provisioned-service.md
```diff
@@ -0,0 +1,108 @@
...
+When there are many microservices connecting to the same service, the
+label selector feature helps to easily bind all of them with minimal
+configuration. Service Binding Specification standard provides a uniform
+experience for all the applications connectivity with services.
+
+### Implementation
+
+Since the API changes required for various custom resources are
+standard, and they can be generated for all controllers. However, the
+changes for each of these controllers can be implemented gradually.
```
RedbackThomson, line 72: This implementation doesn't really touch on how we know which fields are required for the binding specification? I imagine that each of the bindings will need to be manually crafted. It would be good to have a section that describes how controller maintainers can add new bindings and how they know what the spec should be.

baijum:
I am closing this proposal as [FieldExport](https://gist.github.com/baijum/96158408eaf544f692acfad42f2b49df) can be used as a solution for the problem.

RedbackThomson:
@baijum I see you have closed this PR. Do you have an example for the ACK community with how FieldExport could implement the service binding spec?

baijum:
> @baijum I see you have closed this PR. Do you have an example for the ACK community with how FieldExport could implement the service binding spec?

As I explained here: https://gist.github.com/baijum/96158408eaf544f692acfad42f2b49df
The Secret generated through FieldExport can be used for ServiceBiding. The Secret must have `type` field and optionally it can also add `provider`, something like this:
```
apiVersion: v1
kind: Secret
metadata:
  name: rds-postgres-field-exports
  namespace: ack-system
stringData:
  type: postgresql
  provider: was
```

PR #1614, opened 2023-01-06
Add a user doc page for deletion policy

RedbackThomson:
Description of changes:
Adds a page to the `Getting Started` section of the user docs with information about the new deletion policy feature

### DIFF
--- a/docs/content/docs/user-docs/deletion-policy.md
+++ b/docs/content/docs/user-docs/deletion-policy.md
@@ -1,0 +1,84 @@
+---
+title: "Retain AWS Resources after Deletion"
+description: "Using the ACK deletion policy configuration"
+lead: "Using the ACK deletion policy configuration"
+draft: false
+menu:
+  docs:
+    parent: "getting-started"
+weight: 70
+toc: true
+---
+
+The ACK controllers are designed to create, update and delete AWS resources
+following the lifecycle of their respective Kubernetes custom resources. As a
+result, when deleting an ACK resource, the underlying AWS resource is first
+deleted before deleting its Kubernetes custom resource. This behavior is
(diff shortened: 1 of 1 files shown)


docs/content/docs/user-docs/deletion-policy.md
```diff
@@ -0,0 +1,84 @@
+---
+title: "Retain AWS Resources after Deletion"
```
jljaco, line 2: consider calling this `Retain AWS resources after CR deletion` for clarity

docs/content/docs/user-docs/deletion-policy.md
```diff
@@ -0,0 +1,84 @@
...
+- `delete` - **(Default)** Deletes the resource from AWS before deleting it from
+  K8s
+- `retain` - Keeps the AWS resource intact before deleting it from K8s
+
+## Configuring the deletion policy
+
+### Using Helm values
+
+To set a controller-wide deletion policy, which will apply to all ACK resources
+owned by the ACK controller, you can set the `deletionPolicy` Helm chart value.
```
jljaco, line 45: might as well explicitly state that this Helm chart var can take on the same values described above (i.e., `delete` and `retain`)

docs/content/docs/user-docs/deletion-policy.md
```diff
@@ -0,0 +1,84 @@
...
+  name: retained-bucket
+  annotations:
+    services.k8s.aws/deletion-policy: retain
+spec:
+  name: my-retained-bucket
+```
+
+*Note: The key for annotating a single resource is not the same as the key when
+annotating a namespace. You do not need to provide the name of the service as a
+suffix for a single resource.*
```
jljaco, line 84: `s/suffix/prefix/`

PR #1736, opened 2023-03-17
Update command to retrieve Sagemaker controller version

ananth102:
Description of changes:

Dynamically retrieve the latest version of the Sagemaker controller instead of hardcoding, script is borrowed from https://aws-controllers-k8s.github.io/community/docs/user-docs/install/.

### DIFF
--- a/docs/content/docs/tutorials/autoscaling-example.md
+++ b/docs/content/docs/tutorials/autoscaling-example.md
@@ -142,7 +142,12 @@ Get the Application Auto Scaling Helm chart and make it available on the client 
 ```bash
 export HELM_EXPERIMENTAL_OCI=1
 export SERVICE=applicationautoscaling
-export RELEASE_VERSION=v0.2.15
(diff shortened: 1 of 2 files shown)


jljaco (commented):
I'm ambivalent about this method of obtaining the most recent release version.  

On the one hand, it's obviously more maintainable.

On the other, the GH API is frequently unavailable; for example, I just attempted to retrieve this value now and got:

```
 $ curl -sL https://api.github.com/repos/aws-controllers-k8s/sagemaker-controller/releases/latest 
{"message":"API rate limit exceeded for XXXXXXXX. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

Could you perhaps split this command pipeline into 2 distinct steps with an error check in between to verify that the user has gotten a valid tag value?

surajkota (commented):
lets do the same for application autoscaling tutorial as well

A-Hilaly (approved):
Niice, i like it!
/lgtm

A-Hilaly (commented):
Can you also add `curl` to the list of requirements?

ananth102:
> Can you also add `curl` to the list of requirements?

Sure will do

